PR #39, opened 2021-01-11
Loader and initializer tweaks.

olonho:


skiko/src/jvmMain/kotlin/org/jetbrains/skiko/Library.kt
```diff
@@ -38,7 +37,12 @@ object Library {
         val name = "skiko-$hostId"
         val platformName = System.mapLibraryName(name)
 
-        val hash = Library::class.java.getResourceAsStream("$resourcePath$platformName.sha256").use {
+        val hashResourceStream = Library::class.java.getResourceAsStream(
+            "$resourcePath$platformName.sha256"
+        ) ?: throw Error(
+            "Missing $platformName, use `implementation(compose.desktop.currentOs)`"
```
jimgoog, line 43: Probably don't want to reference Compose in Skiko, right?  Probably Compose should catch this error and rethrow with the more compose-specific message.

olonho: Right, reworked.

igordmn (approved):

PR #55, opened 2021-01-29
Immediate redraw on resize and window start

igordmn:
Now it is noticeably better on Windows, but there is still issues (we still can see the black lines on window resize).

On Linux/macOs there is no noticiable improvement, but it is the right thing to do - redraw synchronously before vsync. Without immediate redraw we will refresh content only after vsync, so we lose one frame after start/resize.

To implement a perfect start/resize we probably shouldn't rely on JFrame, instead we should manipulate the native window.

### DIFF
--- a/skiko/src/jvmMain/kotlin/org/jetbrains/skiko/redrawer/LinuxRedrawer.kt
+++ b/skiko/src/jvmMain/kotlin/org/jetbrains/skiko/redrawer/LinuxRedrawer.kt
@@ -12,10 +12,7 @@ internal class LinuxRedrawer(
     private val layer: HardwareLayer
 ) : Redrawer {
     private val context = layer.lockDrawingSurface {
-        val context = it.createContext()
-        it.makeCurrent(context)
-        it.setSwapInterval(if (SkikoProperties.vsyncEnabled) 1 else 0)
-        context
+        it.createContext()
     }
     private var isDisposed = false
 
@@ -31,6 +28,16 @@ internal class LinuxRedrawer(
         check(!isDisposed)
         toRedraw.add(this)
         frameDispatcher.scheduleFrame()
+    }
+
+    override fun redrawImmediately() = layer.lockDrawingSurface {
+        check(!isDisposed)
+        update(System.nanoTime())
+        it.makeCurrent(context)
+        draw()
+        it.setSwapInterval(0)
+        it.swapBuffers()
+        OpenGLApi.instance.glFinish()
     }
 
     private fun update(nanoTime: Long) {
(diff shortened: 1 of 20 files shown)


skiko/src/jvmMain/kotlin/org/jetbrains/skiko/redrawer/LinuxRedrawer.kt
```diff
@@ -37,10 +30,18 @@ internal class LinuxRedrawer(
...
-            layer.update(nanoTime)
-        }
+    override fun redrawImmediately() = layer.lockDrawingSurface {
+        check(!isDisposed)
+        update(System.nanoTime())
+        it.makeCurrent(context)
+        draw()
+        it.setSwapInterval(0)
+        it.swapBuffers()
+        OpenGLApi.instance.glFinish()
```
olonho, line 40: It doesn’t look right to add it here, can we make backend-specific sync code.

igordmn: LinuxRedrawer/WindowsRedrawer/MacOsRedrawer are bound completely to OpenGL.

We can rename them to LinuxOpenGLRedrawer/WindowsOpenGLRedrawer/MacOsOpenGLRedrawer

igordmn: Extract backend-independent code we should only after we implement another hardware backends.
Before that we don't know exactly what is backend-independent :).

olonho: Maybe merge after https://github.com/JetBrains/skiko/pull/56 and put move code to proper place?

igordmn: Event after #56 this is the proper place. #56 only adds raster render and platform for other renderings.

All functions (createContext, makeCurrent, swapBuffers) are OpenGL specific, if I am not mistaken.

olonho: OK, then probably slight rename of classes, or eventual splitting redrawer into backend independent and dependent parts looks reasonable. Anyway, if it works OK with software rasterizer - I'm fine with it.

olonho (approved):

skiko/src/jvmMain/kotlin/org/jetbrains/skiko/redrawer/LinuxRedrawer.kt
igordmn: Ok, will rename and merge after #56

Rsedaikin (approved):

skiko/src/jvmMain/kotlin/org/jetbrains/skiko/redrawer/LinuxRedrawer.kt
igordmn: Renamed

PR #58, opened 2021-02-03
FrameDispatcherTest

igordmn:


samples/SkijaInjectSample/build.gradle.kts
```diff
@@ -64,4 +67,8 @@ tasks.withType<Test> {
         systemProperty("sun.java2d.dpiaware", "false")
         systemProperty("sun.java2d.uiScale", "1")
     }
+}
+
+tasks.withType<KotlinCompile>().configureEach {
+    kotlinOptions.freeCompilerArgs += "-Xopt-in=kotlin.RequiresOptI"
```
olonho, line 73: kotlin.RequiresOptI*n*. Why it is actually needed, just to get rid of warnings?

samples/SkijaInjectSample/src/test/kotlin/org/jetbrains/skiko/FrameDispatcherTest.kt
```diff
@@ -0,0 +1,165 @@
...
+        lateinit var frameDispatcher: FrameDispatcher
+        frameDispatcher = FrameDispatcher(scope = this) {
+            frameCount++
+            frameDispatcher.scheduleFrame()
+            frameDispatcher.scheduleFrame()
+            frameDispatcher.scheduleFrame()
+        }
+        frameDispatcher.scheduleFrame()
+
+        yield()
```
olonho, line 115: Is this test actually reliable?

samples/SkijaInjectSample/build.gradle.kts
igordmn: Without it there is a warning:

"This class can only be used with the compiler argument ‘-Xopt-in=kotlin.RequiresOptIn’"

https://kotlinlang.org/docs/reference/opt-in-requirements.html#experimental-status-of-the-opt-in-requirements

samples/SkijaInjectSample/src/test/kotlin/org/jetbrains/skiko/FrameDispatcherTest.kt
igordmn: Yes, we should dispatch frame immediately after any suspend point. This is the requirement for FrameDispatcher.

samples/SkijaInjectSample/build.gradle.kts
olonho: OK, just put the proper opt-in name then.

igordmn: Oh, thanks. I didn't notice that I lost letter in the commit :)

olonho (approved):

PR #59, opened 2021-02-03
Fix resize artifacts with fraction DPI

igordmn:
Fixes https://github.com/JetBrains/compose-jb/issues/179


olonho (approved):

Rsedaikin (approved):

igordmn:
I made mistake, there is always an offset -1 px now :). After I fix this, it requires more testing, on compose side as well

igordmn:
> I made a mistake, there is always an offset -1 px now :). After I fix this, it requires more testing, on compose side as well

Ok, I was not able to get rid of this. Try another time.

PR #60, opened 2021-02-03
macOs. call update only once per draw if vsync is disabled

igordmn:


skiko/src/jvmMain/kotlin/org/jetbrains/skiko/redrawer/MacOsOpenGLRedrawer.kt
```diff
@@ -128,11 +135,25 @@ private open class AWTGLLayer(private val containerPtr: Long, setNeedsDisplayOnB
...
     open fun setNeedsDisplay() = setNeedsDisplayOnMainThread(ptr)
 
+    suspend fun display() = display.runAndAwait {
+        setNeedsDisplay()
+    }
+
     // Called in AppKit Thread
     protected open fun canDraw() = true
 
+    @Suppress("unused")
```
olonho, line 145: Maybe comment that it's called from JNI?

skiko/src/jvmMain/kotlin/org/jetbrains/skiko/redrawer/MacOsOpenGLRedrawer.kt
```diff
@@ -144,3 +165,26 @@ private external fun initContainer(layer: HardwareLayer): Long
...
+internal class Task {
+    @Volatile
+    private var onFinish: CompletableDeferred<Unit>? = null
+
+    /**
+     * Run task and await its finishing (i.e. calling of [finish])
+     */
+    suspend fun runAndAwait(run: suspend () -> Unit) {
+        check(onFinish == null)
+        onFinish = CompletableDeferred()
```
olonho, line 178: Still not channels? But why?

igordmn: I am not able to write a simple code with channels. This code is wrong:

```
internal class Task {
    @Volatile
    private var channel = Channel<Unit>(Channel.RENDEZVOUS)

    suspend fun runAndAwait(run: suspend () -> Unit) {
        run()
        channel.receive()
    }

    fun finish() {
        channel.offer(Unit)
    }
}
```
Because we can call finish between run() and channel.receive().

This code is also wrong:
```
internal class Task {
    @Volatile
    private var channel = Channel<Unit>(1)

    suspend fun runAndAwait(run: suspend () -> Unit) {
        run()
        channel.receive()
    }

    fun finish() {
        channel.offer(Unit)
    }
}
```
We can call finish() before run(), so channel.receive() will not wait.

If one understands how CompletableDeferred works, code become simple and error prone. Because we await/recreate CompletableDeferred always from the same coroutine.

skiko/src/jvmMain/kotlin/org/jetbrains/skiko/redrawer/MacOsOpenGLRedrawer.kt
igordmn: Done

skiko/src/jvmMain/kotlin/org/jetbrains/skiko/redrawer/MacOsOpenGLRedrawer.kt
olonho: If I understand expected semantics correctly, you need smth like
internal class Task {
    private val channel = Channel<Unit>(1)
    private var done = false

    suspend fun runAndAwait(run: suspend () -> Unit) {
        done = false
        run()
        if (!done) channel.receive()
    }

    fun finish() {
        if (!done) {
          done = true
          channel.offer(Unit)
        }
    }
}

igordmn: Seems ok, rewrote to it (with some changes).

Probably it is slighly less robust )), so I wrote tests.

olonho (approved):

PR #61, opened 2021-02-03
Rewrite FPS counting

igordmn:
- property `skiko.fps.longFrames.show` to show frames longer than display refresh rate
- replace `skiko.fps.count` by `skiko.fps.periodSeconds`
- log timestamp
- remove `skiko.fps.probability`. min/max FPS are more informative

### DIFF
--- a/.run/SkijaInjectSample (show long frames) .run.xml
+++ b/.run/SkijaInjectSample (show long frames) .run.xml
@@ -1,0 +1,25 @@
+<component name="ProjectRunConfigurationManager">
+  <configuration default="false" name="SkijaInjectSample (show long frames) " type="GradleRunConfiguration" factoryName="Gradle">
+    <ExternalSystemSettings>
+      <option name="executionName" />
+      <option name="externalProjectPath" value="$PROJECT_DIR$/samples/SkijaInjectSample" />
+      <option name="externalSystemIdString" value="GRADLE" />
+      <option name="scriptParameters" value="-Dskiko.fps.longFrames.show=true" />
+      <option name="taskDescriptions">
+        <list />
+      </option>
+      <option name="taskNames">
+        <list>
+          <option value="run" />
+        </list>
+      </option>
+      <option name="vmOptions" value="" />
+    </ExternalSystemSettings>
+    <ExternalSystemDebugServerProcess>true</ExternalSystemDebugServerProcess>
+    <ExternalSystemReattachDebugProcess>true</ExternalSystemReattachDebugProcess>
+    <DebugAllEnabled>false</DebugAllEnabled>
+    <method v="2">
+      <option name="Gradle.BeforeRunTask" enabled="false" tasks="publishToMavenLocal" externalProjectPath="$PROJECT_DIR$/skiko" vmOptions="" scriptParameters="" />
+    </method>
+  </configuration>
+</component>
--- a/skiko/src/jvmMain/kotlin/org/jetbrains/skiko/FPSCounter.kt
+++ b/skiko/src/jvmMain/kotlin/org/jetbrains/skiko/FPSCounter.kt
@@ -51,12 +51,9 @@ internal fun defaultFPSCounter(
 
     FPSCounter(
         periodSeconds = fpsPeriodSeconds,
-        showLongFrames = fpsLongFramesShow
-    ) {
-        when {
-            fpsLongFramesShow && fpsLongFramesMillis != null -> fpsLongFramesMillis!!
-            fpsLongFramesShow -> 1.5 * 1000 / refreshRate
-            else -> 0.0
+        showLongFrames = fpsLongFramesShow,
+        getLongFrameMillis = {
+            fpsLongFramesMillis ?: 1.5 * 1000 / refreshRate
         }
-    }
+    )
 }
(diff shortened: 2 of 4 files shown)


.run/SkijaInjectSample (show long frames) .run.xml
```diff
@@ -0,0 +1,25 @@
+<component name="ProjectRunConfigurationManager">
```
olonho, line 1: Why it does belong to VCS?

skiko/src/jvmMain/kotlin/org/jetbrains/skiko/FPSCounter.kt
```diff
@@ -1,43 +1,62 @@
...
-            val min = (1000.0 / times.quantile(1 - quantile)).roundToInt()
-            val max = (1000.0 / times.quantile(quantile)).roundToInt()
-            val probability = (100 * probability).roundToInt()
-            println("FPS $average ($min-$max $probability%)")
+    FPSCounter(
+        periodSeconds = fpsPeriodSeconds,
+        showLongFrames = fpsLongFramesShow
+    ) {
+        when {
+            fpsLongFramesShow && fpsLongFramesMillis != null -> fpsLongFramesMillis!!
```
olonho, line 57: Somewhat confusing logic. What is intention here?

skiko/src/jvmMain/kotlin/org/jetbrains/skiko/SkikoProperties.kt
```diff
@@ -5,8 +5,13 @@ internal object SkikoProperties {
...
     val fpsEnabled: Boolean by property("skiko.fps.enabled", default = false)
-    val fpsCount: Int by property("skiko.fps.count", default = 300)
-    val fpsProbability: Double by property("skiko.fps.probability", default = 0.97)
+    val fpsPeriodSeconds: Double by property("skiko.fps.periodSeconds", default = 2.0)
+    val fpsLongFramesShow: Boolean by property("skiko.fps.longFrames.show", default = false)
+
+    /**
+     * If the property isn't defined, but skiko.fps.longFrames.show is true, this property will be 1.5 * (1000 / displayRefreshRate)
+     */
+    val fpsLongFramesMillis: Double? by property("skiko.fps.longFrames.millis", default = null)
```
olonho, line 14: Why not write it here explicitly, i.e 
```
val fpsLongFramesMillis: Double? by lazy {
    val millis = if (fpsLongFramesShow) 1.1 else null
    property("skiko.fps.longFrames.millis", default = millis) 
}

File: .run/SkijaInjectSample (show long frames) .run.xml (hunk 1 above)
Comment (igordmn) [reply .run/SkijaInjectSample (show long frames) .run.xml:1]: So everyone who uses IDEA can simply run it.

If they will get in the way, we can remove them, or group them into separate folder.

File: skiko/src/jvmMain/kotlin/org/jetbrains/skiko/SkikoProperties.kt (hunk 3 above)
Comment (igordmn) [reply skiko/src/jvmMain/kotlin/org/jetbrains/skiko/SkikoProperties.kt:14]: I think property itself should be simple and don't contain any complex logic. 
It is FpsCounter's job to check this.

File: skiko/src/jvmMain/kotlin/org/jetbrains/skiko/FPSCounter.kt (hunk 2 above)
Comment (igordmn) [reply skiko/src/jvmMain/kotlin/org/jetbrains/skiko/FPSCounter.kt]: Agree, simplified logic a little.

File: .run/SkijaInjectSample (show long frames) .run.xml (hunk 1 above)
Comment (olonho) [reply .run/SkijaInjectSample (show long frames) .run.xml:1]: Guess better way to provide predefine run config is Gradle task.

File: skiko/src/jvmMain/kotlin/org/jetbrains/skiko/SkikoProperties.kt (hunk 3 above)
Comment (olonho) [reply skiko/src/jvmMain/kotlin/org/jetbrains/skiko/SkikoProperties.kt:14]: Well, isn’t that a definition of this property? For other it’s getting from JVM dictionary, for that one - somewhat trickier computation.

Comment (igordmn) [reply skiko/src/jvmMain/kotlin/org/jetbrains/skiko/SkikoProperties.kt:16]: 'null' here means we should use automatic detection of fpsLongFramesMillis.

Without this logic we can define `fpsLongFramesMillis: Double` instead of `fpsLongFramesMillis: Double?`

In this case, code like this:
```
val fpsLongFramesMillis: Double by lazy {
    val millis = if (fpsLongFramesShow) 1.1 else 0.0
    property("skiko.fps.longFrames.millis", default = millis) 
}
```
Will be strange (why 0.0?)

We can do here not only parsing, but also write some domain structure like this:
```
object SkikoProperties {
    sealed class FpsLongFrames {
       class Auto: FpsLongFrames 
       class Manual: FpsLongFrames 
    }

    class Fps {
           val fpsLongFrames: FpsLongFrames?
    }

    val fps: Fps?
}
```

But it is more complex.

Anyway, I think it is not so important :)

I changed comments inside SkikoPoperties, it is slightly better now.

.run/SkijaInjectSample (show long frames) .run.xml
igordmn: After thinking about it more, I agree. Better to make them as tasks.

Will do this in another PR (there is another tasks that should be  rewriten).

I have working prototype.

skiko/src/jvmMain/kotlin/org/jetbrains/skiko/SkikoProperties.kt
olonho: OK, as you like.

olonho (approved):

Rsedaikin (approved):

PR #83, opened 2021-03-26
Using an integrated video card on Metal rendering

Rsedaikin:


olonho (commented):
Can we somehow check, that if discrete device is already active - we prefer it. I.e. we just do not want to turn it on.

skiko/src/jvmMain/objectiveC/macos/MetalRedrawer.mm
```diff
@@ -100,6 +100,25 @@ JNIEXPORT jlong JNICALL Java_org_jetbrains_skiko_redrawer_MetalRedrawer_makeMeta
     return (jlong) renderTarget;
 }
 
+id<MTLDevice> MTLCreateInegratedDevice() {
```
kropp, line 103: typo: t is missing in In**t**egrated

Rsedaikin: Done.

olonho (approved):

Issue #76, opened 2021-03-17
Metal. Fix vsync issues with multiple windows


igordmn:
1. Fix freeze when we resize the second window and overlap the first window:

https://user-images.githubusercontent.com/5963351/111430915-2381d280-870c-11eb-801e-92ecd5a35162.mov
reproducible in SkijaInjectSample, but not reproducible in Compose. Probab

2. vsync doesn't work if window isn't visible (FPS is unlimited, we eat CPU)

3. Support multiple monitors with different refresh rates. For now refresh rate will be equal to the lower refresh rate.

okushnikov:
Please check the <a href="https://youtrack.jetbrains.com/issue/SKIKO-76">following ticket</a> on YouTrack for follow-ups to this issue. GitHub issues will be closed in the coming weeks.

PR #86, opened 2021-03-31
macOs. draw in a background thread.

igordmn:
fixes 1) and 3) in https://github.com/JetBrains/skiko/issues/76

also fixes slow event dispathing in Compose (it is because vsync blocked AWT thread)


olonho:
Is the description on PR correct now?

PR #89, opened 2021-04-05
Fix Metal crazy rendering on invisible windows.

olonho:


skiko/src/jvmMain/kotlin/org/jetbrains/skiko/FrameDispatcher.kt
```diff
@@ -23,14 +22,15 @@ class FrameDispatcher(
...
 
     private val job = scope.launch {
         while (true) {
-            needFrame.await()
-            needFrame = CompletableDeferred()
+            if (!needFrame)
+                channel.receive()
+            needFrame = false
             onFrame()
-            yield()
```
igordmn, line 33: We need this `yield`.

If we have animation then `scheduleFrame` will be called inside `onFrame` and we will never free the Swing dispatcher.

skiko/src/jvmMain/kotlin/org/jetbrains/skiko/FrameDispatcher.kt
```diff
@@ -23,14 +22,15 @@ class FrameDispatcher(
...
-    private var needFrame = CompletableDeferred<Unit>()
+    private var needFrame = false
+    private val channel = Channel<Unit>(Channel.CONFLATED)
 
     private val job = scope.launch {
         while (true) {
-            needFrame.await()
-            needFrame = CompletableDeferred()
+            if (!needFrame)
+                channel.receive()
```
igordmn, line 31: We have one more unnecessary frame. See `perform tasks scheduled in the frame after the frame`.

It is because we push value to the channel, but don't receive it on the second frame. We receive it only on the third frame.

olonho: Yep, simplified that place.

igordmn (approved):

PR #94, opened 2021-04-21
Fix IM handling.

olonho:


skiko/src/jvmMain/kotlin/org/jetbrains/skiko/SkiaLayer.kt
```diff
@@ -163,6 +157,15 @@ open class SkiaLayer(
         backedLayer.enableInputMethods(enable)
     }
 
+    override fun getInputMethodListeners(): Array<InputMethodListener> {
+        return backedLayer.getInputMethodListeners()
+    }
+
+    override fun processInputMethodEvent(e: InputMethodEvent?) {
+        if (e != null)
```
igordmn, line 165: Maybe just make `e` nullable?
```
fun doProcessInputMethodEvent(e: InputMethodEvent?)
```

olonho:
> Probaly we don't need super.removeInputMethodListener / super.addInputMethodListener.
> 
> The idea is to fully delegate work to `backedLayer`

It doesn't work without call to super, you can check.

PR #96, opened 2021-04-26
Macos headerless window

SergeevPavel:


skiko/src/jvmMain/objectiveC/macos/Drawlayer.mm
```diff
@@ -141,16 +141,6 @@ JNIEXPORT void JNICALL Java_org_jetbrains_skiko_HardwareLayer_nativeDispose(JNIE
     }
 }
 
-JNIEXPORT jlong JNICALL Java_org_jetbrains_skiko_HardwareLayer_getWindowHandle(JNIEnv *env, jobject canvas)
-{
-    LayerHandler *layer = findByObject(env, canvas);
-    if (layer != NULL)
-    {
-        return (jlong)[layer window];
```
igordmn, line 149: Let's remove layer.window and layerSet.window initialization inside nativeInit to avoid code duplication.

P.S.
I checked that performance of getting windowHandle every time isn't so bad:
```
repeat(10) {
    JFrame().apply {
        size = Dimension(200, 300)
        isVisible = true
    }
}
window.pack()
println(measureTime {
    repeat(100) {
        window.layer.windowHandle
    }
})
```
takes 6ms on MacBook Pro 2019 2.3 GHz i9

igordmn (approved):

skiko/src/jvmMain/cpp/common/awt_jni.cc
```diff
@@ -36,10 +36,10 @@ extern "C"
         awt->FreeDrawingSurface(ds);
     }
 
-    JNIEXPORT void JNICALL Java_org_jetbrains_skiko_AWTKt_lockDrawingSurface(JNIEnv *env, jobject obj, jlong drawingSurfacePtr)
+    JNIEXPORT jint JNICALL Java_org_jetbrains_skiko_AWTKt_lockDrawingSurface(JNIEnv *env, jobject obj, jlong drawingSurfacePtr)
     {
         JAWT_DrawingSurface *ds = fromJavaPointer<JAWT_DrawingSurface *>(drawingSurfacePtr);
-        ds->Lock(ds);
+        return ds->Lock(ds) & JAWT_LOCK_ERROR;
```
olonho, line 42: `return (ds->Lock(ds) & JAWT_LOCK_ERROR) != 0;`, previous version was subtly buggy is some compilers.

skiko/src/jvmMain/objectiveC/macos/Drawlayer.mm
olonho: Not sure is 6ms is good enough for such a basic op, but anyway.

igordmn: 6ms take 100 calls of windowHandle

Rsedaikin (approved):

skiko/src/jvmMain/objectiveC/macos/Drawlayer.mm
Rsedaikin: > Let's remove layer.window and layerSet.window initialization inside nativeInit to avoid code duplication.
I think that finding NSWindow should only be implemented in one way

olonho (approved):

PR #99, opened 2021-04-29
Added OS/JRE info in debug output.

Rsedaikin:
Added in debug output info about current JDK and current OS name/version/architecture.


skiko/src/jvmMain/kotlin/org/jetbrains/skiko/context/Direct3DContextHandler.kt
```diff
@@ -109,8 +112,11 @@ internal class Direct3DContextHandler(layer: SkiaLayer) : ContextHandler(layer)
     }
 
     override fun hardwareInfo(): String {
-        return "DIRECT3D (dx12) hardware info:\n" +
+        return "DIRECT3D (dx12) rendering info:\n" +
```
olonho, line 115: Code seems to be mostly identical across renderers, why not unify it?

Rsedaikin: Done.

olonho (approved):

PR #104, opened 2021-05-11
Window state notifications.

olonho:


skiko/src/jvmMain/kotlin/org/jetbrains/skiko/SkiaLayer.kt
```diff
@@ -101,6 +108,19 @@ open class SkiaLayer(
         isInited = true
     }
 
+    private val stateHandlers =
+            mutableMapOf<PropertyKind, MutableList<(SkiaLayer) -> Unit>>()
+
+    fun onStateChanged(kind: PropertyKind, handler: (SkiaLayer) -> Unit) {
```
igordmn, line 114: 1. I think the simplier API would be not to use an enum class and a common listener.

Just `onRenderApiChanged {}` and `onContentScaleChanged {}`

2. Probably we don't need to pass `SkiaLayer` as an argument

olonho: We could do it either way, adding specific wrappers is easy. Passing `SkiaLayer` allow us no to remember layer we're care about in closure.

PR #112, opened 2021-05-21
Bump Skija 0.91.3 -> 0.91.6

tonsky:
This is needed in our new secret IDE

### DIFF
--- a/skiko/gradle.properties
+++ b/skiko/gradle.properties
@@ -2,7 +2,7 @@ kotlin.code.style=official
 deploy.version=0.0.0
 kotlin.native.version=1.5.30-dev-184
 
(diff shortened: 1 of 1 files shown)


olonho:
Generally better keep Skija mostly bindings to Skia API, and not an extending Skia library with own functionality. We could merge this change now, but maybe this code shall better be moved to extension functions in product that need them?

tonsky:
That’s the idea, yes. But CSS shadows are more like de-facto standard than app-specific functionality

tonsky:
How I can I use the new version?

olonho:
https://github.com/JetBrains/skiko/releases/tag/v0.2.34

PR #114, opened 2021-05-28
Fix window.layer.focusTraversalKeysEnabled = false

igordmn:
Disabling focus keys didn't work

Test:
```
    window.layer.focusTraversalKeysEnabled = false
    window.layer.addKeyListener(object : KeyAdapter() {
        override fun keyPressed(event: KeyEvent) {
            println("keyPressed")
        }
    })
```

Press Tab or Shift-Tab. Before fix there was no output in console

### DIFF
--- a/skiko/src/jvmMain/kotlin/org/jetbrains/skiko/SkiaLayer.kt
+++ b/skiko/src/jvmMain/kotlin/org/jetbrains/skiko/SkiaLayer.kt
@@ -8,6 +8,7 @@ import org.jetbrains.skija.*
 import org.jetbrains.skiko.context.ContextHandler
 import org.jetbrains.skiko.context.createContextHandler
 import org.jetbrains.skiko.redrawer.Redrawer
+import java.awt.AWTKeyStroke
 import java.awt.Graphics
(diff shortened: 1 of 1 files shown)


skiko/src/jvmMain/kotlin/org/jetbrains/skiko/SkiaLayer.kt
```diff
@@ -8,6 +8,7 @@ import org.jetbrains.skija.*
 import org.jetbrains.skiko.context.ContextHandler
 import org.jetbrains.skiko.context.createContextHandler
 import org.jetbrains.skiko.redrawer.Redrawer
+import java.awt.AWTKeyStroke
```
olonho, line 11: Why this import?

olonho (approved):

skiko/src/jvmMain/kotlin/org/jetbrains/skiko/SkiaLayer.kt
igordmn: Removed it, it is unused. Thanks!

PR #116, opened 2021-06-04
Remote rendering prototype (WIP)

olonho:
To allow easy embedding of Skiko apps into other context where we better not execute in-proc, for example in IDE, we need to be able to bind Skiko rendering to particular window, and closely follow its behavior (move, resize, focus change). 
As AWT only provides higher level to control windows behavior we extend Skiko with generic reparent API, which allows 
to bind out of process Skia window to the particular AWT frame. Actual implementation of this reparent is platform-dependent. This PR provides implementation for macOS using Accessibility API, which allows closely follow particular window behavior. Possible alternatives for macOS include using IOSurfaces API, but it is less generic, and can have problems with IDE process.

Also this PR provides sample client/server architecture implementation, which tests remote API in real-life scenario. So far it works decent on macOS, but it needs further work to better accommodate focus changes.


olonho:
No longer relevant.

natanfudge:
Out of curiosity, why is this no longer relevant?

PR #118, opened 2021-06-07
Support getting of system theme.

olonho:


skiko/src/jvmMain/kotlin/org/jetbrains/skiko/SystemTheme.kt
```diff
@@ -0,0 +1,17 @@
+package org.jetbrains.skiko
+
+enum class SystemTheme {
+    DARK,
+    LIGHT,
+    UNKNOWN
+}
+
+val currentSystemTheme: SystemTheme
```
igordmn, line 9: Can we write something like `addCurrentSystemThemeListener"?

olonho: As discussed, let's use generic configuration callback.

PR #128, opened 2021-06-22
Fix linux dpi

Rsedaikin:
Force enable dpi support on Linux. To achieve this, we must set the system property "sun.java2d.uiScale" before initializing the AWT/Swing.

### DIFF
--- a/skiko/src/jvmMain/kotlin/org/jetbrains/skiko/PlatformOperations.kt
+++ b/skiko/src/jvmMain/kotlin/org/jetbrains/skiko/PlatformOperations.kt
@@ -99,18 +99,6 @@ internal val platformOperations: PlatformOperations by lazy {
 
                 override fun getDpiScale(component: Component): Float {
                     return component.graphicsConfiguration.defaultTransform.scaleX.toFloat()
-                    // TODO doesn't work well because java doesn't scale windows (content has offset with 200% scale)
-                    //
-                    // Two solutions:
-                    // 1. dynamically change sun.java2d.uiScale (it is global property,
-                    // so we have to be careful) and update all windows
-                    //
-                    // 2. apply contentScale manually to all windows
-                    // (it is not good, because on different platform windows will have different size.
-                    // Maybe we will apply contentScale manually on all platforms?)
-
-                    // see also comment for HardwareLayer.checkContentScale
-                    // return (component as HardwareLayer).useDrawingSurfacePlatformInfo(::linuxGetDpiScaleNative)
                 }
 
                 override fun createRedrawer(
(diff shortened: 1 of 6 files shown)


skiko/src/jvmMain/kotlin/org/jetbrains/skiko/Setup.kt
```diff
@@ -24,4 +24,14 @@ object Setup {
             // Not all platforms allow this.
         }
     }
-}
\ No newline at end of file
+}
+
+fun tryEnableUIScaling() {
```
igordmn, line 29: Lets move all global properties initialization in one place (`Setup.init`):

1. Move this to https://github.com/JetBrains/skiko/blob/master/skiko/src/jvmMain/kotlin/org/jetbrains/skiko/Setup.kt. 

2. Change signature of `Setup.init`:
```
object Setup {
    fun init(
        noEraseBackground: Boolean = System.getProperty("skiko.rendering.noerasebackground") != "false",
        globalLAF: Boolean = System.getProperty("skiko.rendering.laf.global") == "true",
        useScreenMenuBar: Boolean = System.getProperty("skiko.rendering.useScreenMenuBar") != "false",
        autoLinuxDpi = System.getProperty("skiko.linux.autodpi") == "true"
    ) {
```

We can call it [here](https://cs.android.com/androidx/platform/frameworks/support/+/androidx-main:compose/ui/ui/src/desktopMain/kotlin/androidx/compose/ui/window/Application.desktop.kt;l=160?q=Application%20desktop), and in `AppWindow` (`companion { init {} }`), before any Swing initialization.

Probably we can try to call it in `companion { init {} }` block of `ComposeWindow` / `ComposeDialog`. But don't know if it will work.

`skiko.linux.autodpi` will be set automatically in our Gradle plugin. in IDEA plugin application or other Swing application we shouldn't override it (when our Gradle plugin isn't used)

skiko/src/jvmMain/kotlin/org/jetbrains/skiko/Setup.kt
```diff
@@ -23,5 +30,8 @@ object Setup {
         } catch (e: UnsupportedOperationException) {
             // Not all platforms allow this.
         }
+        isInited = true
```
igordmn, line 33: Lets remove this, because we don't use it

skiko/src/jvmMain/kotlin/org/jetbrains/skiko/PlatformOperations.kt
```diff
@@ -110,7 +110,7 @@ internal val platformOperations: PlatformOperations by lazy {
                     // Maybe we will apply contentScale manually on all platforms?)
 
                     // see also comment for HardwareLayer.checkContentScale
-                    // return component.useDrawingSurfacePlatformInfo(::linuxGetDpiScaleNative)
+                    // return (component as HardwareLayer).useDrawingSurfacePlatformInfo(::linuxGetDpiScaleNative)
```
igordmn, line 113: Lets remove this comment, because it is obsolete now

igordmn (approved):

PR #134, opened 2021-07-14
Add Dockerfile for Linux arm64 image

AlexeyTsvetkov:


olonho (approved):
OK, but let's use standard openjdk package.

PR #141, opened 2021-07-23
Create automated background memory manager.

olonho:


skiko/src/jvmMain/kotlin/org/jetbrains/skiko/FrameWatcher.kt
```diff
@@ -0,0 +1,35 @@
...
+ * making sure that we'll have memory consumption under control.
+ */
+internal object FrameWatcher {
+    fun start() {
+        // We initiate GC on IO threads, so that rendering is not blocked in
+        // cases of concurrent collectors.
+        @OptIn(DelicateCoroutinesApi::class)
+        GlobalScope.launch(Dispatchers.IO) {
+            while (true) {
+                delay(5000L)
```
igordmn, line 21: Maybe instead of delay we will use something like frameCountChannel?

So there will be no background daemon eating CPU (not sure how it affects battery on laptops)

Rsedaikin: We could use `wait() / notify()` instead of `delay` and inside the `nextFrame()` function check the counter and notify if `frameCounter > 1000.

olonho: Guess with this kind of update frequency it's not too important which mechanism will be used. 30 seconds is not very frequent. Will merge current solution to make a build for customers, and then can work on improvements, if will be needed.

PR #144, opened 2021-08-04
Fix fallback

Rsedaikin:
Fixed fallback operation on virtual machines (VMWare, VirtualBox)
- https://youtrack.jetbrains.com/issue/TBX-6467
- https://youtrack.jetbrains.com/issue/TBX-6514
and other cases when OpenGL is falled back and an exception is thrown in the SwapBuffers or setSwapInterval GL functions.


skiko/src/jvmMain/cpp/common/openglapi.cc
```diff
@@ -1,50 +1,81 @@
...
+#import <OpenGL/gl3.h>
+#else
+#include <GL/gl.h>
+#endif
+#include <jni.h>
+#include <string>
+#include <vector>
+
+const std::vector<std::string> vendorBlacklist{ "VMware, Inc." };
+const std::vector<std::string> adapterBlacklist{
```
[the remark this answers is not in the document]
igordmn: For example, why only LLVM 5.0 and 11.0.1?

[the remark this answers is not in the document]
igordmn: I think we can use blacklist, but only if we can't do that:
> Some function returned not 0, or version of DirectX isn't supported, etc.

And we should be 99% sure, that it is right to add card to the blacklist: either test it at least on the second machine, or find it in mainstream apps like Firefox (https://wiki.mozilla.org/Blocklisting/Blocked_Graphics_Drivers) or Chromium (https://chromium.googlesource.com/chromium/src/gpu/+/refs/heads/main/config/software_rendering_list.json)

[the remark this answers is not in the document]
igordmn: >https://chromium.googlesource.com/chromium/src/gpu/+/refs/heads/main/config/software_rendering_list.json

Every item in this list has a reason why it was blacklisted:
- GPU rasterization on CrOS is blocklisted on anything but Intel, Mali T8/G, Imagination, Freedreno, or AMD GPU/drivers for now."
- VideoCore V has corrupt rendering with GPU Rasterization on Android versions below N
- NVidia drivers older than 2010 on Windows are possibly unreliable

skiko/src/jvmMain/cpp/windows/directx.cc
```diff
@@ -0,0 +1,70 @@
...
+#include "SkSurface.h"
+
+#include "d3d/GrD3DTypes.h"
+#include <d3d12sdklayers.h>
+#include "d3d/GrD3DBackendContext.h"
+#include <d3d12.h>
+#include <dxgi1_4.h>
+#include <dxgi1_6.h>
+
+UINT adapterIndex = 0;
```
igordmn, line 19: I am not sure about this global variable, will it be thread-safe, or work with multiple SkiaLayer's

Can we get rid of it? It will much safer and cleaner. For example:

```
bool defineAdapter(IDXGIFactory4 *pFactory, IDXGIAdapter1 **ppAdapter, int* index)

Java_org_jetbrains_skiko_GraphicsApiKt_getNextDirectXAdapter(oldIndex: Int): Int
```

skiko/src/jvmMain/kotlin/org/jetbrains/skiko/GraphicsApi.kt
```diff
@@ -10,4 +14,45 @@ enum class GpuPriority(val value: String) {
...
+
+internal fun isVideoCardSupported(renderApi: GraphicsApi): Boolean {
+    return when (renderApi) {
+        GraphicsApi.DIRECT3D -> {
+            val adaptersList = notSupportedAdapters.filter { it.startsWith("directx:") }.map {
+                it.replace("directx:", "")
+            }
+            var adapter = getNextDirectXAdapter()
+            while (adapter != null) {
+                adaptersList.forEach {
```
igordmn, line 32: Not sure about performance, can we replace it by Set?

```
private val notSupportedAdapters: Set<String> by lazy {
    val resource = SkiaLayer::class.java.getResource("/not-supported-adapter.list").readText()
    resource.split(";").mapTo(HashSet()) { it.trim() }.first()
}

internal fun isVideoCardSupported(renderApi: GraphicsApi): Boolean {
    return when (renderApi) {
        GraphicsApi.DIRECT3D -> {
            while (adapter != null) {
                if (notSupportedAdapters.contains("directx:$adapter") {
                    return false
                }
                adapter = getNextDirectXAdapter()
            }
            true
        }
```

skiko/src/jvmMain/kotlin/org/jetbrains/skiko/GraphicsApi.kt
```diff
@@ -10,4 +14,45 @@ enum class GpuPriority(val value: String) {
...
+                    if (adapter!!.startsWith(it)) {
+                        return false
+                    }
+                }
+                adapter = getNextDirectXAdapter()
```
igordmn, line 37: If the first adapter is supported, and second isn't - we switch to Software

Maybe we should just check the first adapter?

But I believe we should use the same logic as in defineHardwareAdapter, or better not separate them into two places.

If we need to check adapter as soon as possible, we shouldn't write the same code twice - we should call createDevice as soon as possible. The dispose problem (see above) will also be fixed if we will do that.

skiko/src/jvmMain/cpp/windows/directx.cc
```diff
@@ -0,0 +1,70 @@
...
+bool defineAdapter(IDXGIFactory4 *pFactory, IDXGIAdapter1 **ppAdapter)
+{
+    *ppAdapter = nullptr;
+    IDXGIAdapter1 *pAdapter = nullptr;
+    if (DXGI_ERROR_NOT_FOUND == pFactory->EnumAdapters1(adapterIndex, &pAdapter))
+    {
+        adapterIndex = 0;
+        return false;
+    }
+    if (SUCCEEDED(D3D12CreateDevice(pAdapter, D3D_FEATURE_LEVEL_11_0, _uuidof(ID3D12Device), nullptr)))
```
igordmn, line 32: We create divide but never dispose it

igordmn: Or I was mistaken, everything is disposed? :) 

> pAdapter->Release();

skiko/src/jvmMain/resources/not-supported-adapter.list
```diff
@@ -0,0 +1,9 @@
+directx:Intel(R) HD Graphics 520;
```
olonho, line 1: Please add human readable reason as last field, and show it when adapter is disabled in error message.

PR #148, opened 2021-08-23
Emulate vsync tick on Software renderer and Linux with unsupported vsync

igordmn:
Otherwise there is no delay between frames and we will have high CPU usage

### DIFF
--- a/.github/workflows/blank.yml
+++ b/.github/workflows/blank.yml
@@ -58,7 +58,7 @@ jobs:
           sudo apt-get install gcc-9 g++-9 -y
           sudo update-alternatives --install /usr/bin/gcc gcc /usr/bin/gcc-9 60 --slave /usr/bin/g++ g++ /usr/bin/g++-9
           sudo update-alternatives --config gcc 
-          sudo apt-get install ninja-build fontconfig libfontconfig1-dev libglu1-mesa-dev zip -y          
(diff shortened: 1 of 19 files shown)


olonho (approved):

skiko/build.gradle.kts
```diff
@@ -368,6 +368,7 @@ tasks.withType(LinkSharedLibrary::class.java).configureEach {
                     "-static-libgcc",
                     "-lGL",
                     "-lfontconfig",
+                    "-lXrandr",  // requires "apt-get install libxrandr-dev" to get headers
```
olonho, line 371: Please, let's not add another dependency.

.github/workflows/blank.yml
```diff
@@ -58,7 +58,7 @@ jobs:
           sudo apt-get install gcc-9 g++-9 -y
           sudo update-alternatives --install /usr/bin/gcc gcc /usr/bin/gcc-9 60 --slave /usr/bin/g++ g++ /usr/bin/g++-9
           sudo update-alternatives --config gcc 
-          sudo apt-get install ninja-build fontconfig libfontconfig1-dev libglu1-mesa-dev zip -y          
+          sudo apt-get install ninja-build fontconfig libfontconfig1-dev libglu1-mesa-dev libxrandr-dev zip -y
```
olonho, line 61: We shall add same dependency to skiko/docker/* and rebuild images on CI.

Issue #137, opened 2021-07-21
Don't call renderer.onRender if window is hidden or minimized


igordmn:
If window is hidden there is no need to update/draw content. If Application needs some backgrounds tasks - it should run them explicitly and shouldn't use rendering loop for that.

For example, Compose, that used skiko, can have animations in the window, and we should automatically stop them if the user will minimize the window.

igordmn:
In https://github.com/JetBrains/skiko/pull/149 rendering isn't called if window is invisible (i.e. doesn't appear on desktop/taskbar).

To stop rendering when the window is overlapped by other windows (occluded) or minimized, we should use graphic/platform specific API ([macOs](https://developer.apple.com/documentation/appkit/nswindowocclusionstate/nswindowocclusionstatevisible), [DirectX](https://docs.microsoft.com/en-us/windows/win32/direct3ddxgi/waiting-when-occluded)). If we just check if the window is minimized - it covers only half of the cases and probably doesn't worth to implement.

Thomas-Vos:
> To stop rendering when the window is overlapped by other windows (occluded) or minimized, we should use graphic/platform specific API ([macOs](https://developer.apple.com/documentation/appkit/nswindowocclusionstate/nswindowocclusionstatevisible), [DirectX](https://docs.microsoft.com/en-us/windows/win32/direct3ddxgi/waiting-when-occluded)).

It would be amazing if this could be implemented. The app I am working on is meant to be active in the background while the user is using other apps/games. This feature would reduce CPU usage in that case, so my app would really benefit from it.

okushnikov:
Please check the <a href="https://youtrack.jetbrains.com/issue/SKIKO-137">following ticket</a> on YouTrack for follow-ups to this issue. GitHub issues will be closed in the coming weeks.

PR #149, opened 2021-08-24
Don't draw if window is invisible (JFrame.isVisible = false)

igordmn:
Partially fix https://github.com/JetBrains/skiko/issues/137

### DIFF
--- a/skiko/src/jvmMain/kotlin/org/jetbrains/skiko/redrawer/MetalRedrawer.kt
+++ b/skiko/src/jvmMain/kotlin/org/jetbrains/skiko/redrawer/MetalRedrawer.kt
@@ -9,7 +9,6 @@ import org.jetbrains.skia.DirectContext
 import org.jetbrains.skiko.*
 import javax.swing.SwingUtilities.convertPoint
 import javax.swing.SwingUtilities.getRootPane
-import kotlin.time.ExperimentalTime
 
 internal class MetalRedrawer(
     private val layer: SkiaLayer,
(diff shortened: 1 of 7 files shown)


olonho (approved):

skiko/src/jvmMain/kotlin/org/jetbrains/skiko/redrawer/MetalRedrawer.kt
```diff
@@ -28,8 +32,10 @@ internal class MetalRedrawer(
     private val windowHandle = layer.windowHandle
 
     private val frameDispatcher = FrameDispatcher(Dispatchers.Swing) {
-        update(System.nanoTime())
-        draw()
+        if (layer.isShowing) {
```
olonho, line 35: How expensive is isShowing check?

igordmn: It checks a boolean property recursively through all parents.

With 1000 deeply nested components (30 level-deep) we have 0.5ms.

Let's cached it then.

skiko/src/jvmMain/kotlin/org/jetbrains/skiko/redrawer/MetalRedrawer.kt
```diff
@@ -28,8 +32,10 @@ internal class MetalRedrawer(
...
+            update(System.nanoTime())
```
igordmn, line 36: I just realized that this change conflicts with `awaitRedraw` method. `awaitRedraw` should draw into the native buffer, even if window is invisible.

Probably this change is okay, but `awaitRedraw` should be changed. I also realized why it was slow on macOs (because of the delay(300) when the window is occluded)

Let's discuss it next week.

PR #150, opened 2021-08-31
Remove Skija, prepare for truly MPP Skiko

olonho:


skiko/src/jvmMain/kotlin/org/jetbrains/skija/RuntimeEffect.kt
```diff
@@ -1,12 +1,11 @@
 package org.jetbrains.skija
 
 import org.jetbrains.skija.impl.Library.Companion.staticLoad
-import org.jetbrains.annotations.ApiStatus
 import org.jetbrains.skija.impl.RefCnt
 import org.jetbrains.skija.impl.Native
 import org.jetbrains.skija.impl.Stats
 
-class RuntimeEffect @ApiStatus.Internal constructor(ptr: Long) : RefCnt(ptr) {
```
igordmn: Skija was designed that it exposes all [private code as public](https://github.com/JetBrains/skija/blob/master/docs/Conventions.md#visibility), so if we forget to implement something, users can implement it themselves.

Do we want to change this rule?

olonho: As we are both vendors and users of Skiko, not sure if this is relevant in our case.

PR #184, opened 2021-09-12
JS artifact compilation.

olonho:


chris-hatton:
This PR makes me hot under the collar.

PR #194, opened 2021-09-15
Introduce minimal initialization layer for using it as API on the frontend side

Schahen:
Basically whenever we need to work with skia we just make sure that it's loaded via calling `onWasmReady`


samples/SkiaJsSample/src/jsMain/kotlin/org/jetbrains/skiko/sample/js/App.kt
```diff
@@ -1,7 +1,10 @@
 package org.jetbrains.skiko.sample.js
 
 import org.jetbrains.skia.Paint
+import org.jetbrains.skiko.wasm.onWasmReady
 
 fun main(args: Array<String>) {
-    println(Paint())
+    onWasmReady {
```
olonho, line 7: Could we make `composeMain()` an entry point after Wasm is initialized?

PR #195, opened 2021-09-15
interopScope {} for all functions receiving primitive arrays

alexander-gorshenev:
but excluding functions returning non-primitive types

### DIFF
--- a/skiko/src/commonMain/kotlin/org/jetbrains/skia/Bitmap.kt
+++ b/skiko/src/commonMain/kotlin/org/jetbrains/skia/Bitmap.kt
@@ -1,12 +1,8 @@
 @file:Suppress("NESTED_EXTERNAL_DECLARATION")
 package org.jetbrains.skia
 
+import org.jetbrains.skia.impl.*
 import org.jetbrains.skia.impl.Library.Companion.staticLoad
-import org.jetbrains.skia.impl.Managed
-import org.jetbrains.skia.impl.Stats
-import org.jetbrains.skia.impl.reachabilityBarrier
-import org.jetbrains.skia.impl.NativePointer
-import org.jetbrains.skia.impl.getPtr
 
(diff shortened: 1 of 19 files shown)


skiko/src/commonMain/kotlin/org/jetbrains/skia/Bitmap.kt
```diff
@@ -564,16 +560,18 @@ class Bitmap internal constructor(ptr: NativePointer) : Managed(ptr, _FinalizerH
...
-                _ptr,
-                info.width,
-                info.height,
-                info.colorInfo.colorType.ordinal,
-                info.colorInfo.alphaType.ordinal,
-                getPtr(info.colorInfo.colorSpace),
-                pixels,
-                rowBytes
-            )
+            interopScope {
```
olonho, line 563: To avoid unneeded memory allocations we probably shall make interopScope expect, and actualize it as 

```
object theScope: InteropScope
inline fun <T> interopScope(block: InteropScope.() -> T): T {
       return theScope.block()
    }
}
```

skiko/src/commonMain/kotlin/org/jetbrains/skia/TextBlobBuilder.kt
```diff
@@ -264,7 +266,7 @@ private external fun _nAppendRunPosH(
 
 
 @ExternalSymbolName("org_jetbrains_skia_TextBlobBuilder__1nAppendRunPos")
-private external fun _nAppendRunPos(ptr: NativePointer, fontPtr: NativePointer, glyphs: ShortArray?, pos: FloatArray?, bounds: Rect?)
+private external fun _nAppendRunPos(ptr: NativePointer, fontPtr: NativePointer, glyphs: InteropPointer, pos: InteropPointer, bounds: Rect?)
```
olonho, line 269: How could we pass `Rect?` still?

olonho (commented):
Looks OK, let's fix interopScope to make it less intrusive on JVM.

olonho (approved):

PR #201, opened 2021-09-17
Publish kotlin metadata artefact

alexander-gorshenev:


skiko/buildSrc/src/main/kotlin/properties.kt
```diff
@@ -138,6 +138,7 @@ class SkikoProperties(private val myProject: Project) {
 object SkikoArtifacts {
     // names are also used in samples, e.g. samples/SkijaInjectSample/build.gradle
     val commonArtifactId = "skiko-jvm"
```
olonho, line 140: Better rename to `commonJvmArtifactId`.

olonho (approved):

PR #211, opened 2021-09-22
Fix deadlock in FrameLimiter

igordmn:
We run FrameLimiter in another thread in SoftwareRedrawer. So FrameLimiter should be thread-safe. It is not thread safe now.

We can run it from the Swing thread though, but probably it will be used in Compose directly, where can be multiple threads (in Application frame clock, for example)

Deadlock:
[couroutine 2] onRequest.trySend(Unit)
[couroutine 1] onRequest.receive()
[couroutine 1] onResult.getAndSet(CompletableDeferred()).complete(value)
[couroutine 2] onResult.get().await() // wait
[couroutine 1] onRequest.receive()  // wait

### DIFF
--- a/skiko/src/jvmMain/kotlin/org/jetbrains/skiko/RendezvousBroadcastChannel.kt
+++ b/skiko/src/jvmMain/kotlin/org/jetbrains/skiko/RendezvousBroadcastChannel.kt
@@ -1,30 +1,45 @@
 package org.jetbrains.skiko
 
-import kotlinx.coroutines.CompletableDeferred
 import kotlinx.coroutines.channels.Channel
-import java.util.concurrent.atomic.AtomicReference
+import kotlinx.coroutines.suspendCancellableCoroutine
+import kotlin.coroutines.Continuation
+import kotlin.coroutines.resume
 
 /**
  * Behaves as Channel<Unit>(Channel.RENDEZVOUS), but with ability to send value to all current consumers
- * (which await on `receive` method)
+ * (which await on `receive` method).
  */
 internal class RendezvousBroadcastChannel<T> {
     private val onRequest = Channel<Unit>(Channel.CONFLATED)
-    private val onResult = AtomicReference(CompletableDeferred<T>())
+    private val receivers = mutableListOf<Continuation<T>>()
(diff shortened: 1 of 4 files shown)


skiko/src/jvmMain/kotlin/org/jetbrains/skiko/RendezvousBroadcastChannel.kt
```diff
@@ -1,30 +1,45 @@
...
 /**
  * Behaves as Channel<Unit>(Channel.RENDEZVOUS), but with ability to send value to all current consumers
- * (which await on `receive` method)
+ * (which await on `receive` method).
  */
 internal class RendezvousBroadcastChannel<T> {
     private val onRequest = Channel<Unit>(Channel.CONFLATED)
-    private val onResult = AtomicReference(CompletableDeferred<T>())
+    private val receivers = mutableListOf<Continuation<T>>()
+    private val receiversCopy = mutableListOf<Continuation<T>>()
```
olonho, line 15: Why make it a field, not a local? It would mean two threads in send all will race on copy.

skiko/src/jvmTest/kotlin/org/jetbrains/skiko/RendezvousBroadcastChannelTest.kt
```diff
@@ -71,6 +66,30 @@ class RendezvousBroadcastChannelTest {
...
+        runBlocking {
+            val receiverJobs = (1..10).map {
+                launch(Dispatchers.IO) {
+                    repeat(1000) {
+                        channel.receive()
+                    }
+                }
+            }
+
+            val sendingJob = launch(Dispatchers.IO) {
```
olonho, line 82: There shall be several concurrent senders in test as well.

olonho (approved):

skiko/src/jvmTest/kotlin/org/jetbrains/skiko/RendezvousBroadcastChannelTest.kt
igordmn: The class is internal, and we don't use it in such case

Issue #214, opened 2021-09-24
Deadlock in MetalRedrawer


ignatov:
Found one Java-level deadlock:
=============================
"DefaultDispatcher-worker-2":
  waiting to lock monitor 0x00007fe250048400 (object 0x00000007f0950b18, a java.awt.Component$AWTTreeLock),
  which is held by "AWT-EventQueue-0"
"AWT-EventQueue-0":
  waiting to lock monitor 0x00007fe23801f500 (object 0x00000007f18c8a38, a java.lang.Object),
  which is held by "DefaultDispatcher-worker-2"

Java stack information for the threads listed above:
===================================================
"DefaultDispatcher-worker-2":
	at java.awt.Component.getLocationOnScreen(java.desktop@11.0.12/Unknown Source)
	- waiting to lock <0x00000007f0950b18> (a java.awt.Component$AWTTreeLock)
	at javax.swing.SwingUtilities.convertPointToScreen(java.desktop@11.0.12/Unknown Source)
	at javax.swing.SwingUtilities.convertPoint(java.desktop@11.0.12/Unknown Source)
	at javax.swing.SwingUtilities.convertPoint(java.desktop@11.0.12/Unknown Source)
	at org.jetbrains.skiko.redrawer.MetalRedrawer.syncSize(MetalRedrawer.kt:97)
	at org.jetbrains.skiko.context.MetalContextHandler.initContext(MetalContextHandler.kt:18)
	at org.jetbrains.skiko.SkiaLayer.prepareDrawContext$skiko(SkiaLayer.kt:338)
	at org.jetbrains.skiko.redrawer.MetalRedrawer.performDraw(MetalRedrawer.kt:89)
	- locked <0x00000007f18c8a38> (a java.lang.Object)
	at org.jetbrains.skiko.redrawer.MetalRedrawer.access$performDraw(MetalRedrawer.kt:13)
	at org.jetbrains.skiko.redrawer.MetalRedrawer$draw$2.invokeSuspend(MetalRedrawer.kt:77)
	at kotlin.coroutines.jvm.internal.BaseContinuationImpl.resumeWith(ContinuationImpl.kt:33)
	at kotlinx.coroutines.DispatchedTask.run(DispatchedTask.kt:106)
	at kotlinx.coroutines.scheduling.CoroutineScheduler.runSafely(CoroutineScheduler.kt:571)
[...]

PR #217, opened 2021-09-27
Fix metal deadlock

igordmn:
1. Fix deadlock on macOs (fixes https://github.com/JetBrains/skiko/issues/214)

[RendertingThread]
```
MetalRedrawer.performDraw
synchronized(drawLock)
MetalRedrawer.syncSize
java.awt.Component.getLocationOnScreen
synchonized(component.awtLock)
```

[AWTThread]
```
synchonized(component.awtLock)
SkiaLayer.paint
MetalRedrawer.redrawImmediately
synchronized(drawLock)
```

To fix this, we move syncSize to another place, which should be called in AWT thread:
- setBounds
- findNextWorkingRenderApi

2. Now we only catch rendering exceptions, not all exceptions. Before, any exception happened in the user code (in SkiaRendere.onRender) was treated as rendering exception. We shouldn't fallback to another rendering API in this case.

3. We catch exceptions, not only happened in `paint` method, but also if they happened in FrameDispatcher loop

4. Fallback tests

5. Remove OpenGL for macOs


skiko/src/jvmMain/kotlin/org/jetbrains/skiko/PlatformOperations.kt
```diff
@@ -11,6 +10,8 @@ import java.awt.Component
 import java.awt.Window
 import javax.swing.SwingUtilities
 
+internal var testNonSoftwareRedrawer: ((SkiaLayer) -> Redrawer)? = null
```
olonho, line 13: This kind of top level variable looks somewhat ugly. Why cannot we keep this kind of state out of the top level?

igordmn: Moved it into factory

olonho (approved):

Rsedaikin (approved):

PR #219, opened 2021-09-28
Publish skiko as mpp library

alexander-gorshenev:


skiko/build.gradle.kts
```diff
@@ -741,8 +741,13 @@ tasks.withType<Test>().configureEach {
     }
 }
 
-fun Publication.isSkikoPublication() =
-    (this as? MavenPublication)?.name?.startsWith("skiko") == true
+fun Publication.isSkikoPublication(): Boolean {
```
olonho, line 744: Not sure if I like it, looks very hacky. Can we rethink it?

skiko/gradle.properties
```diff
@@ -1,6 +1,9 @@
 kotlin.code.style=official
 deploy.version=0.0.0
 kotlin.native.version=1.5.21-dev-329
+kotlin.mpp.enableCompatibilityMetadataVariant=true
```
olonho, line 4: Could you please comment what it does, better with link on docs.

alexander-gorshenev: It allows an hmpp library to be used in non-hmpp project.

In addition to `metadata only klib jar` in the root component it places the old  `*.kotlin_metadata` jar.

The flag was recommended in the 1.4 migration doc 
https://kotlinlang.org/docs/migrating-multiplatform-project-to-14.html#migrate-to-the-hierarchical-project-structure 

The flag seems to be true by default in the currently used compiler already, 
so no need to explicitly specify it anymore. 

I pushed a commit to remove it.

olonho (approved):
Looks good, but let's CI finish.

PR #221, opened 2021-09-28
Introduce minimal viable api for renderer in web, update web sample

Schahen:


skiko/src/jsMain/kotlin/org/jetbrains/skiko/wasm/api/CanvasRenderer.kt
```diff
@@ -0,0 +1,48 @@
...
+
+    init {
+        GL.makeContextCurrent(contextPointer)
+    }
+
+    val surface = Surface.createFromGL(width, height)
+    val canvas: Canvas
+        get() = surface.canvas
+
+    abstract fun drawFrame(currentTimestamp: Double)
```
olonho, line 38: Maybe make `
class CanvasRenderer constructor(htmlCanvas: HTMLCanvasElement, val width: Int, val height: Int, val drawFrame: (Double) -> Unit)` instead to avoid unneeded subclassing?

samples/SkiaJsSample/src/jsMain/kotlin/org/jetbrains/skiko/sample/js/App.kt
```diff
@@ -2,46 +2,127 @@ package org.jetbrains.skiko.sample.js
...
+    val circle: Circle,
+    val width: Int,
+    val height: Int,
+    val velocity: Float,
+    var angle: Double
+) {
+    fun recalculate(dt: Float) {
+        val position = calculatePosition(circle, width, height)
+
+        when (position) {
```
olonho, line 61: Make when exhaustive.

samples/SkiaJsSample/src/jsMain/kotlin/org/jetbrains/skiko/sample/js/App.kt
```diff
@@ -2,46 +2,127 @@ package org.jetbrains.skiko.sample.js
...
+}
+
+private class BouncingBallsApp(htmlCanvas: HTMLCanvasElement, private var prevTimestamp: Double) : CanvasRenderer(htmlCanvas) {
+    private fun Color4f.asPaint() = Paint().apply {
+        color4f = this@asPaint
+        mode = PaintMode.FILL
+        isAntiAlias = true
+    }
+
+    val data = listOf(
```
olonho, line 83: After switching tabs in browser rendering breaks, maybe canvas size is miscomputed or GL context gone.

skiko/src/jsMain/kotlin/org/jetbrains/skiko/wasm/api/CanvasRenderer.kt
Schahen: This would be  a quite javascript thing to do, however my arguments for not doing so are following:
1) In future we might need more than one method to implement (for instance, needsRedraw)
2) By the very nature of  usage of such renderer there most likely be a data which is not supposed to be calculated in onDraw  - that said, in real-life scenarious there still will be some class/object to work with.

samples/SkiaJsSample/src/jsMain/kotlin/org/jetbrains/skiko/sample/js/App.kt
Schahen: It looks like this is something that browser does intentionally  to save some energy from not running the process. This needs to be addressed but this particular PR neither introduced, nor, unfrotunately fixed this issue (

samples/SkiaJsSample/src/jsMain/kotlin/org/jetbrains/skiko/sample/js/App.kt
Schahen: well, I can add else or check for inside but in this case angle just doesn't change at all, we basically do nothing in that case.

olonho: Kotlin compiler warns on non-exhaustive checks, enough to add INSIDE.

samples/SkiaJsSample/src/jsMain/kotlin/org/jetbrains/skiko/sample/js/App.kt
olonho: Fixed, was not entirely correct bouncing logic for large deltas.

olonho (approved):

PR #222, opened 2021-09-28
Skiko interop scope tests

eymar:


skiko/build.gradle.kts
```diff
@@ -348,10 +351,12 @@ val wasmCompile = project.tasks.register<Exec>("wasmCompile") {
         dependsOn(skiaWasmDir)
         val skiaDir = skiaWasmDir.get().absolutePath
         val outDir = "$buildDir/wasm"
-        val srcs = listOf(
+        val srcs = mutableListOf(
             "$projectDir/src/jsMain/cpp",
             "$projectDir/src/commonMain/cpp"
-        ).findAllFiles(".cc").toTypedArray()
+        ).apply {
+            if (skiko.includeTestHelpers) add("$projectDir/src/commonTest/cpp/TestHelpers.cc")
```
olonho, line 358: Why conditionally? Cannot we have two .wasm artefacts?

skiko/src/commonTest/cpp/TestHelpers.cc
```diff
@@ -0,0 +1,19 @@
+#include "common.h"
+
+SKIKO_EXPORT void org_jetbrains_skiko_tests_TestHelpers__1nFillByteArrayOf5(KNativePointer byteArray) {
+    char *bytes = reinterpret_cast<char*>(byteArray);
```
olonho, line 4: Let's make char -> uint8_t.

skiko/src/commonTest/cpp/TestHelpers.cc
```diff
@@ -0,0 +1,19 @@
...
+    bytes[0] = 1;
+    bytes[1] = 2;
+    bytes[2] = 3;
+    bytes[3] = 4;
+    bytes[4] = 5;
+}
+
+SKIKO_EXPORT void org_jetbrains_skiko_tests_TestHelpers__1nFillFloatArrayOf5(KNativePointer floatArray) {
```
olonho, line 12: Shalln't we test all types here (ShortArray, CharArray)?

olonho (commented):
Looks neat!

skiko/src/commonTest/cpp/TestHelpers.cc
eymar: fixed

skiko/build.gradle.kts
eymar: Yes, we can have skiko-test-helpers.wasm to include it in js tests. We also can have skiko-test-helpers.a and include it in native tests.
I didn't manage to build a separate lib for jvm tests though (couldn't configure CppLibrary to make it build two separate libs (skiko and skiko-test-helpers). Not sure if it's possible within the same module). 

I didn't add those changes for wasm and native in this PR now, since I don't have such a change for jvm yet. I plan to figure it out in a separate branch.

skiko/src/commonTest/cpp/TestHelpers.cc
eymar: Sure. Added tests for ints, doubles, shorts.

CharArray is not clear yet, since there is no toInterop(charArray). Although there is `toInterop(string)` but it's not the same on all platforms.
So maybe we need to have all pairs of methods (toInterop, fromInterop) for every type? Although not many of existing methods are actually used in the skiko code.

olonho:
Looks great, please resolve conflicts and let's merge it.

PR #224, opened 2021-09-28
Port JVM tests from skija 

Schahen:
A few things to notice:

1) So far I've ported them to the jvmTarget to have them ASAP in our codebase, as soon as they be commonized - they end up being in the skija subpackage (at least this is the suggestion)

2) The only test failing was surprisingly PaintTest - I've commented out failing tests and want to discuss them with the team. 

3) It looks like ImageTests test nothing but rather creates resources (that supposedly should be checked manually) - this is the logic from skija and should be addressed later when we'll have reference screenshotes to check against.

4) Apart from PaintTest with commented out tests the only thing I removed are two linux-specic snippets of code that looked like following:
```
if (Platform.CURRENT !== Platform.LINUX) assertContentEquals(
            FontVariation.parse("wght=400 slnt=0"),
            interV.variations
        )
```


olonho:
Are those tests actually run by CI?

Schahen:
@olonho if we are building the skiko project (the skiko/skiko I mean)  and not switching off explicitly the "test" task the answer is yes - since jvmTest is a part of test phase.

skiko/src/jvmTest/kotlin/org/jetbrains/skiko/BitmapTest.kt
```diff
@@ -0,0 +1,46 @@
+package org.jetbrains.skiko
```
olonho, line 1: Shalln't it be in just `org.jetbrains.skia` package? Would simplify imports a lot. Not blocking, ofc.

olonho (approved):

skiko/src/jvmTest/kotlin/org/jetbrains/skiko/ColorTest.kt
```diff
@@ -0,0 +1,35 @@
...
+import kotlin.test.Test
+
+import kotlin.test.assertEquals
+
+class ColorTest {
+    @Test
+    fun testColor() {
+        val cases = mapOf(
+        0x00000000 to Color4f(0f, 0f, 0f, 0f),
+        -0x1000000 to Color4f(0f, 0f, 0f, 1f),
```
olonho, line 14: Negated hex number looks odd, can we use Long or UInt here?

PR #228, opened 2021-09-30
Post processing skiko.js via script passed to emcc

Schahen:
The emcc compiler itself has a postprocessor. I suggest to use it as a slightly better solution compared to introducing a separate gradle task and / or postprocessing while adding to the jar itself.


olonho (approved):

Schahen:
After discussing with @AlexeyTsvetkov we've decided to rather adopt his changes

PR #229, opened 2021-10-01
Use codesign tool instead of curl

olonho:


skiko/build.gradle.kts
```diff
@@ -559,6 +559,35 @@ fun remoteSign(signHost: String, lib: File, out: File) {
     }
 }
 
+fun remoteSignCodesign(signHost: String, lib: File, out: File) {
+    val user = skiko.signUser ?: error("signUser is null")
+    val token = skiko.signToken ?: error("signToken is null")
+    val cmd = """
```
olonho, line 565: Shall work on ARM macOS as well with binary translator

PR #240, opened 2021-10-06
An example of commonized entity

Schahen:
This is an example of how I intended to deal with commonization - if we are fine with this I will in very fast fashion do the rest in exactly same fashion.

### DIFF
--- a/skiko/src/commonMain/cpp/generated/PathSegmentIterator.cc
+++ b/skiko/src/commonMain/cpp/generated/PathSegmentIterator.cc
@@ -6,7 +6,7 @@
 
 SKIKO_EXPORT KNativePointer org_jetbrains_skia_PathSegmentIterator__1nMake
   (KNativePointer pathPtr, KBoolean forceClose) {
-    SkPath* path = reinterpret_cast<SkPath*>(pathPtr);
+    SkPath* path = reinterpret_cast<SkPath*>((pathPtr));
     SkPath::Iter* iter = new SkPath::Iter(*path, forceClose);
     return reinterpret_cast<KNativePointer>(iter);
 }
@@ -16,12 +16,15 @@ static void deletePathSegmentIterator(SkPath::Iter* iter) {
     delete iter;
 }
 
-SKIKO_EXPORT KNativePointer org_jetbrains_skia_PathSegmentIterator__1nGetFinalizer() {
+SKIKO_EXPORT KNativePointer org_jetbrains_skia_PathSegmentIterator__1nGetFinalizer
+  () {
     return reinterpret_cast<KNativePointer>((&deletePathSegmentIterator));
 }
 
-SKIKO_EXPORT void org_jetbrains_skia_PathSegmentIterator__1nNext(KNativePointer ptr, KInt* data) {
-    SkPath::Iter* instance = reinterpret_cast<SkPath::Iter*>(ptr);
+
+SKIKO_EXPORT void org_jetbrains_skia_PathSegmentIterator__1nNext
+  (KNativePointer ptr, KInt* data) {
+    SkPath::Iter* instance = reinterpret_cast<SkPath::Iter*>((ptr));
     SkPoint pts[4];
     SkPath::Verb verb = instance->next(pts);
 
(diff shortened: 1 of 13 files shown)


skiko/src/jvmMain/cpp/common/PathSegmentIterator.cc
```diff
@@ -19,32 +19,50 @@ extern "C" JNIEXPORT jlong JNICALL Java_org_jetbrains_skia_PathSegmentIteratorKt
...
-extern "C" JNIEXPORT jobject JNICALL Java_org_jetbrains_skia_PathSegmentIteratorKt_PathSegmentIterator_1nNext
-  (JNIEnv* env, jclass jclass, jlong ptr) {
+extern "C" JNIEXPORT void JNICALL Java_org_jetbrains_skia_PathSegmentIteratorKt_PathSegmentIterator_1nNext
+  (JNIEnv* env, jclass jclass, jlong ptr, jfloatArray result) {
     SkPath::Iter* instance = reinterpret_cast<SkPath::Iter*>(static_cast<uintptr_t>(ptr));
     SkPoint pts[4];
     SkPath::Verb verb = instance->next(pts);
-    jobject segment;
+
+    std::bitset<8> context(verb);
```
olonho, line 28: What's that code with context? Why not return context as int from the method?

Schahen: this is basically an byte which encodes following values:
type of segment (first three bytes), flag for isCloseLine and flag for isClosed. 

we can, of course return it as a separate value but I though that this will compilcate signature for helper functions like (withResult) - I am not arguing - I have a lot of doubts about this kind of serialization/deseralization

skiko/src/commonMain/kotlin/org/jetbrains/skia/PathSegmentIterator.kt
```diff
@@ -56,11 +58,57 @@ class PathSegmentIterator internal constructor(val _path: Path?, ptr: NativePoin
...
 
+private fun NativePointer.nextSegment() = pathSegmentFromFloatArray(withResult(FloatArray(10)) {
+    PathSegmentIterator_nNext(this, it)
+})
+
+
+private fun pathSegmentFromFloatArray(points: FloatArray): PathSegment {
+    val context = points.last().toInt()
+
+    val verb = (context) and ((1 shl 3) - 1)
```
olonho, line 69: Why not convert to enum here?

Schahen: Yep, let's check again enum value rather than its ordinal counterpart - pushed the fix

skiko/src/commonMain/cpp/generated/PathSegmentIterator.cc
```diff
@@ -22,40 +22,32 @@ SKIKO_EXPORT KNativePointer org_jetbrains_skia_PathSegmentIterator__1nGetFinaliz
...
-SKIKO_EXPORT KInteropPointer org_jetbrains_skia_PathSegmentIterator__1nNext
-  (KNativePointer ptr) {
-    TODO("implement org_jetbrains_skia_PathSegmentIterator__1nNext");
-}
-     
-#if 0 
-SKIKO_EXPORT KInteropPointer org_jetbrains_skia_PathSegmentIterator__1nNext
-  (KNativePointer ptr) {
+SKIKO_EXPORT void org_jetbrains_skia_PathSegmentIterator__1nNext
+  (KNativePointer ptr, float* points) {
```
olonho, line 26: Better use KInt* here, just for consistency, and recover from bits on Kotlin side.

skiko/src/jvmTest/kotlin/org/jetbrains/skiko/PathTest.kt
```diff
@@ -18,57 +18,6 @@ import kotlin.test.assertNull
 import kotlin.test.assertTrue
 
 class PathTest {
-    @Test
```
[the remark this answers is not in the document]
olonho: Ah, I see - moved to common.

olonho (approved):

PR #241, opened 2021-10-06
Fixes for multiple windows on Linux: proper dispose and vsync

igordmn:


olonho (approved):

olonho:
Shalln't we do the same in `WindowsOpenGLRedrawer`?

igordmn:
> WindowsOpenGLRedrawer

On windows we don't sync on swap buffers, we sync on dwmFlush.

As for makeCurrent, let's call it. I checked, and we have the same issue on window close, as in Linux. Thanks for the suggestion :)

PR #244, opened 2021-10-07
Cleanup generated files.

olonho:


skiko/src/commonMain/kotlin/org/jetbrains/skia/ColorType.kt
```diff
@@ -182,17 +181,18 @@ enum class ColorType {
      *
      * @return  ColorAlphaType if can be associated with colorType
      */
-    fun validateAlphaType(alphaType: ColorAlphaType): ColorAlphaType? {
-        var alphaType = alphaType
+    fun validateAlphaType(alphaType0: ColorAlphaType): ColorAlphaType? {
```
igordmn, line 184: Better not to add suffixes/digits to public API

igordmn (approved):

PR #252, opened 2021-10-11
Maybe fix type in App.ios.kt

leo6dev:
there appears to be a typo in App.ios.kt
```kotlin
/ /Use `xcodegen` first, then `open ./SkikoSample.xcodeproj` and then Rub
```
it looks like it should be
```kotlin
/ /Use `xcodegen` first, then `open ./SkikoSample.xcodeproj` and then Run
```


olonho:
Fixed in https://github.com/JetBrains/skiko/pull/258

PR #256, opened 2021-10-12
fix line metrics assertions in SkiaWindowTest.kt

eymar:
Apparently new skia contains a fix for line metrics. Our test had an assertions that is not valid anymore.


eymar:
fixed skia issue https://bugs.chromium.org/p/skia/issues/detail?id=12383

PR #260, opened 2021-10-13
Basic iOS Metal-based renderer

olonho:


skiko/src/iosMain/kotlin/org/jetbrains/skiko/redrawer/MetalRedrawer.ios.kt
```diff
@@ -1,25 +1,142 @@
...
     private val layer: SkiaLayer,
     private val properties: SkiaLayerProperties
 ) : Redrawer {
+    private var isDisposed = false
+    internal val device = MTLCreateSystemDefaultDevice()!!
+    private val queue = device.newCommandQueue()!!
+    private var currentDrawable: CAMetalDrawableProtocol? = null
+    private val metalLayer = MetalLayer(this.layer, device)
+
+    private val frameDispatcher = FrameDispatcher(AppDispatcher) {
```
igordmn, line 34: Are methods of MetalRedrawer/SkiaLayer called from the single UI thread?

skiko/src/iosMain/kotlin/org/jetbrains/skiko/redrawer/MetalRedrawer.ios.kt
```diff
@@ -1,25 +1,142 @@
...
+        check(!isDisposed) { "MetalRedrawer is disposed" }
+        frameDispatcher.scheduleFrame()
     }
 
     override fun redrawImmediately() {
-        TODO()
+        check(!isDisposed) { "MetalRedrawer is disposed" }
+        metalLayer.setNeedsDisplay()
+        update(getTimeNanos())
+        performDraw()
```
igordmn, line 66: Shouldn't we call `draw` here? (with autoreleasepool). Also. it is better to merge performDraw/draw into one function.

skiko/src/iosMain/kotlin/org/jetbrains/skiko/redrawer/MetalRedrawer.ios.kt
```diff
@@ -1,25 +1,142 @@
...
     override fun needRedraw() {
-        TODO()
+        check(!isDisposed) { "MetalRedrawer is disposed" }
+        frameDispatcher.scheduleFrame()
     }
 
     override fun redrawImmediately() {
-        TODO()
+        check(!isDisposed) { "MetalRedrawer is disposed" }
+        metalLayer.setNeedsDisplay()
```
igordmn, line 64: We don't need this, if `performDraw` works correctly. This function schedules a new frame in the built-in frame dispatcher, that calls drawInContext on vsync.

skiko/src/iosMain/kotlin/org/jetbrains/skiko/redrawer/MetalRedrawer.ios.kt
```diff
@@ -1,25 +1,142 @@
...
+        this.removeFromSuperlayer()
+        // TODO: anything else to dispose the layer?
+    }
+
+    @Suppress("unused")
+    private fun performDraw() {
+        try {
+            draw()
+        } catch (e: Throwable) {
+            e.printStackTrace()
```
igordmn, line 134: Better not to eat exceptions here.

Fallback won't work when we implement it, and log will be polluted if the exception happened during animation.

skiko/src/iosMain/kotlin/org/jetbrains/skiko/Dispatchers.kt
```diff
@@ -14,8 +16,13 @@ import kotlin.native.internal.NativePtr
 // This is the only dispatcher that shall be used in Skiko on iOS.
 // Current (as of 1.5.2) dispatchers in kotlinx.coroutines are not usable
 // for needs of Skiko.
-@SharedImmutable
-internal val AppDispatcher: CoroutineDispatcher = NsQueueDispatcher(dispatch_get_main_queue())
+object SkikoDispatchers {
```
igordmn, line 19: If we intend to make it public, it is worth to describe why we don't use `Dispatchers.Main` or `Dispatchers.IO`. If it is a bug, maybe we should fill an issue in kotlinx.coroutines, and switch to `Dispatchers` after it will be fixed?

Also, if fixing `Dispatcher.Main` seems reasonable, because it a very specific dispatcher, tied to the platform, fixing `Dispatchers.IO` seems strange.

olonho: Guess it’s a long story, and most likely we’ll need to expose that dispatcher at least to Compose.

PR #267, opened 2021-10-14
Make publications comply with Maven Central requirements

AlexeyTsvetkov:


skiko/build.gradle.kts
```diff
@@ -1048,35 +1058,82 @@ publishing {
         }
     }
     publications {
+        val pomNameForPublication = HashMap<String, String>()
+        pomNameForPublication["kotlinMultiplatform"] = "Skiko MPP"
+        kotlin.targets.forEach {
+            pomNameForPublication[it.name] = "Skiko ${it.name.capitalize(Locale.ROOT)}"
```
olonho, line 1064: Oh, that's ugly - can we make utility 
`fun String.simpleCapitalize()L String` instead and hide locale ugliness there?

PR #276, opened 2021-10-15
Input handling

olonho:


skiko/build.gradle.kts
```diff
@@ -203,7 +203,7 @@ val linkWasm = tasks.register<LinkSkikoWasmTask>("linkWasm") {
 val skiaBinSubdir = "out/${buildType.id}-${targetOs.id}-${targetArch.id}"
 
 val Project.supportNative: Boolean
-   get() = properties.get("skiko.native.enabled") == "true"
+   get() = true // properties.get("skiko.native.enabled") == "true"
```
alexander-gorshenev, line 206: looks like a leftover

alexander-gorshenev (approved):

skiko/src/commonMain/kotlin/org/jetbrains/skiko/Events.kt
```diff
@@ -0,0 +1,43 @@
...
+    val kind: SkikoKeyboardEventKind,
+    val platform: SkikoPlatformKeyboardEvent?
+)
+
+enum class SkikoMouseEventKind {
+    UP, DOWN, MOVE
+}
+expect class SkikoPlatformPointerEvent
+data class SkikoMouseEvent(
+    val x: Int, val y: Int,
```
igordmn, line 32: As discussed in Slack - better to make position as Double

skiko/src/commonMain/kotlin/org/jetbrains/skiko/SkiaLayer.kt
```diff
@@ -9,25 +9,13 @@ expect open class SkiaLayer {
...
-    val layer: SkiaLayer,
-    val app: SkiaRenderer
-): SkiaRenderer {
-    override fun onRender(canvas: Canvas, width: Int, height: Int, nanoTime: Long) {
-        val contentScale = layer.contentScale
-        canvas.scale(contentScale, contentScale)
-        app.onRender(canvas, (width / contentScale).toInt(), (height / contentScale).toInt(), nanoTime)
-        layer.needRedraw()
-    }
+fun SkiaLayer.setApp(app: SkikoApp) {
```
igordmn, line 17: SkiaLayer usually used per window.

And App usually means the whole application, not only the content of some particular Window.

Maybe we rename SkikoApp to something else? SkikoScene or SkikoContent for example.

skiko/src/commonMain/kotlin/org/jetbrains/skiko/SkiaLayer.kt
```diff
@@ -9,25 +9,13 @@ expect open class SkiaLayer {
...
+    this.app = app
```
igordmn, line 18: Maybe we remove the separate function? I think, property is enough.

skiko/src/commonMain/kotlin/org/jetbrains/skiko/SkikoApp.kt
```diff
@@ -0,0 +1,54 @@
...
+    // Rendering
+    fun onRender(canvas: Canvas, width: Int, height: Int, nanoTime: Long)
+}
+
+open class GenericSkikoApp(
+    val layer: SkiaLayer,
+    val app: SkikoApp): SkikoApp {
+
+    init {
+        layer.setApp(this)
```
igordmn, line 21: Probably we should remove this, because it is a side effect during the class creation. The more straightforward way to use this API would be:
```
layer.app = GenericSkikoApp(layer, makeApp())
```

skiko/src/commonMain/kotlin/org/jetbrains/skiko/SkikoApp.kt
```diff
@@ -0,0 +1,54 @@
+package org.jetbrains.skiko
+
+import org.jetbrains.skia.Canvas
+
+interface SkikoApp {
+    // Input
+    fun onKeyboardEvent(event: SkikoKeyboardEvent)
+    fun onMouseEvent(event: SkikoMouseEvent)
```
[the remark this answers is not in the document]
olonho: Yeah, maybe, generally this visitor-like pattern in pretty flexible in that regard.

skiko/src/commonMain/kotlin/org/jetbrains/skiko/Events.kt
```diff
@@ -0,0 +1,43 @@
+package org.jetbrains.skiko
+
+// Clicked mouse buttons bitmask.
+object SkikoMouseButtons {
```
igordmn, line 4: If we have renamed MouseEvent to PointerEvent, we probably should rename it too:
```
object SkikoPointerButtons {
    const val NONE = 0
    const val PRIMARY = 1 shl 0
    const val SECONDARY = 1 shl 1
    const val TERTIARY = 1 shl 2
    const val BACK = 1 shl 3
    const val FORWARD = 1 shl 4
}
```

or rename `buttons` to `mouseButtons` in SkikoPointerEvent.

skiko/src/commonMain/kotlin/org/jetbrains/skiko/SkiaLayer.kt
olonho: Yep.

skiko/src/commonMain/kotlin/org/jetbrains/skiko/SkikoApp.kt
```diff
@@ -0,0 +1,54 @@
+package org.jetbrains.skiko
+
+import org.jetbrains.skia.Canvas
+
+interface SkikoApp {
+    // Input
+    fun onKeyboardEvent(event: SkikoKeyboardEvent)
```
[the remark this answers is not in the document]
olonho: This is tricky question, as it generally presumes certain external event system introp. Guess we better have it as method in event instead, if we'll need it. And it requires design.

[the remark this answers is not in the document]
igordmn: > method in event instead

`consume` method is fine, I guess. It will be similar to AWT, web and Compose itself (don't know about other platforms). But we should be aware that it will make event mutable.

> if we'll need it

We'll probably need this. SkiaLayer can be used not only in windows, but in separate views (ComposePanel, for example), where we should know if the view has event consumed (for example, scroll on panel should be consumed, if the panel itself can be scrolled)

> And it requires design.

I guess, if we don't switch to this API for CfD 1.0, it is fine to make breaking changes in this interface.

skiko/src/commonMain/kotlin/org/jetbrains/skiko/SkikoApp.kt
```diff
@@ -0,0 +1,54 @@
...
+    override fun onKeyboardEvent(event: SkikoKeyboardEvent) {
+        app.onKeyboardEvent(event)
+    }
+
+    override fun onMouseEvent(event: SkikoMouseEvent) {
+        app.onMouseEvent(event)
+    }
+}
+
+abstract class NoInputSkikoApp: SkikoApp {
```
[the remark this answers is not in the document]
igordmn: It is matter of taste, I guess, but optional methods in combined interfaces like SkikoApp looks easier to use and more familiar to me as for the user of the library.

At least make it as `interface`, not `abstract class` . We can have a situation, when we add new features into SkikoApp, which can be also optional. We can't inherit two abstract classes.

samples/SkiaMultiplatformSample/src/jsMain/kotlin/org/jetbrains/skiko/sample/App.js.kt
```diff
@@ -1,17 +1,18 @@
...
 import org.jetbrains.skiko.wasm.onWasmReady
 import org.w3c.dom.HTMLCanvasElement
 
 fun main() {
     onWasmReady {
         val skiaLayer = SkiaLayer()
-        skiaLayer.renderer = GenericRenderer(skiaLayer, RotatingSquare())
+        val game = BouncingBalls()
+        skiaLayer.skikoView = GenericSkikoView(skiaLayer, game)
         val canvas = document.getElementById("SkikoTarget") as HTMLCanvasElement
```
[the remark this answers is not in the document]
olonho: Thanks, added here, as it indeed can be tricky for embedders to overwrite.

PR #284, opened 2021-10-19
IOS: rotation support, dpi support, clocks sample. Partial commonization [SystemTheme].

Rsedaikin:


samples/SkiaMultiplatformSample/SkikoSample.xcodeproj/project.pbxproj
```diff
@@ -0,0 +1,610 @@
+// !$*UTF8*$!
```
olonho, line 1: Please do not commit this directory.

PR #292, opened 2021-10-21
Universal and mostly automated dispose logic

olonho:


skiko/src/iosMain/kotlin/org/jetbrains/skiko/SkiaLayer.ios.kt
```diff
@@ -76,28 +77,41 @@ actual open class SkiaLayer(
...
         view.addGestureRecognizer(UITapGestureRecognizer(controller, NSSelectorFromString("onTap:")))
+        // TODO: maybe add observer for view.viewDidDisappear() to detach us?
         redrawer = MetalRedrawer(this, properties)
         redrawer?.redrawImmediately()
     }
 
+    private var isDisposed = false
+    actual fun detach() {
+        if (!isDisposed) {
+            redrawer?.dispose()
```
igordmn, line 88: Shouldn't we dispose contextHandler too?

PR #297, opened 2021-10-21
Fixed SkiaLayer background color on macos.

Rsedaikin:


skiko/src/jvmMain/kotlin/org/jetbrains/skiko/SkiaLayer.jvm.kt
```diff
@@ -40,7 +41,17 @@ actual open class SkiaLayer internal constructor(
...
-    actual var transparency: Boolean = false
+    private var _transparency: Boolean = false
+    actual var transparency: Boolean
+        get() = _transparency
+        set(value) {
+            _transparency = value
+            if (!value && hostOs == OS.MacOS) {
+                background = UIManager.getColor("Panel.background")
+            } else {
+                background = Color(0, 0, 0, 0)
```
igordmn, line 52: We will reset to transparent background when we set `transparency = false` on other platforms.

We should remove `&& hostOs == OS.MacOS` from the condition

PR #322, opened 2021-10-28
Implement Path methods

VeselovAlex:


eymar (approved):

skiko/src/commonMain/kotlin/org/jetbrains/skia/Path.kt
```diff
@@ -444,7 +460,8 @@ class Path internal constructor(ptr: NativePointer) : Managed(ptr, _FinalizerHol
     val isFinite: Boolean
         get() = try {
             Stats.onNativeCall()
-            _nIsFinite(_ptr)
+            // TODO For some reason this method returns 0 instead of false in JS target, investigate
+            !!_nIsFinite(_ptr)
```
Schahen, line 464: can we do something more intuitive here - the !! de-facto is used more often in context of a different idiom.

skiko/src/commonTest/kotlin/org/jetbrains/skia/PathTests.kt
```diff
@@ -213,11 +202,11 @@ class PathTests {
...
         assertFalse(Path.isCubicDegenerate(Point(0f, 0f), Point(10f, 0f), Point(0f, 0f), Point(0f, 0f), false))
         assertTrue(Path.isCubicDegenerate(Point(0f, 0f), Point(0f, 0f), Point(0f, 0f), Point(0f, 0f), false))
-        assertContentEquals(
+        assertContentCloseEnough(
             arrayOf(Point(0f, 20f), Point(6.666667f, 13.333334f)),
             Path.convertConicToQuads(Point(0f, 20f), Point(20f, 0f), Point(40f, 20f), 0.5f, 1)
         )
-        assertContentEquals(
+        assertContentCloseEnough(
             arrayOf(
```
Schahen, line 210: Just to be sure - what exactly changed so that we need to relax out equality assertions? 
May be it worth (not in this PR I mean - in general) to have a platform-specific assertions that are strictly equal wherever possible and relaxed on platform we know we can not guarantee it?

skiko/src/commonTest/kotlin/org/jetbrains/skia/util/assertCloseEnough.kt
```diff
@@ -21,10 +23,39 @@ internal fun assertCloseEnough(expected: Matrix33, actual: Matrix33?, epsilon: F
...
     assertCloseEnough(expected.r, actual!!.r, epsilon)
     assertCloseEnough(expected.g, actual.g, epsilon)
     assertCloseEnough(expected.b, actual.b, epsilon)
     assertCloseEnough(expected.a, actual.a, epsilon)
-}
\ No newline at end of file
+}
+
+internal fun assertCloseEnough(expected: Rect, actual: Rect?, epsilon: Float = EPSILON) {
+    assertCloseEnough(expected.left, actual!!.left, epsilon)
```
Schahen, line 34: should we have nullable Rect here? after all we can do the !! while passing as param if needed.

skiko/src/commonTest/kotlin/org/jetbrains/skia/util/assertCloseEnough.kt
```diff
@@ -21,10 +23,39 @@ internal fun assertCloseEnough(expected: Matrix33, actual: Matrix33?, epsilon: F
...
+
+    for (i in expected.indices) {
+        if (abs(expected[i] - actual[i]) > epsilon) {
+            throw AssertionError("results differ at index$i, expected ${expected[i]}, got ${actual[i]}")
+        }
+    }
+}
+
+internal fun assertContentCloseEnough(expected: Array<Point>, actual: Array<Point>?, epsilon: Float = EPSILON) {
+    assertNotNull(actual)
```
Schahen, line 53: a similar question about whether we need an assertion - I mean we still fail with exception if we'll try t cast.

Schahen (commented):
Looks good to me, the four comments I left are more of a subject to discuss rather then intervene to this particular PR

skiko/src/commonTest/kotlin/org/jetbrains/skia/PathTests.kt
VeselovAlex: This is made for JS purposes: in JS code "Float"s have greater precision, so strict tests fail

skiko/src/commonTest/kotlin/org/jetbrains/skia/util/assertCloseEnough.kt
VeselovAlex: We have nullable `actual` in other assertions, so let it be just for ergonomics — null and equality checks at once

PR #328, opened 2021-10-30
Look for the skiko libraries in the JVM library directory first

mikehearn:
For people who unpack Skia to their install directory.

1. Makes signing easier (don't have to sign shared libs inside zips).
2. Avoids littering the user's home directory.
3. Faster startup.

### DIFF
--- a/skiko/src/jvmMain/kotlin/org/jetbrains/skiko/Library.kt
+++ b/skiko/src/jvmMain/kotlin/org/jetbrains/skiko/Library.kt
@@ -1,9 +1,7 @@
 package org.jetbrains.skiko
 
 import org.jetbrains.skia.Bitmap
-import java.io.BufferedReader
 import java.io.File
-import java.io.InputStreamReader
 import java.nio.file.Files
 import java.nio.file.StandardCopyOption
 import java.util.concurrent.atomic.AtomicBoolean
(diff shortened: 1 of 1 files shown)


skiko/src/jvmMain/kotlin/org/jetbrains/skiko/Library.kt
```diff
@@ -52,50 +50,70 @@ object Library {
...
-            }
+            return
         }
 
-        // TODO move properties to SkikoProperties
-        Setup.init()
+        // Second try: load it from the bin/ or lib/ directory relative to the JVM home.
+        // The user might have placed the native files alongside the other JVM libraries
+        // for signing purposes, so we'll find it here if so.
+        val jvmFiles = File(System.getProperty("java.home"), if (hostOs.isWindows) "bin" else "lib")
```
olonho, line 89: Wouldn't it be natural to search in java.library.home instead?

mikehearn: Hmm, where do you see that property? I don't seem to have it when I run `java -XshowSettings:properties` and I can't find it by grepping the OpenJDK sources either.

olonho: Smth like that could print all the props

```java
import java.util.*;

class a {
    public static void main(String[] args) {
        Properties p = System.getProperties();
        Enumeration keys = p.keys();
        while (keys.hasMoreElements()) {
            String key = (String)keys.nextElement();
            String value = (String)p.get(key);
            System.out.println(key + ": " + value);
        }
    }
}
```
and then `java a.java`.

mikehearn: Scanning `java.library.path` is what `System.loadLibrary` does, but in our conversation on Friday you expressed that this would not be ideal because it would potentially pick up library conflicts. And indeed as you see, `java.library.path` has quite a few legacy directories in it where there may be system-installed stuff as well as `.`. Checking the lib directory in the JVM install path only, avoids the potential for conflicts.

olonho: Well, I am just a bit concerned about cases where we search in JVM home  directory, what is desired behavior here? Why we assume that people put Skiko’s native library to the JDK distribution? Per https://javarevisited.blogspot.com/2013/04/what-is-javalibrarypath-how-to-set-in-Eclipse.html?m=1 and similar docs it seems that java.libary.path is the property intended for exactly the usecase we discuss.

olonho: And we need to scan that explicitly to be able to make a copy for several classloaders case.

mikehearn: The code doesn't assume that - if someone does put it alongside the other JNI libs there, it'll work after this change, but if they don't there should be no difference.

I think the code in this PR will happily make copies for multiple classloaders because `loadLibraryOrCopy` is still called.

`java.library.path` can be used. It's equivalent to starting with `loadLibrary` and only continuing on the old codepath if it doesn't work. It means if there's a skiko library in the user's current working directory or , that will override the one in the JAR or JVM. Also consider that `java.library.path` contains many other entries on Windows, for instance, mine has:

```
    java.library.path = C:\Program Files\Amazon Corretto\jdk16.0.2_7\bin
        C:\Windows\Sun\Java\bin
        C:\Windows\system32
        C:\Windows
        C:\Program Files (x86)\Parallels\Parallels Tools\Applications
        C:\Program Files\Amazon Corretto\jdk16.0.2_7\bin
        C:\Windows\system32
        C:\Windows
        C:\Windows\System32\Wbem
        C:\Windows\System32\WindowsPowerShell\v1.0\
        C:\Windows\System32\OpenSSH\
        C:\Program Files\CMake\bin
        C:\Program Files\dotnet\
        C:\Program Files (x86)\Windows Kits\10\Windows Performance Toolkit\
        C:\Users\mike\AppData\Local\Microsoft\WindowsApps
        .
```

i.e. the entire PATH is searched. So that's a lot of places where Skiko's library might be found. In particular, if you have two programs using Skiko and one is on the PATH, then that will potentially take precedence (unless `java.library.path` is overridden from the command line). That's probably not what's wanted.

I think there's really only a handful of places an app might want to load a JNI library from. The JAR itself is one, the source tree during development is another, and the bin/lib directory is the other one because that's where all the JNI libs used by e.g. Swing, JavaFX, etc go by convention.

olonho: Generally we could merge this PR, if it will make your life easier (and also makes code a bit more readable), although I'm still not 100% convinced that this is indeed "the right thing" ®.

olonho (approved):

skiko/src/jvmMain/kotlin/org/jetbrains/skiko/Library.kt
mikehearn: Thanks! It's an incremental improvement and we can refine it further if need be.

PR #331, opened 2021-11-01
Implement native methods for RuntimeEffect

VeselovAlex:


skiko/src/nativeJsMain/cpp/RuntimeEffect.cc
```diff
@@ -5,75 +5,62 @@
...
     }
-    env->ReleaseLongArrayElements(childrenPtrsArr, childrenPtrs, 0);
 
     sk_sp<SkShader> shader = runtimeEffect->makeShader(sk_ref_sp<SkData>(uniform),
                                                        children.data(),
                                                        childCount,
                                                        localMatrix.get(),
                                                        isOpaque);
-    return ptrToKNativePointer(shader.release());
+    return reinterpret_cast<KNativePointer>(shader.release());
```
olonho, line 25: Technically this cast is not required.

skiko/src/commonMain/kotlin/org/jetbrains/skia/impl/Native.kt
```diff
@@ -109,8 +109,8 @@ inline fun withResult(result: NativePointerArray, block: (InteropPointer) -> Uni
 }
 
 @Suppress("NON_PUBLIC_CALL_FROM_PUBLIC_INLINE")
-inline fun withStringResult(block: () -> NativePointer): String {
-    val string = ManagedString(block())
+inline fun withStringResult(managed: Boolean = true, block: () -> NativePointer): String {
```
olonho, line 112: Why would we want `managed = false` here?

VeselovAlex: Some SkString instances are not owned by another objects, we are receiving them by const&. So it is UB to delete them in ManagedString

olonho: For such cases we shall hold reference to those objects, as in here https://github.com/JetBrains/skiko/blob/09aad6044e04a0b87ae4a9489de9c2236417d735/skiko/src/commonMain/kotlin/org/jetbrains/skia/Pixmap.kt#L12

VeselovAlex: This requires a special Managed class just to keep temporary objects. IMHO it is an overkill

olonho (approved):

PR #333, opened 2021-11-01
Add zero terminator to native interop strings

VeselovAlex:


skiko/src/nativeMain/kotlin/org/jetbrains/skia/impl/Native.native.kt
```diff
@@ -58,7 +58,8 @@ actual class InteropScope actual constructor() {
     actual fun toInterop(string: String?): InteropPointer {
         return if (string != null) {
             // encodeToByteArray encodes to utf8
-            val pinned = string.encodeToByteArray().pin()
+            val utf8 = string.encodeToByteArray()
```
olonho, line 61: Not sure if additional copy looks like a good idea. Why not pass produced byte array to C++ side and use SkString(pointer, size) constructor?

VeselovAlex: Yes, zero-copy solution will be much better, but this requires some refactoring. Can changes be postponed?

olonho: Ok, at least add TODO

olonho (approved):

PR #340, opened 2021-11-02
Fix memory leaks on macOs, enable Arc compilation

igordmn:


skiko/src/jvmMain/objectiveC/macos/Drawlayer.mm
```diff
@@ -281,90 +273,105 @@ - (void) makeFullscreen: (BOOL) value
...
     }
 }
 
 JNIEXPORT jboolean JNICALL Java_org_jetbrains_skiko_PlatformOperationsKt_osxIsFullscreenNative(JNIEnv *env, jobject properties, jobject component)
 {
-    LayerHandler *layer = findByObject(env, component);
-    if (layer != NULL)
-    {
-        return [layer isFullScreen];
+    @autoreleasepool {
```
olonho, line 307: Do we need autorelease pools like that with ARC?

igordmn: Yes, we still need pools. Without them, we have memory leaks.

I just followed the simple rule - if the function is JNI function, it should be wrapped into `autoreleasepool`, without thinking about what this function do, or from where it can be called. It simplifies the code, and reduce the number of possible leaks.

olonho: It's not exactly like that. If it's a JNI function in .mm file at least. Our Skia wrappers do not need such thing. And I still consider that as a hack.

igordmn: > If it's a JNI function in .mm file at least.

Yes, in mm file, wher we use Objective C style functions.

> And I still consider that as a hack.

Any option we use to fix that will be a hack. The option just to create pool and clear every frame I consider a very huge hack. Can explain, if interested.

olonho: It really depends on performance. So far, none of our ObjC JNI methods are high frequency, so that would work, I guess.

skiko/src/jvmMain/objectiveC/macos/NativeApplication.mm
```diff
@@ -0,0 +1,20 @@
+#import "jawt.h"
+#import "jawt_md.h"
+#import <Cocoa/Cocoa.h>
+#import <QuartzCore/QuartzCore.h>
+
+
+extern "C" void* objc_autoreleasePoolPush(void);
```
olonho, line 7: Why it is needed here?

olonho (approved):

PR #341, opened 2021-11-02
Implement Animation native methods, add test

VeselovAlex:


skiko/src/jvmTest/kotlin/org/jetbrains/skia/skottie/AnimationTest.jvm.kt
```diff
@@ -0,0 +1,31 @@
...
+import org.jetbrains.skiko.kotlinBackend
+import org.jetbrains.skiko.resourcePath
+import org.junit.Test
+import kotlin.test.assertEquals
+
+
+class AnimationTestJVM {
+    @Test
+    fun canCreateFromFile() {
+        if (kotlinBackend == KotlinBackend.JS) {
```
olonho, line 16: How JVM-only code could be called on JS?

skiko/src/commonMain/kotlin/org/jetbrains/skia/skottie/Animation.kt
```diff
@@ -285,10 +276,10 @@ class Animation internal constructor(ptr: NativePointer) : Managed(ptr, _Finaliz
 private external fun _nGetFinalizer(): NativePointer
 
 @ExternalSymbolName("org_jetbrains_skia_skottie_Animation__1nMakeFromString")
-private external fun _nMakeFromString(data: String?): NativePointer
+private external fun _nMakeFromString(data: InteropPointer): NativePointer
 
 @ExternalSymbolName("org_jetbrains_skia_skottie_Animation__1nMakeFromFile")
-private external fun _nMakeFromFile(path: String?): NativePointer
+internal external fun _nMakeFromFile(path: InteropPointer): NativePointer
```
olonho, line 282: Aren't internal functions fancy mangled?

VeselovAlex: Probably are, but seems that `external` wins in this case. We already use `internal external` in several places. Maybe there is better way to make native functions available only inside `nativeMain` and `jvmMain` source sets?

olonho (approved):

PR #342, opened 2021-11-02
use ManagedString._nMake from toInterop(String)

eymar:
the previous variant didn't crash, but the resulting SkString was invalid.

### DIFF
--- a/skiko/src/commonTest/kotlin/org/jetbrains/skia/ManagedStringTest.kt
+++ b/skiko/src/commonTest/kotlin/org/jetbrains/skia/ManagedStringTest.kt
@@ -6,6 +6,7 @@ import kotlin.test.Test
 import kotlin.test.assertEquals
 
 class ManagedStringTest {
+
     @Test
     fun basicTest() {
         val s1 = withStringResult {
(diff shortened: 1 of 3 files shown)


VeselovAlex (approved):

skiko/src/commonTest/kotlin/org/jetbrains/skia/ManagedStringTest.kt
```diff
@@ -21,4 +22,37 @@ class ManagedStringTest {
...
+        val ms2 = ManagedString("Привет!")
+        assertEquals("Привет!", ms2.toString())
+
+        val ms3 = ManagedString("你好!")
+        assertEquals("你好!", ms3.toString())
+    }
+
+    @Test
+    fun canAppend() {
+        val ms = ManagedString("Hello").append(" World!")
```
olonho, line 40: Guess we miss append and insert tests in non-8bit locale.

eymar: good point! I'll add them too

eymar: done

olonho (approved):

PR #345, opened 2021-11-03
add commonMain/cpp for common bindings (js,native,jvm)

eymar:
this will help avoiding code duplication for jvmMain and nativeJsMain sources


igordmn (approved):

skiko/src/commonMain/cpp/common/shaper/FontRunIterator.cc
```diff
@@ -1,5 +1,5 @@
 #include "SkShaper.h"
-#include "FontRunIterator.hh"
+#include "../include/FontRunIterator.hh"
```
olonho, line 2: Somewhat ugly, maybe link directly via `-I`.

olonho (approved):

PR #348, opened 2021-11-03
Fix crash when transparent window displayed on external display on linux.

Rsedaikin:


skiko/src/jvmMain/kotlin/org/jetbrains/skiko/SkiaLayer.jvm.kt
```diff
@@ -95,10 +95,21 @@ actual open class SkiaLayer internal constructor(
     }
 
     override fun removeNotify() {
-        dispose()
+        disposeSkiaResources()
```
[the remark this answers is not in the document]
igordmn: Also we should call full `dispose` (with `backedLayer.dispose()`), as it frees resources on macOs. Without it, we will have a memory leak, if user won't manually call `dispose`.

skiko/src/jvmMain/kotlin/org/jetbrains/skiko/SkiaLayer.jvm.kt
```diff
@@ -275,17 +284,27 @@ actual open class SkiaLayer internal constructor(
...
-            contextHandler?.dispose()
-            picture?.instance?.close()
-            pictureRecorder.close()
+            disposeSkiaResources()
             backedLayer.dispose()
             isDisposed = true
         }
     }
 
+    private fun disposeSkiaResources() {
```
igordmn, line 294: Can we make disposeSkiaResources/initSkiaResources symmetric? For better representation of the data flow. I.e. don't close picture/pictureRecorder here.

Also there is a PR, where I move contextHanlder from here. So we will have only `createRedrawer` + `syncSize` and `redrawer.dispose`, so we probably don't need separate functions.

skiko/src/jvmMain/kotlin/org/jetbrains/skiko/SkiaLayer.jvm.kt
```diff
@@ -53,7 +53,7 @@ actual open class SkiaLayer internal constructor(
             }
         }
 
-    internal val backedLayer: HardwareLayer
+    internal var backedLayer: HardwareLayer
```
igordmn, line 56: We don't change it. So it still can be `val`?

skiko/src/jvmMain/kotlin/org/jetbrains/skiko/SkiaLayer.jvm.kt
```diff
@@ -229,7 +240,7 @@ actual open class SkiaLayer internal constructor(
 
     @Volatile
     private var picture: PictureHolder? = null
-    private val pictureRecorder = PictureRecorder()
+    private var pictureRecorder = PictureRecorder()
```
olonho, line 243: Agree with comment about symmetry. Either all init shall be in init function and var's be null inited, or, what I believe it better to move all changeable state to object with val's and dispose it as a whole.

skiko/src/jvmMain/kotlin/org/jetbrains/skiko/SkiaLayer.jvm.kt
```diff
@@ -95,10 +95,21 @@ actual open class SkiaLayer internal constructor(
...
     override fun removeNotify() {
-        dispose()
+        disposeSkiaResources()
+        isDisposed = true
         super.removeNotify()
     }
 
+    override fun addNotify() {
+        super.addNotify()
+        if (isInited) {
```
igordmn, line 105: We have initialization in `init` method, which we happened when window is Displayeable.

I am afraid we can call `addNotify` and `init` at the same time. That can lead to memory leaks and double initialization.

The more right approach will be to remove `if (it.changeFlags and HierarchyEvent.DISPLAYABILITY_CHANGED.toLong() != 0L) {` and call `init` here.

Probably the easiest code will be:
```
override fun addNotify() {
        super.addNotify()
        init()
}


override fun removeNotify() {
        dispose()
        super.addNotify()
}
```

Rsedaikin: No - we shouldn't call `backedLayer.init ()` in case of a "recreation window". All we need to do is recreate the `ContextHandler` and `Redrawer` with the current `renderApi`.

skiko/src/jvmMain/kotlin/org/jetbrains/skiko/SkiaLayer.jvm.kt
Rsedaikin: Done.

skiko/src/jvmMain/kotlin/org/jetbrains/skiko/SkiaLayer.jvm.kt
Rsedaikin: Removed these methods. It looks like after combining `ContextHandler` and `Redrawer` we won't need them.

skiko/src/jvmMain/kotlin/org/jetbrains/skiko/SkiaLayer.jvm.kt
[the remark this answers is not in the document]
Rsedaikin: In the case of the "recreation window" we must not remove the `backedLayer`, otherwise we lost all the added event listeners and, unfortunately, the case of the "recreation window" was not intended to remove any other native resources other than the window itself (AWT/Swing just re-adds all the components in place). So in this case, there should be no memory leak (by the way on macos I have not encountered this "recreation window" case).

And, of course, there is a problem with swing interop - when we want to remove `ComposePanel`, we have to remove `backedLayer` manually.

skiko/src/jvmMain/kotlin/org/jetbrains/skiko/SkiaLayer.jvm.kt
igordmn: > we shouldn't call backedLayer.init () in case of a "recreation window".

Why? Imagine the window is recreated on macOs. In that case `window` property will point to a wrong window.

1. `addNotify` not only called in case of recreation window, but in the usual case.
2. decision of calling or not calling of `backedLayer.init()` shouldn't be based on its implementation.
3. init/dispose symmetry without additional init/dispose will significally simplify the code

skiko/src/jvmMain/kotlin/org/jetbrains/skiko/SkiaLayer.jvm.kt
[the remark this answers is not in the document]
igordmn: > In the case of the "recreation window" we must not remove the backedLayer, otherwise we lost all the added event listeners

We don't recreate class `HardwareLayer`, we just call `nativeInit`. In that case we don't lose listeners.

> as not intended to remove any other native resources other than the window itself

`addNotify`/`removeNotify` exists for the case if third-party developers want to init/free their resources without providing additional `dispose` method to the user.

Imagine we don't have `dispose` method (it can be our plan), similar to ComposePanel.

>  So in this case, there should be no memory leak

But it is, I see it with my eyes :). User don't call `dispose` - we don't free Window on macOs. I am still talking about the usual case, not the `recreation case`.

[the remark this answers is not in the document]
Rsedaikin: > But it is, I see it with my eyes :)
How?:) As I say - there is no recreation window case on macos - only if you manually call `remove` `SkiaLayer` you get memory leak and I say:
> when we want to remove `ComposePanel`, we have to remove (dispose) `backedLayer` manually.
But I get what you mean - I'm implementing it right now.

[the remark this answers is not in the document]
igordmn: > How?:) 
```
val frame = JFrame()
val layer = SkiaLayer()
frame.contentPane.add(layer)
frame.isVisible = true
frame.dispose()
```

> we have to remove (dispose) backedLayer manually.

That is exactly what I am talking about. If we call `dispose` in `removeNotify`, we don't have to remove it manually. And that was our intention, when we overrode `removeNotify`, to not call `SkiaLayer.dispose` manually.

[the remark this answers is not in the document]
igordmn: By the way, we also can deprecate `dispose` method, an leave it empty, if it will simplify the PR.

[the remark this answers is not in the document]
Rsedaikin: > By the way, we also can deprecate dispose method,

For now it should be discussed because of now we have `expect detach()` which in JVM is `dispose()`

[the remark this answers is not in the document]
Rsedaikin: Check new commit.

skiko/src/jvmMain/kotlin/org/jetbrains/skiko/SkiaLayer.jvm.kt
Rsedaikin: Check new commit.

igordmn (approved):

PR #349, opened 2021-11-03
Fix unclickable area where the hidden window title would be on macOS

manu-unter:


olonho (approved):

olonho:
Could we have tests on that functionality?

PR #351, opened 2021-11-03
Move Linux OpenGL renderer to another thread.

igordmn:
To avoid Swing thread to be wasted.

It should reduce input lag from input events.

### DIFF
--- a/skiko/src/jvmMain/kotlin/org/jetbrains/skiko/redrawer/LinuxOpenGLRedrawer.kt
+++ b/skiko/src/jvmMain/kotlin/org/jetbrains/skiko/redrawer/LinuxOpenGLRedrawer.kt
@@ -4,28 +4,40 @@ import kotlinx.coroutines.*
 import kotlinx.coroutines.swing.Swing
 import org.jetbrains.skiko.*
 import org.jetbrains.skiko.context.OpenGLContextHandler
+import java.util.concurrent.Executors
 
 internal class LinuxOpenGLRedrawer(
     private val layer: SkiaLayer,
     private val properties: SkiaLayerProperties
 ) : Redrawer {
     private var isDisposed = false
-    private var context = 0L
-    private val swapInterval = if (properties.isVsyncEnabled) 1 else 0
 
-    init {
-    	layer.backedLayer.lockLinuxDrawingSurface {
-            context = it.createContext(layer.transparency)
-            if (context == 0L) {
-                throw RenderException("Cannot create Linux GL context")
-            }
-            it.makeCurrent(context)
-            if (!isVideoCardSupported(layer.renderApi)) {
-                throw RenderException("Cannot create Linux GL context")
-            }
-            it.setSwapInterval(swapInterval)
+    private var context = 0L
+    private val defaultSwapInterval = if (properties.isVsyncEnabled) 1 else 0
+    private var swapInterval = -1
+
+    private fun LinuxDrawingSurface.setSwapIntervalFast(swapInterval: Int) {
+        if (this@LinuxOpenGLRedrawer.swapInterval != swapInterval) {
+            setSwapInterval(swapInterval)
         }
     }
+
+    init {
+        runBlocking {
+            inDrawThread {
+                context = it.createContext(layer.transparency)
+                if (context == 0L) {
+                    throw RenderException("Cannot create Linux GL context")
+                }
+                it.makeCurrent(context)
+                if (!isVideoCardSupported(layer.renderApi)) {
+                    throw RenderException("Cannot create Linux GL context")
+                }
+                it.setSwapIntervalFast(defaultSwapInterval)
+            }
+        }
+    }
+
     private val frameJob = Job()
     @Volatile
     private var frameLimit = 0.0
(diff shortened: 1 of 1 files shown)


skiko/src/jvmMain/kotlin/org/jetbrains/skiko/redrawer/LinuxOpenGLRedrawer.kt
```diff
@@ -4,28 +4,40 @@ import kotlinx.coroutines.*
...
+            inDrawThread {
+                context = it.createContext(layer.transparency)
+                if (context == 0L) {
+                    throw RenderException("Cannot create Linux GL context")
+                }
+                it.makeCurrent(context)
+                if (!isVideoCardSupported(layer.renderApi)) {
+                    throw RenderException("Cannot create Linux GL context")
+                }
+                it.setSwapIntervalFast(defaultSwapInterval)
```
olonho, line 36: Seems here `setSwapInterval` shall be used.

skiko/src/jvmMain/kotlin/org/jetbrains/skiko/redrawer/LinuxOpenGLRedrawer.kt
```diff
@@ -95,6 +125,12 @@ internal class LinuxOpenGLRedrawer(
             .filterNot(LinuxOpenGLRedrawer::isDisposed)
             .filter { it.layer.isShowing }
 
+        private val drawDispatcher = Executors.newSingleThreadExecutor {
```
olonho, line 128: So all drawing happens in this thread, no matter how many open windows we got?

skiko/src/jvmMain/kotlin/org/jetbrains/skiko/redrawer/LinuxOpenGLRedrawer.kt
```diff
@@ -68,23 +83,38 @@ internal class LinuxOpenGLRedrawer(
...
     }
 
     private fun update(nanoTime: Long) {
         layer.update(nanoTime)
     }
 
-    private fun draw() {
-        layer.inDrawScope(layer::draw)
+    private suspend fun draw(withVsync: Boolean) {
+        layer.inDrawScope {
```
olonho, line 99: Shalln't we swap the order and do
```
inDrawThread {
   layer.inDrawScope {
```

skiko/src/jvmMain/kotlin/org/jetbrains/skiko/redrawer/LinuxOpenGLRedrawer.kt
olonho: Maybe write comment, and also on how we ensure that no races on drawing context happen?

skiko/src/jvmMain/kotlin/org/jetbrains/skiko/redrawer/LinuxOpenGLRedrawer.kt
```diff
@@ -68,23 +83,38 @@ internal class LinuxOpenGLRedrawer(
...
+    override fun redrawImmediately() {
         check(!isDisposed) { "LinuxOpenGLRedrawer is disposed" }
         update(System.nanoTime())
-        it.makeCurrent(context)
-        draw()
-        it.setSwapInterval(0)
-        it.swapBuffers()
-        OpenGLApi.instance.glFinish()
-        it.setSwapInterval(swapInterval)
+        runBlocking {
```
olonho, line 89: So we'll block AWT thread here? Shall we?

skiko/src/jvmMain/kotlin/org/jetbrains/skiko/redrawer/LinuxOpenGLRedrawer.kt
igordmn: `setSwapIntervalFast` implemented in a way, that it has uninitialized value `-1`, so setSwapIntervalFast will work, and it is more consistent with the other code than simple setSwapInterval

skiko/src/jvmMain/kotlin/org/jetbrains/skiko/redrawer/LinuxOpenGLRedrawer.kt
igordmn: No, inDrawScope should be called in Swing thread, as it contain `fallback` logic

olonho:
With this patch VirtualBox with OpenGL hardware acceleration hard locks the VM.

skiko/src/jvmMain/kotlin/org/jetbrains/skiko/redrawer/LinuxOpenGLRedrawer.kt
igordmn: Yes, the semantic of this method is to draw content in a blocking way, but without vsync synchronization

igordmn:
> With this patch VirtualBox with OpenGL hardware acceleration hard locks the VM when running JVM sample.

I was afraid of it, but didn't reproduce on the real machine :(.

It seems it is because of `fallback`.

Will investigate, and if it is not simple, do it after Compose 1.0

PR #353, opened 2021-11-04
TextLine common implementations

eymar:
Changes:
- use `-sALLOW_MEMORY_GROWTH=1` for wasm (js tests started to fail because of memory limits ~16MB).
-  Moved `TestLineTest` into `TextLineTest` (the filename had a typo)

Not solved issues:
- some characters don't have correct glyphs in js. (native and jvm have correct glyphs). All targets use the same fonts from resources.


skiko/src/commonTest/kotlin/org/jetbrains/skia/TextLineTest.kt
```diff
@@ -125,22 +123,108 @@ class TextLineTest {
...
     fun emojiTest() = runTest {
+        assertEquals("☺", ManagedString("☺").toString())
+        assertEquals("☺️", ManagedString("☺️").toString())
+
         TextLine.make("☺", firaCode36()).use { misc ->
             TextLine.make("☺️", firaCode36()).use { emoji ->
                 assertContentEquals(shortArrayOf(1706), misc.glyphs)
                 assertEquals(1, emoji.glyphs.size)
-                assertNotEquals(misc.glyphs.get(0), emoji.glyphs.get(0))
+                if (kotlinBackend.isNotJs()) { // TODO(karpovich): figure out why js gives different result
```
VeselovAlex, line 149: I'm not sure that emojis are present in Fira Code's glyph set. In this case emojis may be rendered from another font

skiko/build.gradle.kts
```diff
@@ -186,6 +186,7 @@ val linkWasm = tasks.register<LinkSkikoWasmTask>("linkWasm") {
         "-l", "GL",
         "-s", "USE_WEBGL2=1",
         "-s", "OFFSCREEN_FRAMEBUFFER=1",
+        "-s", "ALLOW_MEMORY_GROWTH=1", // Is there a better way? Should we use `-s INITIAL_MEMORY=X`?
```
Schahen, line 189: I suggest to add TODO because it's easier to find all places where we actually consider to improve/investigate something

skiko/src/nativeJsMain/cpp/TextLine.cc
```diff
@@ -166,89 +98,75 @@ SKIKO_EXPORT void org_jetbrains_skia_TextLine__1nGetGlyphs
...
+  (KNativePointer ptr, KInt* resultArray) {
+    TextLine* instance = reinterpret_cast<TextLine*>(ptr);
 
-SKIKO_EXPORT KInt org_jetbrains_skia_TextLine__1nGetOffsetAtCoord
-  (KNativePointer ptr, KFloat x) {
-    TODO("implement org_jetbrains_skia_TextLine__1nGetOffsetAtCoord");
+    size_t added = 0;
+    for (auto& run: instance->fRuns) {
+        size_t count = run.fBreakOffsets.size();
+        std::memcpy(resultArray + added, run.fBreakOffsets.data(), count * sizeof(uint32_t));
```
Schahen, line 165: may be we should refer in such places to the size of the array we actualy expect to get (in out case sizeof(KInt)?

Schahen (commented):
Apart from comment left has nothing else to say (this is pretty much how I would've done it)

skiko/src/commonTest/kotlin/org/jetbrains/skia/TextLineTest.kt
olonho: Right, font callbacks (no available on JS) could be the reason why emojis are properly rendered on those platforms. Maybe execute tests with special font manager not having system fallbacks?

olonho: Or at least add TODO for that.

skiko/build.gradle.kts
olonho: Unfortunate part here is that per https://groups.google.com/g/emscripten-discuss/c/m10q3iDixnQ this option is incompatible with threading. Would be cool to come up with some sensible approach.

olonho: But after some thinking, maybe that's OK, as potentially Skia can manipulate with rather large arrays, so we have to allow it to grow memory, and maybe we can do concurrrency on JS side (i.e. Workers) if needed.

skiko/src/commonTest/kotlin/org/jetbrains/skia/TextLineTest.kt
eymar: Unfortunately NotoEmoji didn't help - both glyphs were the same.  So added TODO for now.

P.S.
Interesting:
Screenshot 2021-11-08 at 10 50 51

Github ui also treats these both emojis equally, although they're different in IDE.

skiko/build.gradle.kts
eymar: added TODO to not forget about this discussion.

skiko/src/nativeJsMain/cpp/TextLine.cc
eymar: done

olonho (approved):

PR #354, opened 2021-11-05
withResult functions accept InteropScopeContext functions

Schahen:
Otherwise we have to write things like:

```
  withResult(ShortArray(uni.size)) {
     Typeface_nGetUTF32Glyphs(_ptr, interopScope { toInterop(uni) }, uni.size, it)
 }
```
Which can be error-prone and actualy misleading (because we shouldn't wrap separate params ito nteropScope - this works by accident.

### DIFF
--- a/skiko/src/commonMain/kotlin/org/jetbrains/skia/impl/Native.kt
+++ b/skiko/src/commonMain/kotlin/org/jetbrains/skia/impl/Native.kt
@@ -47,14 +47,14 @@ expect class InteropScope() {
 
 expect inline fun <T> interopScope(block: InteropScope.() -> T): T
 
-inline fun withResult(result: ByteArray, block: (InteropPointer) -> Unit): ByteArray = interopScope {
+inline fun withResult(result: ByteArray, block: InteropScope.(InteropPointer) -> Unit): ByteArray = interopScope {
     val handle = toInterop(result)
     block(handle)
     handle.fromInterop(result)
     result
 }
 
-inline fun withNullableResult(result: ByteArray, block: (InteropPointer) -> Boolean): ByteArray? = interopScope {
+inline fun withNullableResult(result: ByteArray, block: InteropScope.(InteropPointer) -> Boolean): ByteArray? = interopScope {
     val handle = toInterop(result)
     return if (block(handle)) {
         handle.fromInterop(result)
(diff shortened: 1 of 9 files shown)


eymar (approved):

skiko/src/commonMain/kotlin/org/jetbrains/skia/impl/Native.kt
```diff
@@ -44,16 +44,18 @@ expect class InteropScope() {
...
-inline fun withResult(result: ByteArray, block: (InteropPointer) -> Unit): ByteArray = interopScope {
+inline fun withResult(result: ByteArray, block: InteropScopeContext): ByteArray = interopScope {
     val handle = toInterop(result)
     block(handle)
     handle.fromInterop(result)
     result
 }
 
-inline fun withNullableResult(result: ByteArray, block: (InteropPointer) -> Boolean): ByteArray? = interopScope {
+inline fun withNullableResult(result: ByteArray, block: InteropScope.(InteropPointer) -> Boolean): ByteArray? = interopScope {
```
VeselovAlex, line 58: Cool. Maybe also introduce alias for `InteropScope.(InteropPointer) -> Boolean`?

VeselovAlex (approved):

skiko/src/commonMain/kotlin/org/jetbrains/skia/impl/Native.kt
Schahen: We have one or two places with such signature and I decided it would be overkill.
It's a pity we don't have default type params in Kotlin

skiko/src/commonMain/kotlin/org/jetbrains/skia/impl/Native.kt
```diff
@@ -45,16 +45,18 @@ expect class InteropScope() {
     fun release()
 }
 
+typealias InteropScopeContext = InteropScope.(InteropPointer) -> Unit
```
olonho, line 48: `InteropScopeContext` is somewhat misname, IMO, it's not context, but block of code, maybe `InteropScopeBlock` or `InteropScopeFunction`, interop context is created by functions like `withResult`.
Maybe readability wouldn't suffer much is no type alias was involved at all.

Schahen: I do agree with both statements - InteropScopeBlock is a better name and not having alias at all is also an option.
Choosing between this two options I decided to choose one that introduce less entities - so no alias at all. This change is pushed.

olonho (approved):

olonho:
Would be interesting to see performance implications of our changes like that, especially for targets like JVM where interop scope is essentially no-op.

PR #358, opened 2021-11-08
TextBlob common

eymar:
Even though CI still complains about some tests, I think this PR is ready for review. I'm fixing tests issues now.

Changes:
- Tried to reuse as much logic as possible: moved some classes and functions from jni to commonMain/cpp `mppinterop.cc`


.github/workflows/blank.yml
```diff
@@ -33,6 +33,7 @@ jobs:
           ./gradlew -Pskiko.native.enabled=true macosX64Test iosX64Test
           ./gradlew -Pskiko.native.enabled=true publishToMavenLocal
       - uses: actions/upload-artifact@v2
+        if: ${{ true }}
```
olonho, line 36: Maybe remove those.

skiko/src/commonMain/cpp/common/include/RunRecordClone.hh
```diff
@@ -0,0 +1,63 @@
...
+    uint8_t positioning() const {
+        return fFlags & 0x3; // kPositioning_Mask
+    }
+
+    bool isExtended() const {
+        return fFlags & 0x8; // kExtended_Flag
+    }
+
+    static unsigned ScalarsPerGlyph(uint8_t pos) {
+        const uint8_t gScalarsPerPositioning[] = {
```
olonho, line 55: Why we cannot do that with `switch`?

skiko/src/commonTest/kotlin/org/jetbrains/skia/TextBlobTest.kt
```diff
@@ -0,0 +1,228 @@
...
+import kotlin.math.sin
+import kotlin.test.*
+
+class TextBlobTest {
+
+    private val inter36: suspend () -> Font = suspend {
+        Font(Typeface.makeFromResource("./fonts/Inter-Hinted-Regular.ttf"), 36f)
+    }
+
+    private val eps = 0.001f // only this value works for kotlin/js
```
olonho, line 18: Shall we select epsilon based on target platform?

.github/workflows/blank.yml
eymar: that was an attempt to make `upload-test-results` step run despite failing tests. But `${{true}}` didn't help for some reason. By default a step gets executed when preceding steps were successful. This issue doesn't let us download the test report when any test failed - not very useful. I'll check this problem later.

Removed for now.

skiko/src/commonMain/cpp/common/include/RunRecordClone.hh
eymar: sure. Didn't think about it when copy/pasted from jni.

skiko/src/commonTest/kotlin/org/jetbrains/skia/TextBlobTest.kt
eymar: applied your suggestion

PR #365, opened 2021-11-08
Use correct Library class for library loading

smallshen:


olonho:
Do not think this is intended fix. Essentially, it was an artefact of separate Skia bindings we used to have, and in case of Skiko rendering APIs calling to Skia loader is not needed. While not harmful, could be misleading.

olonho (changes requested):
We need either abandon this change or make wider change to make single Library class. Guess second one is more correct.

olonho:
@smallshen see https://github.com/JetBrains/skiko/pull/449 - feel free to add more stubs, if needed.

olonho:
Let's close it for now, reopen if/when relevant.

PR #366, opened 2021-11-09
Commonize TextStyle

Schahen:
This PR was getting too big so I decided to stop so that we can review the most important thing in this PR - that is, passing what-to-be-nullable as NaNs from the backend.


skiko/src/nativeJsMain/cpp/paragraph/TextStyle.cc
```diff
@@ -233,137 +239,134 @@ SKIKO_EXPORT KInteropPointerArray org_jetbrains_skia_paragraph_TextStyle__1nGetF
...
+    return reinterpret_cast<KNativePointer>(new SkString(instance->getLocale()));
 }
 
-#if 0
 SKIKO_EXPORT void org_jetbrains_skia_paragraph_TextStyle__1nSetLocale
   (KNativePointer ptr, KInteropPointer locale) {
-    TextStyle* instance = reinterpret_cast<TextStyle*>((ptr));
-    instance->setLocale(skString(env, locale));
+    TextStyle* instance = reinterpret_cast<TextStyle*>(ptr);
+    std::cout << "TEXT STYLE: " << skString(locale).c_str() << "\r\n";
```
VeselovAlex, line 308: Is is a debug print?

skiko/src/commonMain/kotlin/org/jetbrains/skia/paragraph/TextStyle.kt
```diff
@@ -110,10 +110,24 @@ class TextStyle internal constructor(ptr: NativePointer) : Managed(ptr, _Finaliz
...
         get() = try {
             Stats.onNativeCall()
-            _nGetDecorationStyle(_ptr)
+            val decorationStyleData = withResult(IntArray(4)) {
+                _nGetDecorationStyle(_ptr, it)
+            }
+
+            val decorationStyleFlags = decorationStyleData[0]
+
+            DecorationStyle(
```
eymar, line 122: perhaps making some static method in `DecorationStyle.kt` to create `DecorationStyle` from  decorationStyleData might help to reuse this logic in other places when/if needed?

skiko/src/commonMain/kotlin/org/jetbrains/skia/paragraph/TextStyle.kt
```diff
@@ -110,10 +110,24 @@ class TextStyle internal constructor(ptr: NativePointer) : Managed(ptr, _Finaliz
         }
     }
 
-    var decorationStyle: org.jetbrains.skia.paragraph.DecorationStyle
+    var decorationStyle: DecorationStyle
         get() = try {
             Stats.onNativeCall()
-            _nGetDecorationStyle(_ptr)
+            val decorationStyleData = withResult(IntArray(4)) {
```
VeselovAlex, line 116: Maybe it will be better to move deserialization into `DecorationStylez, like `Rect.Companion.fromInteropPointer`

VeselovAlex (approved):

skiko/src/commonMain/kotlin/org/jetbrains/skia/paragraph/TextStyle.kt
```diff
@@ -359,10 +378,31 @@ class TextStyle internal constructor(ptr: NativePointer) : Managed(ptr, _Finaliz
         return this
     }
 
+    private inline fun Float.asNumberOrNull(): Float? = if (isNaN()) null else this
+
     val fontMetrics: FontMetrics
         get() = try {
             Stats.onNativeCall()
-            _nGetFontMetrics(_ptr)
+            val fontMetricsData = withResult(FloatArray(15)) {
```
VeselovAlex, line 386: Same as above

skiko/src/nativeJsMain/cpp/paragraph/TextStyle.cc
Schahen: my bad,  removed

skiko/src/commonMain/kotlin/org/jetbrains/skia/paragraph/TextStyle.kt
olonho: Agree, generally extensions on primitives better be avoided for cases like that.

Schahen: @olonho @VeselovAlex yep, that totally make sense - I've moved all to the fromInteropPointer Companion extension functions (it seem to me that even Companion object per se should be agnostic of such  functions) - regarding the fact that this PR otherwise is approved I'll merge it after the pipeline check pass.

If you have strong arguments in favor of keeping such functions inside the Companion itself - let's discuss - if there are strong arguments I'll just will introduce this changed

PR #374, opened 2021-11-11
Run ios tests separately

VeselovAlex:


.github/workflows/blank.yml
```diff
@@ -30,7 +30,29 @@ jobs:
...
+          ./gradlew -Pskiko.native.enabled=true publishToMavenLocal
+      - uses: actions/upload-artifact@v2
+        with:
+          name: test-reports-macos
+          path: ./skiko/build/reports/tests
+          retention-days: 5
+
+   ios:
+    # The type of runner that the job will run on
+    runs-on: macos-10.15
```
olonho, line 43: Let's make it macos-11 instead (K/N will not work on 10.15 after Kotlin 1.6)

VeselovAlex: Set both to `macos-11`

olonho (approved):

PR #375, opened 2021-11-12
Define skiko platforms

AlexeyTsvetkov:


olonho:
Guess we'd better make it more local, and express build file in therms of current MPP vocabulary.

PR #376, opened 2021-11-12
Implement `skottie::Logger` for native

VeselovAlex:


skiko/src/commonMain/kotlin/org/jetbrains/skia/impl/Native.kt
```diff
@@ -150,6 +150,17 @@ inline fun withStringReferenceResult(block: () -> NativePointer): String {
     return string.toString()
 }
 
+@Suppress("NON_PUBLIC_CALL_FROM_PUBLIC_INLINE")
+inline fun withStringReferenceNullableResult(block: () -> NativePointer): String? {
+    val ptr = block()
+    if (ptr == Native.NullPointer) return null
+
+    println("PTR $ptr")
```
olonho, line 158: Debug print.

VeselovAlex: Fixed

olonho (approved):

PR #379, opened 2021-11-12
Implement Surface and DirectContext for native

VeselovAlex:


olonho:
Seems ```/tmp/included15304069549312230411/skiko-native-bridges-linux-x64.a(TestHelpers.o):TestHelpers.cc:function org_jetbrains_skiko_tests_TestHelpers__1nDeleteTestGlContext: error: undefined reference to 'XCloseDisplay'``` hints that we need to pass `-lX11` when linking native example.

skiko/build.gradle.kts
```diff
@@ -378,6 +378,9 @@ fun configureNativeTarget(os: OS, arch: Arch, target: KotlinNativeTarget) {
                     "-linker-option", "-L/usr/lib/x86_64-linux-gnu",
                     "-linker-option", "-lfontconfig",
                     "-linker-option", "-lGL",
+                    // TODO This used for test purposes only, maybe replace with dlopen/dlsym
+                    "-linker-option", "-lX11",
+                    "-linker-option", "-lGLX",
```
olonho, line 383: Better configure those flags on Linux test compilation only, not here.

skiko/src/nativeJsMain/cpp/DirectContext.cc
```diff
@@ -25,12 +35,22 @@ SKIKO_EXPORT KNativePointer org_jetbrains_skia_DirectContext__1nMakeMetal
     sk_sp<GrDirectContext> instance = GrDirectContext::MakeMetal(backendContext);
     return reinterpret_cast<KNativePointer>(instance.release());
 }
-#endif
+#endif // SK_METAL
 
+#ifdef SK_DIRECT3D
```
olonho, line 40: Maybe better always have a function implementation, but throw an exception similar to TODO() in such a case.

skiko/src/nativeJsMain/cpp/common.h
```diff
@@ -129,8 +129,11 @@ KNativePointer ptrToInterop(T* ptr) {
 #ifdef __clang__
 __attribute__((noreturn))
```
olonho, line 130: Guess would be better to make it like that

```
#ifdef __clang__
#define NO_RETURN __attribute__((noreturn))
#elif defined(_MSC_VER) 
#define NO_RETURN __attribute__((noreturn))
#else
#define NO_RETURN __declspec(noreturn)
#endif
```
and declare functions unconditionally.

skiko/src/nativeJsTest/cpp/TestHelpers.cc
```diff
@@ -51,7 +51,9 @@ SKIKO_EXPORT KNativePointer org_jetbrains_skiko_tests_TestHelpers__nStringByInde
         case 0: return new SkString("Hello");
         case 1: return new SkString("Привет");
         case 2: return new SkString("你好");
-        default: TODO("unknown");
+        default:
+            TODO("unknown");
```
olonho, line 55: No needed, if proper prototype on MSVC.

skiko/src/nativeJsTest/cpp/TestHelpers.cc
```diff
@@ -71,3 +73,113 @@ SKIKO_EXPORT KNativePointer org_jetbrains_skiko_tests_TestHelpers__1nWriteArrays
...
+SKIKO_EXPORT void org_jetbrains_skiko_tests_TestHelpers__1nDeleteTestGlContext(KNativePointer ptr);
+SKIKO_EXPORT void org_jetbrains_skiko_tests_TestHelpers__1nMakeGlContextCurrent(KNativePointer ptr);
+SKIKO_EXPORT void org_jetbrains_skiko_tests_TestHelpers__1nGlContextSwapBuffers(KNativePointer ptr);
+
+SKIKO_EXPORT KNativePointer org_jetbrains_skiko_tests_TestHelpers__1nGlContextGetFinalizer() {
+    return reinterpret_cast<KNativePointer>(org_jetbrains_skiko_tests_TestHelpers__1nDeleteTestGlContext);
+}
+
+#ifdef SK_GL
+#ifdef __linux__
```
olonho, line 88: Better `SK_BUILD_FOR_LINUX`

skiko/src/nativeJsMain/cpp/common.h
```diff
@@ -126,16 +126,15 @@ KNativePointer ptrToInterop(T* ptr) {
     return ptr;
 }
 
-#ifdef __clang__
-__attribute__((noreturn))
-void TODO(const char*);
-__attribute__((noreturn))
-void SKIKO_ASSERT(bool, const char*);
+#if defined(__clang__) || defined(_MSC_VER)
```
olonho, line 129: Guess for _MSC_VER we shall use `__declspec(noreturn)`, it's Visual Studio compiler which didn't know clang attributes, IIRC.

olonho (approved):

PR #384, opened 2021-11-15
Configure and inroduce ios and macos targets only if we are building on MacOS

Schahen:
This fixes my troubles on linux machine but I do realize that `if (hostOs == OS.MacOS) {` check was removed for some reason so if it's not the solution - let's discuss.


AlexeyTsvetkov:
Fixed in master

olonho:
@Schahen maybe abandon this one?

PR #385, opened 2021-11-15
Fix fullscreen mode on macos.

Rsedaikin:
- fixed case where fullscreen mode is set before the window becomes visible and nothing happens after it becomes visible
see: https://github.com/JetBrains/compose-jb/issues/1134

### DIFF
--- a/skiko/src/jvmMain/kotlin/org/jetbrains/skiko/PlatformOperations.kt
+++ b/skiko/src/jvmMain/kotlin/org/jetbrains/skiko/PlatformOperations.kt
@@ -6,6 +6,34 @@ import java.awt.event.ComponentAdapter
 import java.awt.event.ComponentEvent
 import javax.swing.SwingUtilities
 
+internal open class FullscreenAdapter(
+    val backedLayer: HardwareLayer
+): ComponentAdapter() {
+    private var _isFullscreenDispatched = false
+    private var _isFullscreen: Boolean = false
+    var fullscreen: Boolean
+        get() = backedLayer.fullscreen
+        set(value) {
+            _isFullscreen = value
+            val window = SwingUtilities.getRoot(backedLayer) as Window
+            if (window.isVisible) {
+                backedLayer.fullscreen = value
+            } else {
+                _isFullscreenDispatched = value
+            }
+        }
+
+    override fun componentShown(e: ComponentEvent) {
+        if (_isFullscreenDispatched) {
+            backedLayer.fullscreen = _isFullscreenDispatched
+        }
+    }
+
+    override fun componentResized(e: ComponentEvent) {
+        _isFullscreen = backedLayer.fullscreen
+    }
+}
+
 internal interface PlatformOperations {
     fun isFullscreen(component: Component): Boolean
     fun setFullscreen(component: Component, value: Boolean)
--- a/skiko/src/jvmMain/kotlin/org/jetbrains/skiko/PlatformOperations.kt
+++ b/skiko/src/jvmMain/kotlin/org/jetbrains/skiko/PlatformOperations.kt
@@ -2,6 +2,8 @@ package org.jetbrains.skiko
 
 import java.awt.Component
 import java.awt.Window
+import java.awt.event.ComponentAdapter
+import java.awt.event.ComponentEvent
 import javax.swing.SwingUtilities
 
 internal interface PlatformOperations {
@@ -20,6 +22,17 @@ internal val platformOperations: PlatformOperations by lazy {
                 }
 
                 override fun setFullscreen(component: Component, value: Boolean) {
+                    // MacOS specific: if fullscreen is set before the window is visible,
+                    // we will add a listener to set fullscreen after the window is visible
+                    val window = SwingUtilities.getRoot(component) as Window
+                    if (value && !window.isVisible) {
+                        window.addComponentListener(object : ComponentAdapter() {
+                            override fun componentShown(e: ComponentEvent) {
+                                println("show window")
+                                osxSetFullscreenNative(component, value)
+                            }
+                        })
+                    }
                     osxSetFullscreenNative(component, value)
                 }
 
--- a/skiko/src/jvmMain/kotlin/org/jetbrains/skiko/PlatformOperations.kt
+++ b/skiko/src/jvmMain/kotlin/org/jetbrains/skiko/PlatformOperations.kt
@@ -16,22 +16,13 @@ internal interface PlatformOperations {
 
 internal val platformOperations: PlatformOperations by lazy {
     when (hostOs) {
-        OS.MacOS -> object: PlatformOperations {
+        OS.MacOS -> {
+            object: PlatformOperations {
                 override fun isFullscreen(component: Component): Boolean {
                     return osxIsFullscreenNative(component)
                 }
 
                 override fun setFullscreen(component: Component, value: Boolean) {
-                    // MacOS specific: if fullscreen is set before the window is visible,
-                    // we will add a listener to set fullscreen after the window is visible
-                    val window = SwingUtilities.getRoot(component) as Window
-                    if (value && !window.isVisible) {
-                        window.addComponentListener(object : ComponentAdapter() {
-                            override fun componentShown(e: ComponentEvent) {
-                                osxSetFullscreenNative(component, value)
-                            }
-                        })
-                    }
                     osxSetFullscreenNative(component, value)
                 }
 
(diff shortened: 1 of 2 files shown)


skiko/src/jvmMain/kotlin/org/jetbrains/skiko/PlatformOperations.kt
```diff
@@ -20,6 +22,17 @@ internal val platformOperations: PlatformOperations by lazy {
                 }
 
                 override fun setFullscreen(component: Component, value: Boolean) {
+                    // MacOS specific: if fullscreen is set before the window is visible,
+                    // we will add a listener to set fullscreen after the window is visible
+                    val window = SwingUtilities.getRoot(component) as Window
+                    if (value && !window.isVisible) {
+                        window.addComponentListener(object : ComponentAdapter() {
+                            override fun componentShown(e: ComponentEvent) {
+                                println("show window")
```
olonho, line 31: println not needed.

Rsedaikin: Done.

skiko/src/jvmMain/kotlin/org/jetbrains/skiko/PlatformOperations.kt
```diff
@@ -20,6 +22,16 @@ internal val platformOperations: PlatformOperations by lazy {
                 }
 
                 override fun setFullscreen(component: Component, value: Boolean) {
+                    // MacOS specific: if fullscreen is set before the window is visible,
+                    // we will add a listener to set fullscreen after the window is visible
+                    val window = SwingUtilities.getRoot(component) as Window
+                    if (value && !window.isVisible) {
+                        window.addComponentListener(object : ComponentAdapter() {
```
igordmn, line 29: we add a listener every time when we call setFullscreen and window isn't visible:
```
val layer = SkiaLayer()
layer.isFullscreen = true
layer.isFullscreen = false
layer.isFullscreen = true
layer.isVisible = true
layer.isVisible = false
layer.isFullscreen = false
layer.isFullscreen = true
```

can we call `osxSetFullscreenNative` in `SkiaLayer.addNotify`?

igordmn: Also, this will return false:
```
val layer = SkiaLayer()
layer.isVisible = true
layer.isFullscreen = true
println(layer.isFullscreen)
```

It is an issue of Skiko, but I am not sure, if it is issue for Compose. Theoretically it can be, as we check layer.isFullscreen in resize listener, which can be triggered after we set layer.isFullscreen = true.

To fix that, we need to introduce intermediate state, as you done [here](https://github.com/JetBrains/skiko/commit/a71eca0fdd03caf248a99e4ca2efb537ead76736#diff-1782b97ce06bbcf5e55689e5c50647096808e91aef66c3a60fef15beb477441eR47), but with careful handling of concurrency.

igordmn: I would implement something like this:
```
class SkiaLayer {
        var isFullscreen: Boolean = false
           set(value) {
              field = value
              platformOperation.setFullscreen(value)
           }

        private var window: Window? = null
        private var windowListener = object : WindowAdapter {
            override fun windowResized(e: WindowEvent) {
                isFullscreen = platformOperations.isFullscreen()
            }
        }

        override fun addNotify() {
            super.addNotify()
            ...
            window = SwingUtilities.getWindowAncestor(this)
            window?.addWindowListener(windowListener)
            platformOperation.setFullscreen(isFullscreen)
        }

        override fun removeNotify() {
            window?.removeWindowListener(windowListener)
            window = null
            ...
            super.removeNotify()
        }
}
```

skiko/src/jvmMain/kotlin/org/jetbrains/skiko/SkiaLayer.jvm.kt
```diff
@@ -135,9 +155,15 @@ actual open class SkiaLayer internal constructor(
         get() = backedLayer.windowHandle
 
     actual var fullscreen: Boolean
-        get() = backedLayer.fullscreen
+        get() = if (fullscreenDispatcher.isFullscreenDispatched) true
```
olonho, line 158: Looks somewhat fragile, cannot we always keep proper state on `backedLayer`?

skiko/src/jvmMain/kotlin/org/jetbrains/skiko/SkiaLayer.jvm.kt
```diff
@@ -135,9 +155,15 @@ actual open class SkiaLayer internal constructor(
...
+            else backedLayer.fullscreen
         set(value) {
-            backedLayer.fullscreen = value
+            val window = SwingUtilities.getRoot(this) as Window
+            if (!window.isVisible) {
```
igordmn, line 162: The logic is very complex, as it have many branching paths (`if`'s). I have found a case where the SkiaLayer doesn't behave as expected:
```
layer.fullscreen = true
window.isVisible = true
layer.fullscreen = false
println(layer.fullscreen) // prints true
```

Also `isFullscreen` will be lying to as, if the user exit fullscreen manually, after we set it programmatically. We should subscribe to resize in AWT (on to window fullscreen in objective c code).

skiko/src/jvmMain/kotlin/org/jetbrains/skiko/SkiaLayer.jvm.kt
```diff
@@ -92,13 +94,31 @@ actual open class SkiaLayer internal constructor(
         }
     }
 
+    private var fullscreenDispatcher = object : ComponentAdapter() {
+        var isFullscreenDispatched: Boolean = false
+        override fun componentShown(e: ComponentEvent) {
```
igordmn, line 99: Is componentShown mandatory? Changing fullscreen doesn't work in addNotify?

When we in `addNotify`, we already have the native window created. If we change fullscreen only after it is showed, we will have fullscreen change glitch at the start of the application.

skiko/src/jvmMain/kotlin/org/jetbrains/skiko/SkiaLayer.jvm.kt
```diff
@@ -98,10 +98,13 @@ actual open class SkiaLayer internal constructor(
         var isFullscreenDispatched: Boolean = false
         override fun componentShown(e: ComponentEvent) {
             if (isFullscreenDispatched) {
```
igordmn, line 100: We probably should check `_isFullscreen` here (and remove `isFullscreenDispatched`). Because:

1. This code won't preserve fullscreen:
```
layer.isFullscreen = true
window.isVisible = true
window.isVisible = false
window.isVisible = true
```

2. When the user manually exits fullscreen, `isFullscreenDispatched` will stay `true`.

igordmn: > We probably should check _isFullscreen

Not with if, but just:
```
backedLayer.isFullscreen = _isFullscreen
```

or
```
if (backedLayer.isFullscreen != _isFullscreen) {
   backedLayer.isFullscreen = _isFullscreen
}
```

skiko/src/jvmMain/kotlin/org/jetbrains/skiko/SkiaLayer.jvm.kt
```diff
@@ -134,9 +159,12 @@ actual open class SkiaLayer internal constructor(
     val windowHandle: Long
         get() = backedLayer.windowHandle
 
+    private var _isFullscreen = false
     actual var fullscreen: Boolean
-        get() = backedLayer.fullscreen
+        get() = _isFullscreen
```
olonho, line 164: This way `backedLayer.fullscreen` becomes bogus, we may want to rename it to something like `fun requestPlatformFullscreen(newValue: Boolean)` and `fun checkPlatformFullscreen(): Boolean`.

igordmn: > something like fun requestPlatformFullscreen(newValue: Boolean) and fun checkPlatformFullscreen(): Boolean

I don't agree that we should introduce these functions on SkiaLayer level. We can discuss online.

The purpose of the property is to simplify the interaction with the native window, as many mainstream frameworks do.

olonho (approved):

igordmn (approved):

PR #387, opened 2021-11-16
Array serialization revisited

Schahen:


skiko/src/commonMain/cpp/common/include/mppinterop.h
```diff
@@ -22,6 +22,10 @@ namespace skikoMpp {
         int getClustersLength(SkTextBlob* instance);
         bool getClusters(SkTextBlob* instance, int* clusters);
     }
+
+    namespace finalizers {
+        void deleteString(void* instance);
```
olonho, line 27: SkString*?

skiko/src/commonMain/kotlin/org/jetbrains/skia/Typeface.kt
```diff
@@ -339,11 +339,21 @@ class Typeface internal constructor(ptr: NativePointer) : RefCnt(ptr) {
...
+                Stats.onNativeCall()
+                arrayDecoder = ArrayDecoder(_nGetFamilyNames(_ptr), ManagedString_nGetFinalizer())
+                val size = arrayDecoder.size
+                (0 until size / 2).map { i ->
+                    val name = arrayDecoder.release(2 * i)
+                    val language = arrayDecoder.release(2 * i + 1)
+                    FontFamilyName(ManagedString(name).toString(), ManagedString(language).toString())
+                }.toTypedArray()
+            } finally {
+                arrayDecoder?.dispose()
```
olonho, line 354: Why not use `interopScope` or create similar `arrayDecoderScope`?

VeselovAlex:
Please delete log file

skiko/src/jvmMain/cpp/common/StdVectorDecoder.cc
```diff
@@ -0,0 +1,33 @@
...
+
+extern "C" JNIEXPORT jint JNICALL Java_org_jetbrains_skia_StdVectorDecoderKt_StdVectorDecoder_1nGetArraySize
+    (JNIEnv* env, jclass jclass, jlong ptr) {
+        std::vector<void*>* vec = reinterpret_cast<std::vector<void*> *>(ptr);
+        return static_cast<jint>(vec->size());
+    }
+
+extern "C" JNIEXPORT jlong JNICALL Java_org_jetbrains_skia_StdVectorDecoderKt_StdVectorDecoder_1nReleaseElement
+    (JNIEnv* env, jclass jclass, jlong ptr, jint index) {
+        std::vector<void*>* vec = reinterpret_cast<std::vector<void*> *>(ptr);
```
olonho, line 12: Probably 
```
auto& vec = *reinterpret_cast<std::vector<void*> *>(ptr);
auto res = vec[index];
vec[index] = nullptr;
``` 
would be a bit more readable?

skiko/src/jvmMain/cpp/common/StdVectorDecoder.cc
```diff
@@ -0,0 +1,33 @@
...
+        auto res = (*vec)[index];
+        (*vec)[index] = nullptr;
+        return reinterpret_cast<jlong>(res);
+    }
+
+extern "C" JNIEXPORT void JNICALL Java_org_jetbrains_skia_StdVectorDecoderKt_StdVectorDecoder_1nDisposeArray
+    (JNIEnv* env, jclass jclass, jlong ptr, jlong disposePtr) {
+        std::vector<void*>* vec = reinterpret_cast<std::vector<void*> *>(ptr);
+
+        void (*dctr)(void*) = reinterpret_cast<void (*)(void*)>(disposePtr);
```
olonho, line 22: Usually it's called `dtor`.

skiko/src/commonMain/kotlin/org/jetbrains/skia/Typeface.kt
```diff
@@ -339,11 +339,21 @@ class Typeface internal constructor(ptr: NativePointer) : RefCnt(ptr) {
...
+        get() {
+            var arrayDecoder: ArrayDecoder? = null
+            return try {
+                Stats.onNativeCall()
+                arrayDecoder = ArrayDecoder(_nGetFamilyNames(_ptr), ManagedString_nGetFinalizer())
+                val size = arrayDecoder.size
+                (0 until size / 2).map { i ->
+                    val name = arrayDecoder.release(2 * i)
+                    val language = arrayDecoder.release(2 * i + 1)
+                    FontFamilyName(ManagedString(name).toString(), ManagedString(language).toString())
```
VeselovAlex, line 351: Please use withStringResult as discussed

skiko/src/commonMain/cpp/common/include/mppinterop.h
Schahen: this is reintepret_casted in deleteString body so that we don't need to be type-agnostic while disposing arbitrary pointers.

skiko/src/commonMain/kotlin/org/jetbrains/skia/Typeface.kt
Schahen: we haven't discussed it ) we've discussed that ManagedString better to invoke "use" block - which it do now.
It's overkill to invoke block execution in place where we just need to wrap value - however to have a consistent converter still make sense, so I've introduced `withStringResult` to which one can directly pass pointer.

skiko/src/jvmMain/cpp/common/StdVectorDecoder.cc
Schahen: bad habits die hard, this is definitely more readable, pushed

skiko/src/jvmMain/cpp/common/StdVectorDecoder.cc
Schahen: fixed

skiko/src/commonMain/cpp/common/include/mppinterop.h
VeselovAlex: Maybe it can be a something like `KInteropPointer` to mark that pointers are received from Kotlin side?

skiko/src/commonMain/kotlin/org/jetbrains/skia/Typeface.kt
Schahen: regarding the fact that we will using this in quite many places yep, make sense, pushed

skiko/src/commonMain/cpp/common/include/mppinterop.h
olonho: Hmm, not sure if I understand. `deleteString` can only delete a string, so signature shall reflect that. When we want to cast that to generic `void ()(void*)` type - cast shall happen exactly at that moment.

olonho (approved):

skiko/src/commonMain/cpp/common/include/mppinterop.h
VeselovAlex: IMHO destructor function types should be `void (*)(KNativePointer)` or `void (*)(KinteropPointer)` to show, that they must be called on pointers received from Kotlin. We return `KNativePointer` from Cpp, so we probably should use this type everywhere we are receiving pointers

PR #396, opened 2021-11-18
Commonize StrutStyle::fontFamilies and StrutStyle::setFontFamilies

Schahen:


skiko/src/commonMain/kotlin/org/jetbrains/skia/paragraph/StrutStyle.kt
```diff
@@ -36,14 +38,18 @@ class StrutStyle internal constructor(ptr: NativePointer) : Managed(ptr, _Finali
     val fontFamilies: Array<String>
         get() = try {
             Stats.onNativeCall()
-            _nGetFontFamilies(_ptr)
+            arrayDecoderScope({ArrayDecoder(_nGetFontFamilies(_ptr), ManagedString_nGetFinalizer())}) { arrayDecoder ->
```
VeselovAlex, line 41: This becomes an idiom, so maybe we should make a helper from it?

VeselovAlex (approved):

skiko/src/commonMain/kotlin/org/jetbrains/skia/paragraph/StrutStyle.kt
Schahen: Let's wait till we get even more usages of ArrayDecoder and see what can be generalized - I mean your suggestion with moving release logic entirely to the backend has a lot of sense and we'll see how it goes )))

VeselovAlex: @Schahen I mean we can move these 3 lines into a helper function (like `fromInteropStringArray`) and use it here and in the other places where we a receiving strings as an array. Also there is generic implementation of `fromInterop` for `ArrayInteropDecoder` may be we can use it

skiko/src/commonTest/kotlin/org/jetbrains/skia/StrutStyleTests.kt
```diff
@@ -16,6 +17,10 @@ class StrutStyleTests {
             strutStyle.fontStyle = FontStyle(300, 4, FontSlant.ITALIC)
 
             assertEquals(FontStyle(300, 4, FontSlant.ITALIC), strutStyle.fontStyle)
+
+
+            strutStyle.setFontFamilies(arrayOf("MonacoShmonaco"))
```
olonho, line 22: Is it actual font name?

olonho (approved):

skiko/src/commonTest/kotlin/org/jetbrains/skia/StrutStyleTests.kt
Schahen: it's not but in the framework of validation of setting fonts this test is sufficient (we test actual font loading in some different tests)

PR #397, opened 2021-11-18
Enable JS target for Intellij sync

AlexeyTsvetkov:


olonho (approved):

olonho (commented):
Although originally it was only for WASM compilation, which didn't affect JS targets.

PR #400, opened 2021-11-19
Add support of `TextStyle::baselineShift` 

VeselovAlex:


skiko/src/commonMain/kotlin/org/jetbrains/skia/paragraph/TextStyle.kt
```diff
@@ -286,6 +285,23 @@ class TextStyle internal constructor(ptr: NativePointer) : Managed(ptr, _Finaliz
...
+            Stats.onNativeCall()
+            TextStyle_nGetBaselineShift(_ptr)
+        } finally {
+            reachabilityBarrier(this)
+        }
+        set(value) {
+            setBaselineShift(value)
+        }
+
+    fun setBaselineShift(baselineShift: Float): TextStyle {
```
olonho, line 299: Guess we do not need setter returning this, as in Kotlin better idiom is using apply and not setter chaining.

VeselovAlex: Agree, but such methods are used for similar fields in this class, so I've implemented it

olonho (approved):

PR #408, opened 2021-11-25
Fixed case where SkiaLayer size could be 0x0 and surface was not created.

Rsedaikin:


skiko/src/jvmMain/kotlin/org/jetbrains/skiko/context/Direct3DContextHandler.kt
```diff
@@ -10,6 +10,7 @@ import java.lang.ref.Reference
 internal class Direct3DContextHandler(layer: SkiaLayer) : JvmContextHandler(layer) {
     private val bufferCount = 2
     private var surfaces: Array<Surface?> = arrayOfNulls(bufferCount)
+    private fun isSurfacesNull() = surfaces.filterNotNull().isEmpty()
```
igordmn, line 13: Usually better to write `surfaces.any { it == null }`, but probably it shouldn't affect performance, as the count of elements is small.

Rsedaikin: Done.

igordmn (approved):

PR #411, opened 2021-11-26
Commonize BreakIterator::setText and BreakIterator::clone

Schahen:
Following things to consider regarding this PR:

1) tests are still ignored in JS but reasons are, however that with very high probability its the C++ API that behaves differently and causes troubles, not our interop serialization - to check this, I've hardcoded input data. (the other possibility would be that for some reason pointer to UChar array is lost on JS for some reason)

2)  ruleStatuses is yet to be investigated - the thing is that it's failing with U_BUFFER_OVERFLOW_ERROR on existing JVM target - which contradicts to the documentation

3) U16String has lesser methods and it looks like that next step of evolution would be to get rid of it completely.


skiko/src/commonMain/kotlin/org/jetbrains/skia/BreakIterator.kt
```diff
@@ -502,8 +504,8 @@ class BreakIterator internal constructor(ptr: NativePointer) : Managed(ptr, _Fin
     fun setText(text: String?) {
         try {
             Stats.onNativeCall()
-            _text = U16String(text)
-            _nSetText(_ptr, getPtr(_text))
+            _text?.close()
+            _text = interopScope { U16String(_nSetText(_ptr, toInterop(text?.let { ShortArray(text.length) { text[it].code.toShort() } }), text?.length ?: 0)) }
```
VeselovAlex, line 508: Could you split this one-liner into few lines?

olonho (approved):

PR #412, opened 2021-11-28
Fix getRectsForRange (attempt 2)

igordmn:
The previous fix was wrong, we don't fill all indices of the array

Fixes https://github.com/JetBrains/compose-jb/issues/1308#issuecomment-981139635

### DIFF
--- a/skiko/src/jvmMain/cpp/common/paragraph/Paragraph.cc
+++ b/skiko/src/jvmMain/cpp/common/paragraph/Paragraph.cc
@@ -82,15 +82,20 @@ extern "C" JNIEXPORT void JNICALL Java_org_jetbrains_skia_paragraph_ParagraphKt_
 extern "C" JNIEXPORT jobjectArray JNICALL Java_org_jetbrains_skia_paragraph_ParagraphKt__1nGetRectsForRange
   (JNIEnv* env, jclass jclass, jlong ptr, jint start, jint end, jint rectHeightStyle, jint rectWidthStyle) {
     Paragraph* instance = reinterpret_cast<Paragraph*>(static_cast<uintptr_t>(ptr));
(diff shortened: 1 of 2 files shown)


skiko/src/jvmMain/cpp/common/paragraph/Paragraph.cc
```diff
@@ -82,15 +82,20 @@ extern "C" JNIEXPORT void JNICALL Java_org_jetbrains_skia_paragraph_ParagraphKt_
 extern "C" JNIEXPORT jobjectArray JNICALL Java_org_jetbrains_skia_paragraph_ParagraphKt__1nGetRectsForRange
   (JNIEnv* env, jclass jclass, jlong ptr, jint start, jint end, jint rectHeightStyle, jint rectWidthStyle) {
     Paragraph* instance = reinterpret_cast<Paragraph*>(static_cast<uintptr_t>(ptr));
-    std::vector<TextBox> rects = instance->getRectsForRange(start, end, static_cast<RectHeightStyle>(rectHeightStyle), static_cast<RectWidthStyle>(rectWidthStyle));
-    jobjectArray rectsArray = env->NewObjectArray((jsize) rects.size(), skija::paragraph::TextBox::cls, nullptr);
-    for (int i = 0; i < rects.size(); ++i) {
-        TextBox box = rects[i];
+    std::vector<TextBox> originalRects = instance->getRectsForRange(start, end, static_cast<RectHeightStyle>(rectHeightStyle), static_cast<RectWidthStyle>(rectWidthStyle));
```
olonho, line 85: What problem does it change? Why no fix in native/js needed?

olonho (approved):

skiko/src/jvmMain/cpp/common/paragraph/Paragraph.cc
igordmn: In the previous code, we get rects, create java array with rects.size, filter rects

In the current code, we get rects, filter rects, create java array with filteredRects.size

In native that is not needed, because code is correct there

PR #415, opened 2021-11-30
Implement Shaper::shape with RunHandler

VeselovAlex:


olonho (approved):

skiko/src/jsMain/kotlin/org/jetbrains/skia/Actuals.js.kt
```diff
@@ -27,7 +29,13 @@ actual class Matcher constructor(private val regex: Regex, private val input: Ch
     actual fun matches(): Boolean = matches
 }
 
-actual fun defaultLanguageTag(): String = TODO()
+@Suppress("RedundantNullableReturnType")
+val LANG: String by lazy {
```
olonho, line 33: What if system local changes? Doesn't look like a big deal, but we need to provide an API to force reevaluation.

PR #418, opened 2021-12-01
Pass status code for ubrk_ operations to the front and fail in case we got non-zero error code

Schahen:


skiko/src/commonMain/kotlin/org/jetbrains/skia/BreakIterator.kt
```diff
@@ -523,15 +527,30 @@ class BreakIterator internal constructor(ptr: NativePointer) : Managed(ptr, _Fin
...
 }
 
+private fun withErrorGuard(message: String, block: InteropScope.(InteropPointer) -> NativePointer): NativePointer {
+    val errorCode = IntArray(1)
+    return interopScope {
+        val handle = toInterop(errorCode)
+        val res = block.invoke(this, handle)
+        handle.fromInterop(errorCode)
+        if (errorCode[0] > 0) {
+            throw RuntimeException("$message; ubrk_* operation failed with status ${errorCode}")
```
olonho, line 537: Maybe better `"$message: operation failed with status ${errorCode}"` as code could be reused in other places as well.

skiko/src/nativeJsMain/cpp/BreakIterator.cc
```diff
@@ -116,11 +120,12 @@ SKIKO_EXPORT KInt* org_jetbrains_skia_BreakIterator__1nGetRuleStatuses
 
 
 SKIKO_EXPORT KNativePointer org_jetbrains_skia_BreakIterator__1nSetText
-  (KNativePointer ptr, KChar* textArr, KInt len) {
+  (KNativePointer ptr, KChar* textArr, KInt len, KInt* errorCode) {
     UBreakIterator* instance = reinterpret_cast<UBreakIterator*>(ptr);
     std::vector<UChar>* text = new std::vector<UChar>(textArr, textArr + len);
 
     UErrorCode status = U_ZERO_ERROR;
     ubrk_setText(instance, text->data(), len, &status);
```
olonho, line 128: Why not just
```
errorCode[0] = U_ZERO_ERROR;
ubrk_setText(instance, text->data(), len, errorCode);
```?

Comment (Schahen) [reply skiko/src/nativeJsMain/cpp/BreakIterator.cc:128]: @olonho if I got the idea right then, instead of 
```cpp
  UErrorCode status = U_ZERO_ERROR;
    UBreakIterator* instance;
    if (localeStr == nullptr)
      instance = ubrk_open(static_cast<UBreakIteratorType>(type), uloc_getDefault(), nullptr, 0, &status);
    else {
      instance = ubrk_open(static_cast<UBreakIteratorType>(type), skString(localeStr).c_str(), nullptr, 0, &status);
    }

    errorCode[0] = status;
    if (U_FAILURE(status))
      return 0;
    else
      return reinterpret_cast<KNativePointer>(instance);
```

we'll have pretty much something like this
```cpp
errorCode[0] = U_ZERO_ERROR;
    UBreakIterator* instance;
    if (localeStr == nullptr)
      instance = ubrk_open(static_cast<UBreakIteratorType>(type), uloc_getDefault(), nullptr, 0, reinterpret_cast<UErrorCode*>(errorCode));
    else {
      instance = ubrk_open(static_cast<UBreakIteratorType>(type), skString(localeStr).c_str(), nullptr, 0, reinterpret_cast<UErrorCode*>(errorCode));
    }

    if (U_FAILURE(reinterpret_cast<UErrorCode>(errorCode[0])))
      return 0;
    else
      return reinterpret_cast<KNativePointer>(instance);
```

So, for the sake of not having status variable we sacrifice a bit readability with introducing reinterpret_cast and even explicit casting

Schahen: For the JVM, however, it totally make sense - pushed

skiko/src/commonMain/kotlin/org/jetbrains/skia/BreakIterator.kt
Schahen: I agree - pushed

olonho:
Could you please make simplification with using local int instead of fake 1-element array, guess it will increase readability. Could add commit explaining the idea if needed.

olonho (approved):

PR #420, opened 2021-12-02
use FrameDispatcher with AnimationFramesCoroutineDispatcher for js

eymar:


skiko/src/jsMain/kotlin/org/jetbrains/skiko/CanvasRenderer.kt
```diff
@@ -32,15 +31,13 @@ abstract class CanvasRenderer constructor(htmlCanvas: HTMLCanvasElement, val wid
...
 
     fun needRedraw() {
-        window.requestAnimationFrame { timestamp ->
-            GL.makeContextCurrent(contextPointer)
-            canvas.clear(-1)
-            drawFrame(timestamp)
-            surface.flushAndSubmit()
-            context.flush()
-        }
+        GL.makeContextCurrent(contextPointer)
```
igordmn, line 37: `needRedraw` by itself shouldn't do any drawing operations (`canvas.clear(-1)`, `context.flush()`, etc).

It should only inform drawing loop to schedule drawing. `requestAnimationFrame` exists for this purpose, but calling it multiple times will schedule too many draw requests.

eymar: I'll open a new PR with this change

PR #426, opened 2021-12-09
Support hi-dpi for Web targets

olonho:


skiko/src/jsMain/kotlin/org/jetbrains/skiko/SkiaLayer.js.kt
```diff
@@ -83,8 +95,16 @@ actual open class SkiaLayer {
             skikoView?.onKeyboardEvent(toSkikoEvent(event, SkikoKeyboardEventKind.UP))
         })
     }
+
+    private fun setOnChangeScaleNotifier() {
+        state?.initCanvas(desiredWidth, desiredHeight, contentScale)
+        window.matchMedia("(resolution: ${contentScale}dppx)").addEventListener("change", { setOnChangeScaleNotifier() }, true)
```
VeselovAlex, line 101: Shouldn't we remove previously added event listener when it is called?

skiko/src/jsMain/kotlin/org/jetbrains/skiko/SkiaLayer.js.kt
```diff
@@ -11,7 +12,7 @@ actual open class SkiaLayer {
 
     actual var renderApi: GraphicsApi = GraphicsApi.WEBGL
     actual val contentScale: Float
-        get() = 1.0f
+        get() = window.devicePixelRatio.toFloat()
```
Schahen, line 15: not a big deal but may be we should just keep Double and not convert till it needed? contentScale is use in expression after all.

Schahen (approved):
I don't see anything wrong with this PR at all.

PR #430, opened 2021-12-16
Further commonization of skiko events.

Rsedaikin:
- added raw touch events for [uikit]
- added key code and modifiers in [SkikoInputEvent]
- added event [timestamp]
- added delta's to scroll event

### DIFF
--- a/skiko/src/awtMain/kotlin/org/jetbrains/skiko/SkikoKey.awt.kt
+++ b/skiko/src/awtMain/kotlin/org/jetbrains/skiko/SkikoKey.awt.kt
@@ -5,7 +5,7 @@ import java.awt.event.KeyEvent.*
 actual enum class SkikoKey(val value: Int) {
     KEY_UNKNOWN(-1),
     KEY_A(VK_A),
-    KEY_S(VK_B),
+    KEY_S(VK_S),
     KEY_D(VK_D),
     KEY_F(VK_F),
     KEY_H(VK_H),
--- a/skiko/src/commonMain/kotlin/org/jetbrains/skiko/SkikoView.kt
+++ b/skiko/src/commonMain/kotlin/org/jetbrains/skiko/SkikoView.kt
@@ -7,6 +7,7 @@ interface SkikoView {
     fun onKeyboardEvent(event: SkikoKeyboardEvent) = Unit
     fun onPointerEvent(event: SkikoPointerEvent) = Unit
     fun onInputEvent(event: SkikoInputEvent) = Unit
+    fun onTouchEvent(events: Set<SkikoTouchEvent>) = Unit
     fun onGestureEvent(event: SkikoGestureEvent) = Unit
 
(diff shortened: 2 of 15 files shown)


skiko/src/awtMain/kotlin/org/jetbrains/skiko/SkikoKey.awt.kt
```diff
@@ -5,7 +5,7 @@ import java.awt.event.KeyEvent.*
 actual enum class SkikoKey(val value: Int) {
     KEY_UNKNOWN(-1),
     KEY_A(VK_A),
-    KEY_S(VK_B),
+    KEY_S(VK_S),
```
olonho, line 8: :)

skiko/src/commonMain/kotlin/org/jetbrains/skiko/SkikoView.kt
```diff
@@ -7,6 +7,7 @@ interface SkikoView {
     fun onKeyboardEvent(event: SkikoKeyboardEvent) = Unit
     fun onPointerEvent(event: SkikoPointerEvent) = Unit
     fun onInputEvent(event: SkikoInputEvent) = Unit
+    fun onTouchEvent(events: Set<SkikoTouchEvent>) = Unit
```
olonho, line 10: Can we use less resource consuming structure, i.e. `Array<SkikoTouchEvent>`?

skiko/src/iosMain/kotlin/org/jetbrains/skiko/SkiaLayer.ios.kt
```diff
@@ -185,7 +185,8 @@ actual open class SkiaLayer {
 }
 
 // TODO: do properly
+actual typealias SkikoTouchPlatformEvent = UITouch
 actual typealias SkikoGesturePlatformEvent = UIEvent
-actual typealias SkikoPlatformInputEvent = UIEvent
+actual typealias SkikoPlatformInputEvent = UIPress
```
olonho, line 190: Why `UIPress`?

skiko/src/commonMain/kotlin/org/jetbrains/skiko/SkikoView.kt
Rsedaikin: Done. Made it `Array`

olonho (approved):

skiko/src/iosMain/kotlin/org/jetbrains/skiko/SkiaLayer.ios.kt
Rsedaikin: It is because on ios there is no `uievent` when you type, only methods like `override fun insertText(theText: String) {`, but instead of that we also need to pass key event (`uipress`), so I decided to set `SkikoPlatformInputEvent` as `UIPress` alias

PR #436, opened 2021-12-21
Build fixes after android awt split

AlexeyTsvetkov:


olonho:
Nice!

skiko/build.gradle.kts
```diff
@@ -1068,12 +1104,13 @@ publishing {
         allJvmRuntimeJars.forEach { entry ->
             val os = entry.key.first
             val arch = entry.key.second
-            create<MavenPublication>("skikoJvmRuntime${targetId(os, arch)}") {
+            create<MavenPublication>("skikoJvmRuntime${toTitleCase(os.id)}${toTitleCase(arch.id)}") {
```
olonho, line 1107: Not sure if I like this particular change. Readability seems better when target is clearly marked, and this captializing just makes readability worse.

skiko/buildSrc/src/main/kotlin/utils.kt
```diff
@@ -16,6 +18,21 @@ inline fun <reified T : Task> TaskContainer.registerOrGetTask(
...
+    crossinline fn: T.() -> Unit
+): TaskProvider<T> = tasks.register(taskName, T::class) { fn() }
+
+inline fun <reified T : Task> Project.registerSkikoTask(
+    actionName: String,
+    targetOs: OS,
+    targetArch: Arch,
+    crossinline fn: T.() -> Unit
+): TaskProvider<T> {
+    val taskName = joinToTitleCamelCase(actionName, targetOs.id, targetArch.id)
```
olonho, line 32: Same here.

skiko/build.gradle.kts
AlexeyTsvetkov: What's the point of readability, when different cases are mixed in inconsistent and unpredictable way?
If we could change the default case format for all tasks, I would be pro dash case.
But that's not possible, and lack of some predictable naming convention and style seriously bothers me

olonho: OK

olonho (approved):

PR #444, opened 2021-12-29
Fix SkiaJsSample build failed

thaihuynhxyz:
It seems APIs are changed and the sample needs to be updated.


samples/SkiaJsSample/src/jsMain/kotlin/org/jetbrains/skiko/sample/js/App.kt
```diff
@@ -33,8 +32,8 @@ fun main() {
             } else {
                 BouncingBalls()
             }
-            skiaLayer.renderer = GenericSkikoApp(skiaLayer, app)
-            skiaLayer.setCanvas(canvas)
+            skiaLayer.skikoView = app
```
olonho, line 35: Please do
```
skiaLayer.skikoView = GenericSkikoView(skiaLayer, app)
```
as otherwise animation is broken.

PR #448, opened 2022-01-12
AWT fonts interop support

olonho:


skiko/src/awtMain/kotlin/org/jetbrains/skiko/FontInterop.awt.kt
```diff
@@ -0,0 +1,109 @@
...
+        }
+        files
+    }
+
+    private fun cacheSystemFonts() {
+        val fontFiles = systemFontFiles
+        for (file in fontFiles) {
+            try {
+                if (!fontsMap.containsValue(file.absoluteFile)) {
+                    val f = Font.createFont(Font.TRUETYPE_FONT, FileInputStream(file.absolutePath))
```
igordmn, line 90: It should be 

```
FileInputStream(file.absolutePath).use {
    Font.createFont(Font.TRUETYPE_FONT, it)
}
```

skiko/src/awtMain/kotlin/org/jetbrains/skiko/FontInterop.awt.kt
```diff
@@ -0,0 +1,109 @@
...
+            } catch (e: IOException) {
+            }
+        }
+    }
+
+    fun findFontFile(font: Font): File? {
+        return fontsMap[font.family]
+    }
+
+    val allFontsCached: Boolean
```
igordmn, line 104: Can we make caching as implementation details, so we can change it in the future?

Also, we can get rid of `Font` from API and receive only `familyName: String`. And maybe even get rid of AWT from API.

I suggest this API:
```
class SystemFonts {
    val familyNames: List<String>    

    fun fileOf(familyName: String): File

    companion object {
       suspend fun load(): SystemFonts
    }
}
```

It is no longer global, and requires manual creation/passing, but global operations/API's has its own flaws (we can discuss).

igordmn: > Can we make caching as implementation details, so we can change it in the future?

I can explain why this is bad.

Imagine a very big application. Some developer wants to know a file of some font and finds a class `AwtFontManager`, finds the method `findFontFile`, and starts using it. In 9 cases of 10 it works, because fonts are loaded in time. To fix 1 case when it doesn't work, it is needed to catch that case, investigate, and find out, that AwtFontManager requires awaiting for the loading.

I believe a good API should enforce it via its shape, not describing it via docs.

skiko/src/awtMain/kotlin/org/jetbrains/skiko/FontInterop.awt.kt
```diff
@@ -0,0 +1,118 @@
...
+object AwtFontManager {
+    private var fontsMap = ConcurrentHashMap<String, File>()
+    @Volatile
+    private var allFontsCachedImpl = false
+
+    private val waitChannel = RendezvousBroadcastChannel<Int>()
+
+    init {
+        GlobalScope.launch(Dispatchers.IO) {
+            cacheSystemFonts()
```
igordmn, line 23: Because of this, we probably can't quickly exit the application (Compose uses `exitProcess`, but we shouldn't rely on it).

So, I suggest to move it out of the API behavior too (see the previous comment).

Or add `yield` after every font load, so coroutine can quickly exit on cancellation.

skiko/src/awtMain/kotlin/org/jetbrains/skiko/FontInterop.awt.kt
olonho: I reworked API.

olonho: > Can we make caching as implementation details, so we can change it in the future?
> 
> Also, we can get rid of `Font` from API and receive only `familyName: String`. And maybe even get rid of AWT from API.
> 
> I suggest this API:
> 
> ```
> class SystemFonts {
>     val familyNames: List<String>    
> 
>     fun fileOf(familyName: String): File
> 
>     companion object {
>        suspend fun load(): SystemFonts
>     }
> }
> ```
> 
> It is no longer global, and requires manual creation/passing, but global operations/API's has its own flaws (we can discuss).

Don't think like this API much.

skiko/src/awtMain/kotlin/org/jetbrains/skiko/FontInterop.awt.kt
olonho: `yield()` idea sounds right, done it.

skiko/src/awtMain/kotlin/org/jetbrains/skiko/FontInterop.awt.kt
olonho: Because it has different purpose, which makes sense to exist in common code. This class is about AWT interop, and fact how we discover fonts (via system dirs) is an implementation detail.

igordmn (approved):

PR #453, opened 2022-01-17
Cleanup samples

olonho:


samples/SkiaAwtSample/src/main/kotlin/SkiaAwtSample/App.kt
```diff
@@ -15,15 +15,15 @@ fun main(args: Array<String>) {
     val windows = parseArgs(args)
     repeat(windows) {
         when (System.getProperty("skiko.interop")) {
```
Rsedaikin, line 17: Looks like it should be changed to `skiko.swing.interop`

PR #455, opened 2022-01-20
Added support to show/hide screen keyboard. 

Rsedaikin:
- Refactored [SkikoUIViewController]
 - Added [SkikoUIView] for use in compose

Note: to show/hide screen keyboard on simulator you must unset option: I/O -> Keyboard -> Connect Hardware Keyboard

### DIFF
--- a/skiko/src/iosMain/kotlin/org/jetbrains/skiko/SkiaLayer.ios.kt
+++ b/skiko/src/iosMain/kotlin/org/jetbrains/skiko/SkiaLayer.ios.kt
@@ -24,6 +24,16 @@ actual open class SkiaLayer {
         return true
     }
 
+    fun showScreenKeyboard() {
+        view?.becomeFirstResponder()
+    }
+
+    fun hideScreenKeyboard() { view?.resignFirstResponder() }
+
+    fun isScreenKeyboardOpen(): Boolean {
+        return if (view == null) false else view!!.isFirstResponder
+    }
+
     actual var renderApi: GraphicsApi
         get() = GraphicsApi.METAL
         set(value) { throw UnsupportedOperationException() }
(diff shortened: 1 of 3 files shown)


skiko/src/iosMain/kotlin/org/jetbrains/skiko/SkiaLayer.ios.kt
```diff
@@ -24,6 +24,16 @@ actual open class SkiaLayer {
         return true
     }
 
+    fun showScreenKeyboard() {
```
olonho, line 27: How it was tested? If just do 
```
window!!.rootViewController = SkikoViewController(
            SkikoUIView(
                SkiaLayer(SkikoGestureEventKind.values()).apply {
                    skikoView = GenericSkikoView(this, makeApp(this))
                    showScreenKeyboard()
                }
            )
        )
```
in sample keyboard is not  shown.

Rsedaikin: We should show/hide keyboard only after view is loaded, so code should be smth like this:
```
override fun application(application: UIApplication, didFinishLaunchingWithOptions: Map<Any?, *>?): Boolean {
        window = UIWindow(frame = UIScreen.mainScreen.bounds)
        var showKeyboard: () -> Unit = {}
        window!!.rootViewController = SkikoViewController(
            SkikoUIView(
                SkiaLayer(SkikoGestureEventKind.values()).apply {
                    skikoView = GenericSkikoView(this, makeApp(this))
                    showKeyboard = { showScreenKeyboard() }
                }
            )
        )
        window!!.makeKeyAndVisible()
        showKeyboard.invoke()
        return true
    }
```

PS. Also make sure the option I/O -> Keyboard -> Connect Hardware Keyboard is unset.

olonho: Yep, works this way, modulo backspace key.

olonho (approved):

PR #457, opened 2022-01-24
Move to Skia m99-51988d0bc1

olonho:


skiko/src/commonMain/kotlin/org/jetbrains/skia/Canvas.kt
```diff
@@ -798,8 +633,8 @@ open class Canvas internal constructor(ptr: NativePointer, managed: Boolean, int
     fun drawTriangleFan(
         positions: Array<Point>,
         colors: IntArray?,
```
VeselovAlex, line 635: `colors: IntArray? = null`?

VeselovAlex, line 635: `colors: IntArray? = null`?

PR #468, opened 2022-02-01
Rework Android renderer

olonho:


skiko/src/androidMain/kotlin/org/jetbrains/skiko/Renderer.android.kt
```diff
@@ -123,8 +135,8 @@ private class SkikoSurfaceRender(private val layer: SkiaLayer) : GLSurfaceView.R
             canvas?.drawPicture(it.instance)
             Unit
         }
-
         context?.flush()
+        manager.onFrameCompleted()
```
igordmn, line 139: Thanks!

This is better, but we still need to figure out how to wait for swapBuffers, otherwise we still render as fast as possible, consuming the battery.

igordmn (approved):

skiko/src/androidMain/kotlin/org/jetbrains/skiko/Renderer.android.kt
olonho: Why do you think that context flush doesn't swap buffers?

PR #470, opened 2022-02-01
Fixed a case where [event.modifiersEx] does not provide info about the pressed mouse button when using touchpad on MacOS 12 (AWT).

Rsedaikin:
issue: Nearly half of all clicks isn't handled when the Tap to click feature is activated on Mac (see the video attached). OS: macOS Monterey 12.0.1 (21A559).
see: https://youtrack.jetbrains.com/issue/COMPOSE-36

- added [pressedButtons] to SkikoPointerEvent to monitoring current pressed state of mouse buttons
- added [button] to SkikoPointerEvent - current button triggering mouse event

### DIFF
--- a/skiko/src/commonMain/kotlin/org/jetbrains/skiko/Events.kt
+++ b/skiko/src/commonMain/kotlin/org/jetbrains/skiko/Events.kt
@@ -267,7 +267,8 @@ data class SkikoPointerEvent(
     val y: Double,
     val deltaX: Double = 0.0,
     val deltaY: Double = 0.0,
-    val buttons: SkikoMouseButtons = SkikoMouseButtons.NONE,
+    val pressedButtons: SkikoMouseButtons = SkikoMouseButtons.NONE,
+    val button: SkikoMouseButtons = SkikoMouseButtons.NONE,
     val modifiers: SkikoInputModifiers = SkikoInputModifiers.EMPTY,
     val kind: SkikoPointerEventKind,
     val timestamp: Long = 0,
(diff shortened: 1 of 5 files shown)


skiko/src/commonMain/kotlin/org/jetbrains/skiko/Events.kt
```diff
@@ -267,19 +267,20 @@ data class SkikoPointerEvent(
...
+    val button: SkikoMouseButtons = SkikoMouseButtons.NONE,
     val modifiers: SkikoInputModifiers = SkikoInputModifiers.EMPTY,
     val kind: SkikoPointerEventKind,
     val timestamp: Long = 0,
     val platform: SkikoPlatformPointerEvent?
 )
 
 val SkikoPointerEvent.isLeftClick: Boolean
-    get() = buttons.has(SkikoMouseButtons.LEFT) && (kind == SkikoPointerEventKind.UP)
+    get() = pressedButtons.has(SkikoMouseButtons.LEFT) && (kind == SkikoPointerEventKind.UP)
```
igordmn, line 279: Probably we need to check `button` here instead of `pressedButtons`

igordmn (approved):

olonho:
JS is broken, please fix

skiko/src/commonMain/kotlin/org/jetbrains/skiko/Events.kt
Rsedaikin: Done.

PR #475, opened 2022-02-04
Fixed case when user get low refresh rate with VRR capable display.

Rsedaikin:
To disable VRR support, we create a transparent `JPanel` with size (1, 1) and add it on top of the `SkiaLayer`. This causes mixed rendering technologies (java2d and OpenGL/DirectX) to work simultaneously in the same window and VRR cannot be enabled.
see:
- https://github.com/JetBrains/compose-jb/issues/1648
- https://youtrack.jetbrains.com/issue/TBX-6507#focus=Comments-27-5482439.0-0

### DIFF
--- a/skiko/src/awtMain/kotlin/org/jetbrains/skiko/SkiaLayer.awt.kt
+++ b/skiko/src/awtMain/kotlin/org/jetbrains/skiko/SkiaLayer.awt.kt
@@ -90,16 +90,6 @@ actual open class SkiaLayer internal constructor(
         add(backedLayer)
         backedLayer.addHierarchyListener {
             if (it.changeFlags and HierarchyEvent.SHOWING_CHANGED.toLong() != 0L) {
-                if (
-                    vrrDisplayLocker == null
-                    && hostOs == OS.Windows
-                    && (renderApi == GraphicsApi.DIRECT3D || renderApi == GraphicsApi.OPENGL)
-                ) {
-                    vrrDisplayLocker = JPanel().apply {
-                        background = Color(0, 0, 0, 0)
-                    }
-                    add(vrrDisplayLocker, Integer.valueOf(0))
-                }
                 checkShowing()
             }
         }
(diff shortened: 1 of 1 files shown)


skiko/src/awtMain/kotlin/org/jetbrains/skiko/SkiaLayer.awt.kt
```diff
@@ -88,6 +90,16 @@ actual open class SkiaLayer internal constructor(
         add(backedLayer)
         backedLayer.addHierarchyListener {
             if (it.changeFlags and HierarchyEvent.SHOWING_CHANGED.toLong() != 0L) {
+                if (
```
igordmn, line 93: Is there a reason why it is in SHOWING_CHANGED handler, and not in `init {}` block? If we can avoid dynamically adding UI elements, and construct all swing UI in the constructor, we simplify the code.

Rsedaikin: Because in `init {}` block we still do not know about actual `renderApi` but `SHOWING_CHANGED` also not the good place for that: replaced it inside `findNextWorkingRenderApi` method.

igordmn (approved):

skiko/src/awtMain/kotlin/org/jetbrains/skiko/SkiaLayer.awt.kt
```diff
@@ -307,6 +316,7 @@ actual open class SkiaLayer internal constructor(
         }
         super.setBounds(x, y, roundedWidth, roundedHeight)
         backedLayer.setSize(roundedWidth, roundedHeight)
+        vrrDisplayLocker?.setBounds(0, 0, 1, 1)
```
igordmn, line 319: We need to call it after
```
vrrDisplayLocker = JPanel().apply {
 background = Color(0, 0, 0, 0)
}
```
too, because `setBounds` can be called before it

igordmn: probably we need to call it only there, and here we can remove `setBounds `

Rsedaikin: Done.

PR #484, opened 2022-02-08
reuse InteropScope instance in k/js when using interopScope {}

eymar:
@olonho what do you think? 

it's for k/js only.
It seems that InteropScope() doesn't need to instantiated every time. Even though it's a small object, we can save a bit considering we use `interopScope {}` quite often.


olonho (approved):
Wow, looks cool!

PR #486, opened 2022-02-09
fix Shaper functions 

eymar:
ensure ManageString instance is not collected by GC before using _ptr in native calls.


skiko/src/commonMain/kotlin/org/jetbrains/skia/shaper/Shaper.kt
```diff
@@ -191,11 +192,12 @@ class Shaper internal constructor(ptr: NativePointer) : Managed(ptr, _FinalizerH
     fun shapeLine(text: String?, font: Font?, opts: ShapingOptions): TextLine {
         return try {
             Stats.onNativeCall()
+            val manageString = ManagedString(text)
```
olonho, line 195: `managedString`

olonho (approved):
Small typo

PR #487, opened 2022-02-10
Allow fallback if surface cannot be created

olonho:


skiko/src/awtMain/kotlin/org/jetbrains/skiko/context/MetalContextHandler.kt
```diff
@@ -41,7 +42,7 @@ internal class MetalContextHandler(layer: SkiaLayer) : JvmContextHandler(layer)
             SurfaceOrigin.TOP_LEFT,
             SurfaceColorFormat.BGRA_8888,
             ColorSpace.sRGB
-        )
+        ) ?: throw RenderException("Cannot create surface")
```
Rsedaikin, line 45: Should we also add this to uikit, android and macos-native?

Rsedaikin (approved):

PR #489, opened 2022-02-11
add docs 1..15

eymar:


skiko/src/macosMain/kotlin/org/jetbrains/skiko/SkiaLayer.macos.kt
```diff
@@ -25,39 +25,75 @@ import platform.Foundation.addObserver
...
+ */
 actual open class SkiaLayer {
     fun isShowing(): Boolean {
         return true
     }
 
+    /**
+     * [GraphicsApi.METAL] is the only GraphicsApi supported for MacOs.
+     */
     actual var renderApi: GraphicsApi = GraphicsApi.METAL
```
olonho, line 40: Guess we'd better make throwing setter if new value != METAL.

eymar: added a setter throwing an IAE if new value != METAL.

PR #497, opened 2022-02-14
Add multiplatform openUri API

olonho:


skiko/src/awtMain/kotlin/org/jetbrains/skiko/Actuals.awt.kt
```diff
@@ -34,3 +39,22 @@ internal actual fun makeDefaultRenderFactory(): RenderFactory {
...
+    } catch (e: java.awt.HeadlessException) {
+        null
+    }
+}
+
+internal actual fun ClipboardManager_setText(text: String) {
+    systemClipboard?.setContents(StringSelection(text), null)
+}
+internal actual fun ClipboardManager_getText(): String? {
+    return systemClipboard?.getData(DataFlavor.stringFlavor) as String?
```
igordmn, line 59: This function throws an exception, if the clipboard contains non-text (Image, for example).

In Compose we have wrapped it into:
```
try {
  it.getData(DataFlavor.stringFlavor) as String
} catch (_: UnsupportedFlavorException) {
  null
}
```

igordmn (approved):

PR #500, opened 2022-02-16
Cursor control API

olonho:


skiko/src/androidMain/kotlin/org/jetbrains/skiko/Actuals.android.kt
```diff
@@ -64,4 +65,28 @@ internal actual fun ClipboardManager_getText(): String? {
     val clipboard = context.getSystemService(Context.CLIPBOARD_SERVICE) as ClipboardManager
     val clip = clipboard.primaryClip ?: return null
     return clip.getItemAt(0)?.text?.toString()
-}
\ No newline at end of file
+}
+
+actual typealias Cursor = android.view.PointerIcon
```
igordmn, line 70: This class is available only in API 24. Any usage of this class will throw an exception in pre-API 24. Every code block that uses this API should be wrapped into:
```
if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.N) {

}
```

Are we sure that is safe to expose this class as actual, and not as implementation details?

igordmn: the safe option is to write a wrapper `AndroidCursor`

olonho: It's 5 years old platform, guess we may safely not support it.

Issue #493, opened 2022-02-14
renderInfo is not accurate


adorokhine:
SkiaLayer.info prints the requested API regardless of which API was chosen. For example, trying DIRECT3D on Linux happily prints:
```
GraphicsApi: DIRECT3D
OS: linux x64
Vendor: Intel
Model: Mesa Intel(R) UHD Graphics (CML GT2)
Total VRAM: 31 MB
```

I have stepped into it with a debugger and confirmed that the chosen API was actually LinuxOpenGLRedrawer.

I think renderInfo should accurately represent what was actually chosen so that meaningful performance comparisons can be made.

skiko-awt-linux-x64 0.7.7.

okushnikov:
Please check the <a href="https://youtrack.jetbrains.com/issue/SKIKO-493">following ticket</a> on YouTrack for follow-ups to this issue. GitHub issues will be closed in the coming weeks.

PR #506, opened 2022-02-22
Throw an exception if the user tries to create a SkiaLayer with an inappropriate [renderApi] on the current OS.

Rsedaikin:
see: https://github.com/JetBrains/skiko/issues/493

### DIFF
--- a/skiko/src/jvmMain/kotlin/org/jetbrains/skiko/SkikoProperties.kt
+++ b/skiko/src/jvmMain/kotlin/org/jetbrains/skiko/SkikoProperties.kt
@@ -62,6 +62,26 @@ internal object SkikoProperties {
     }
 
     fun fallbackRenderApiQueue(initialApi: GraphicsApi) : List<GraphicsApi> {
+        when(initialApi) {
+            GraphicsApi.DIRECT3D -> {
+                when(hostOs) {
+                    OS.JS,
+                    OS.Ios,
+                    OS.Android,
+                    OS.Linux,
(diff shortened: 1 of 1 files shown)


skiko/src/jvmMain/kotlin/org/jetbrains/skiko/SkikoProperties.kt
```diff
@@ -62,6 +62,26 @@ internal object SkikoProperties {
...
+            GraphicsApi.DIRECT3D -> {
+                when(hostOs) {
+                    OS.JS,
+                    OS.Ios,
+                    OS.Android,
+                    OS.Linux,
+                    OS.MacOS -> throw Exception("$hostOs does not support DirectX rendering API.")
+                }
+            }
+            GraphicsApi.METAL -> {
```
igordmn, line 75: macOs also doesn't support SOFTWARE_FAST, and OPENGL

Maybe it is easier to check fallbackApis?
```
val indexOfInitialApi = fallbackApis.indexOf(initialApi)
require(indexOfInitialApi >= 0) {
    "$hostOs does not support $initialApi rendering API."
}
fallbackApis = fallbackApis.drop(indexOfInitialApi + 1)
```

Rsedaikin: Yes, good point. Done.

igordmn (approved):

olonho (approved):

PR #513, opened 2022-03-07
Convert more fun interfaces

Goooler:


olonho (commented):
What's the purpose of this change?

Goooler:
To be more readable and idiomatic, adopt the fun interfaces feature of Kotlin 1.4.

olonho:
Well, we try to keep Skiko pretty basic in used language features (so that Compose compiler plugin can process code calling Skiko properly), and so try to avoid newer language features without extensive testing, and fun interfaces known to be problematic with Compose plugin at the certain moment.

Goooler:
Alright.

PR #515, opened 2022-03-09
Don't depend on kotlinx-coroutines-swing

igordmn:
Because it overrides Dispatchers.Main

Fixes https://github.com/JetBrains/compose-jb/issues/1943 on skiko side

### DIFF
--- a/skiko/src/awtMain/kotlin/org/jetbrains/skiko/HardwareLayer.kt
+++ b/skiko/src/awtMain/kotlin/org/jetbrains/skiko/HardwareLayer.kt
@@ -2,7 +2,6 @@ package org.jetbrains.skiko
 
 import kotlinx.coroutines.*
 import kotlinx.coroutines.channels.Channel
-import kotlinx.coroutines.swing.Swing
 import java.awt.*
 import java.awt.event.FocusEvent
 import java.awt.event.InputMethodEvent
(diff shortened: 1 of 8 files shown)


skiko/src/awtMain/kotlin/org/jetbrains/skiko/HardwareLayer.kt
```diff
@@ -110,7 +109,7 @@ internal open class HardwareLayer(
         // and its accessibility context. This timeout is used to deal with concurrency
         // TODO Find more reliable procedure
         resetFocusAccessibleJob?.cancel()
-        resetFocusAccessibleJob = GlobalScope.launch(Dispatchers.Swing) {
+        resetFocusAccessibleJob = GlobalScope.launch(MainUIDispatcher) {
```
igordmn, line 112: Maybe we rename `MainUIDispatcher` to `Dispatchers.SkikoUI`?

olonho: Well, it's hard question, as `Dispatchers.XXX` assumes that it's part of coroutines library. Guess we'd better keep as is.

olonho (approved):

skiko/src/awtMain/kotlin/org/jetbrains/skiko/HardwareLayer.kt
igordmn: Ok

PR #516, opened 2022-03-10
Uikit multistage keyboard (Japanese, Chinese)

Rsedaikin:


skiko/src/iosMain/kotlin/org/jetbrains/skiko/SkikoViewController.kt
```diff
@@ -112,6 +112,7 @@ class SkikoUIView : UIView, UIKeyInputProtocol, UITextInputProtocol {
                 )
             }
         }
```
olonho, line 114: Maybe null check?

dima-avdeev-jb (approved):

skiko/src/iosMain/kotlin/org/jetbrains/skiko/SkikoViewController.kt
dima-avdeev-jb: add null check

dima-avdeev-jb (approved):

dima-avdeev-jb (changes requested):
Found bugs with last character replacement

PR #524, opened 2022-04-03
Complete bindings for SkRuntimeShaderBuilder::uniform

kirill-grouchnikov:


kirill-grouchnikov:
@igordmn - this is the second phase, completing support for all types in `SkRuntimeShaderBuilder::uniform`

skiko/src/commonMain/kotlin/org/jetbrains/skia/Matrix22.kt
```diff
@@ -0,0 +1,47 @@
...
+        var result = 1
+        result = result * PRIME + mat.contentHashCode()
+        return result
+    }
+
+    override fun toString(): String {
+        return "Matrix22(_mat=$mat)"
+    }
+
+    companion object {
```
igordmn, line 33: Could you add IDENTITY constant for this matrix?

skiko/src/commonMain/kotlin/org/jetbrains/skia/Matrix22.kt
```diff
@@ -0,0 +1,47 @@
...
+import org.jetbrains.skia.impl.withResult
+
+/**
+ * 2x2 matrix.
+ */
+class Matrix22(vararg mat: Float) {
+    /**
+     * Matrix elements are in row-major order.
+     */
+    internal val mat: FloatArray
```
igordmn, line 14: It would be better to make this public. Because otherwise without matrix operations it will be useless. Also, it will be good, If you make `Matrix44.mat` public too.

igordmn: > Because otherwise without matrix operations it will be useless

Ah, sorry. It would be useless, when we get matrix (for example, canvas.matrix). I encountered on this issues in the past. It isn't useless, when we set matrix. But it is better to fix it anyway

igordmn (approved):

skiko/src/commonMain/kotlin/org/jetbrains/skia/Matrix22.kt
kirill-grouchnikov: Added

skiko/src/commonMain/kotlin/org/jetbrains/skia/Matrix22.kt
kirill-grouchnikov: Done for 2x2 and 4x4, so that all three expose this as a public field

PR #526, opened 2022-04-05
Add bindings for SkRuntimeShaderBuilder::child

kirill-grouchnikov:
For `SkShader` and `SkColorFilter` that can be passed as uniforms into the main SkSL shader


kirill-grouchnikov:
@igordmn - this is the last part, bindings for `SkRuntimeShaderBuilder::child`

PR #535, opened 2022-05-16
Add bindings for half leading to TextStyle and StrutStyle

manu-unter:
This will allow us to create vertically centered text with custom line height

We need it in Fleet to get rid of our custom leading that we are currently applying around the actual lines of text. We can't leave it up to the Paragraph API right now because without the half leading feature, the additional space will not be evenly distributed above and below the text, but instead proportionally to the ascent and descent in the font. This leaves the text vertically uncentered, which is especially notable when there is a selection or a highlight behind it.

I already tested the new bindings locally with Fleet.

(cherry picked from commit f6f5338c1239625dd3ddf0e3b0e56b393e8a43ae)


igordmn:
Thanks! 0.7.19 will be built soon

manu-unter:
Thank you!

PR #538, opened 2022-05-19
Add binding for SkGraphics methods

kropp:


skiko/src/commonMain/kotlin/org/jetbrains/skia/Graphics.kt
```diff
@@ -0,0 +1,172 @@
+package org.jetbrains.skia
+
+import org.jetbrains.skia.impl.Library.Companion.staticLoad
+import org.jetbrains.skia.impl.Stats
+
+class Graphics {
```
igordmn, line 6: It seems tests failed because we import
```
import org.jetbrains.skia.*
```
And it conflicts with `java.awt.Graphics`

We need to change imports in `SkiaLayer.awt.kt`

PR #543, opened 2022-05-26
Delegate focus methods from SkiaLayer to backedLayer

igordmn:
Similar to the other methods

(see the comment "We need to delegate all event listeners..." why we do that)

Also, disable changing focus when click on SkiaLayer if we set `setCanReceiveFocus(false)`, because focus should be changed only when we click on some focusable Swing/Compose component

### DIFF
--- a/skiko/src/awtMain/kotlin/org/jetbrains/skiko/SkiaLayer.awt.kt
+++ b/skiko/src/awtMain/kotlin/org/jetbrains/skiko/SkiaLayer.awt.kt
@@ -1,13 +1,12 @@
 package org.jetbrains.skiko
 
 import org.jetbrains.skia.*
+import org.jetbrains.skia.Canvas
 import org.jetbrains.skiko.redrawer.Redrawer
+import java.awt.*
 import java.awt.Color
-import java.awt.Component
-import java.awt.Graphics
 import java.awt.event.*
 import java.awt.im.InputMethodRequests
-import java.awt.Window
 import java.util.concurrent.CancellationException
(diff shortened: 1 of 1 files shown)


skiko/src/awtMain/kotlin/org/jetbrains/skiko/SkiaLayer.awt.kt
```diff
@@ -83,6 +82,19 @@ actual open class SkiaLayer internal constructor(
...
+                if (canReceiveFocus(cause)) {
+                    super.requestFocus(cause)
+                }
+            }
+
+            override fun requestFocusInWindow(cause: FocusEvent.Cause?): Boolean {
+                return canReceiveFocus(cause) && super.requestFocusInWindow(cause)
+            }
+
+            private fun canReceiveFocus(cause: FocusEvent.Cause?) = cause != FocusEvent.Cause.MOUSE_EVENT ||
```
dima-avdeev-jb, line 96: Why only check to MOUSE_EVENT?
How about "Tab" key on keyboard?

igordmn: By `setRequestFocusEnabled` description, it should only affect mouse:
```
Provides a hint as to whether or not this JComponent should get focus. This is only a hint, and it is up to consumers that are requesting focus to honor this property. This is typically honored for mouse operations, but not keyboard operations
```
It is used in Compose to prevent changing focus on ComposePanel clicks, but keep focus changing when we press Tab

PR #545, opened 2022-06-02
set moduleName for kjs target

eymar:
Kotlin 1.7.0 introduced some changes related to kjs module names (affecting modules consuming skiko).

Therefore, we need to override the moduleName to avoid name collision with a different skiko.js file

### DIFF
--- a/skiko/build.gradle.kts
+++ b/skiko/build.gradle.kts
@@ -228,6 +228,7 @@ kotlin {
 
     if (supportWasm) {
         js(IR) {
+            moduleName = "skiko-kjs" // override the name to avoid name collision with a different skiko.js file
             browser {
                 testTask {
                     dependsOn("linkWasm")


igordmn:
Does this mean, that the names of artifacts will be different?

Currently we have these names:
image

What names will be after this PR?

eymar:
@igordmn 

Screenshot 2022-06-02 at 11 58 29


The compiled js file in the build directory. The file names in maven won't be affected (tested).

As you can see on the screenshot, there is 1 skiko.js file (generated by emscripten and bundled into skiko-js-wasm-runtime). And skiko-kjs.js is generated by kotlin-js (from skiko-js klib). 

Before 1.7.0 skiko-kjs.js was named kotlin_org_jetbrains_skiko.js, but since 1.7.0 it was named skiko.js by default. It's an expected change in behaviour according to k/js team. So they suggested to override the module name.

Issue #551, opened 2022-07-08
Traces after we remove SkiaLayer from some swing container


igordmn:
Skiko 0.7.21, macOs

See the video, the tab "dotMemory Profiler" uses SkiaLayer. After we switch to another tab, the SkiaLayer content still on the screen until we resize the whole window.

https://user-images.githubusercontent.com/5963351/178004441-2e00d36a-104b-4b7c-894e-5ef643a3a6db.mov

okushnikov:
Please check the <a href="https://youtrack.jetbrains.com/issue/SKIKO-551">following ticket</a> on YouTrack for follow-ups to this issue. GitHub issues will be closed in the coming weeks.

PR #552, opened 2022-07-08
Fix traces after we remove SkiaLayer from a Swing container

igordmn:
It seems there are some pending transactions, which weren't flushed after we remove the metal layer

Fixes https://github.com/JetBrains/skiko/issues/551


igordmn:
CI is failing in master too. I ran the CI on the same commit in master, which passed on CI in the past.

Haven't figure out how to fix it yet. Will try to investigate it later.

Let's don't block PR merges, unless it hypothetically can break linux (we won't know about that because of failing CI)

PR #559, opened 2022-07-22
Set proper dylib id and version

SergeevPavel:
Now we can see an agent build path on the place of dylib id and version set to 0.0.0


SergeevPavel:
something on the topic:
https://stackoverflow.com/questions/67055770/usage-of-current-version-and-compatibility-version-on-macos
https://stackoverflow.com/questions/69370484/why-do-dynamic-libraries-on-macos-have-an-id

igordmn (approved):

PR #561, opened 2022-07-28
Fix segfaults

SergeevPavel:
Fix couple of memory issues


skiko/src/jvmMain/cpp/common/Font.cc
```diff
@@ -227,7 +227,10 @@ extern "C" JNIEXPORT jshort JNICALL Java_org_jetbrains_skia_FontKt__1nGetUTF32Gl
 extern "C" JNIEXPORT jint JNICALL Java_org_jetbrains_skia_FontKt__1nGetStringGlyphsCount
   (JNIEnv* env, jclass jclass, jlong ptr, jstring str, jint len) {
     SkFont* instance = reinterpret_cast<SkFont*>(static_cast<uintptr_t>(ptr));
-    return instance->countText(skString(env, str).c_str(), len * sizeof(jchar), SkTextEncoding::kUTF16);
```
igordmn, line 230: Segfault was because we pass char array into countText, and then release the string?

If that is the case, probably we need to replace the other code parts where we use such code. I will look into it, when find time.

igordmn (approved):
Thanks!

PR #562, opened 2022-08-04
migrate to m105

SergeevPavel:


igordmn (approved):

skiko/build.gradle.kts
```diff
@@ -145,10 +145,12 @@ fun compileNativeBridgesTask(os: OS, arch: Arch): TaskProvider<CompileSkikoCppTa
                 val iosArchFlags = when (arch) {
                     Arch.Arm64 -> arrayOf(
                         "-target", "arm64-apple-ios",
+                        "-mios-version-min=11.0",
```
igordmn, line 148: Is it added because this is the minimal version supported by Skia?

SergeevPavel: Yes, the new version of skia uses some language features that are allowed only from ios11

SergeevPavel:
Looks like we are ready to merge. @igordmn what do you think? Also can we issue a new version of skiko after that?

igordmn:
Let's merge. Github Actions seems hangs, but the previous run was almost successful, except wasm still weren't published.

igordmn:
0.7.28 is building

PR #567, opened 2022-08-15
Remove Angle support (it is not fully implemented)

igordmn:
Angle is a stable Skia renderer, which is used by Chromium instead of direct DirectX, OpenGl, Metal renderers. It is considered more stable, but needs the ~30mb binary library. Angle mimics OpengGL ES and calls corresponding DirectX/Metal API's.

Angle support wasn't fully implemented, there are just a bunch of class prototypes in the source code, that gets in the way during refactorings.

Deleting it for now, until we decide to fully support it. When we decide it, we can look at git history.


dima-avdeev-jb (approved):

SergeevPavel:
I'm not against this change, but probably we should consider angle support. In Fleet we have quite suspicious rendering glitches on the D3D backend. I hope some of them should be gone after disabling `AtlasPathRenderer ` (https://bugs.chromium.org/p/skia/issues/detail?id=13519). But some of them are still here.

igordmn:
> I'm not against this change, but probably we should consider angle support

It is an option. But it requires investments, and complicates infrastructure (we probable will need a separate binary). It would be great, if we'll fix all the bugs, because Compose is also hurted by these bugs.

Also, embedding 30 mb into Compose probably isn't a option for us. And it will be wierd if we provide two options for Compose - Angle and not Angle, because it isn't obvious why Angle is better, and why we just don't fix the bugs for non-Angle implementation.

If we find a way to significally shrink the binary of Angle it will be win-win, but I don't know if it is possible.

mcpiroman:
Aaand speaking of binary size, if Angle shell be an optional module, I'd also appreciate if other skia [static libraries](https://github.com/JetBrains/skiko/blob/4561adb56b5b575e9b0e607f334f8ccff50ff596/skiko/build.gradle.kts#L570) would became so as well. Because given versatility of compose multiplatform, and even more of the skiko itself, oftentimes the app is bundled with many skia's features it doesn't use (e.g. skottie, webp) to a space sensitive environment (mobile/web/embedded).

I don't know how to achieve this technically though. I'd be best to link the libs dynamically instead of statically and then ship a pair of a DLL and jar file in a separate maven bundles. But if it's not possible then we'd have to link them statically at build time, or be provided with a artifact matrix, which is not so convenient.

igordmn:
> The main value in using angle is that it's battle tested in chrome

It is a good reason, let's discuss it later. I am not against if someone implements full Angle support, but Compose team can't invest resources to implment it right now. The binary size is a big factor againsts investments.

> people expect that text editor should be useful in any actual hardware

This is true for every application. It is possible that we can provide two options - very stable (Angle), but big binary size, and semi-stable, but small binary size. But from app developer perspective it adds confusion.

> I'd also appreciate if other skia [static libraries](https://github.com/JetBrains/skiko/blob/4561adb56b5b575e9b0e607f334f8ccff50ff596/skiko/build.gradle.kts#L570) would became so as well.

It is a good idea, thanks. If we decide in the future to split skiko, we will consider shipping svg/skottie/etc as separate binaries. Something like skiko-skottie-windows-x64 is an option. But if the separate libraries will be significantly greater in size than one single binary, then probably we'll keep the single binary.

igordmn:
Let's merge it, as the Angle code is half baked, and probably don't work. When we decide to implement Angle, it would be better to write it from scratch, looking at code in this PR as an example.

mcpiroman:
OK, so I created an issue #570 for the split.
As to Angle, wonder how does flutter approach that? It should have analogous requirements to Compose.

PR #568, opened 2022-08-16
SkiaLayerAnalytics - send events for different steps of rendering

igordmn:
It will help us to detect GPU's, on which we have crashes. After we compare two events, and if in the second event we have less hits, it means we have a crash. We can investigate it, and if the issue only in GPU, we will add it to not-supported-adapter.list

The plan to integrate this analytics to Toolbox.

### DIFF
--- a/skiko/src/jvmMain/cpp/windows/directXRedrawer.cc
+++ b/skiko/src/jvmMain/cpp/windows/directXRedrawer.cc
@@ -19,18 +19,6 @@
 
 const int BuffersCount = 2;
 
-// This is a list of not supported graphics cards that have rendering issues (black screen, flickering)
-// with the current Swing/Skia integration.
-// If PC has other graphics cards suitable for DirectX12, one of them will be used. Otherwise,
-// rendering will falls back to OpenGL.
-const std::vector<std::wstring> notSupportedAdapters{
-    L"Intel(R) HD Graphics 520",
-    L"Intel(R) HD Graphics 530",
-    L"Intel(R) HD Graphics 4400",
-    L"NVIDIA GeForce GTX 750 Ti",
-    L"NVIDIA GeForce GTX 960M",
-    L"NVIDIA Quadro M2000M"};
-
 class DirectXDevice
 {
 public:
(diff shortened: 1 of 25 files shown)


skiko/src/jvmMain/cpp/windows/directXRedrawer.cc
```diff
@@ -19,18 +19,6 @@
...
-// with the current Swing/Skia integration.
-// If PC has other graphics cards suitable for DirectX12, one of them will be used. Otherwise,
-// rendering will falls back to OpenGL.
-const std::vector<std::wstring> notSupportedAdapters{
-    L"Intel(R) HD Graphics 520",
-    L"Intel(R) HD Graphics 530",
-    L"Intel(R) HD Graphics 4400",
-    L"NVIDIA GeForce GTX 750 Ti",
-    L"NVIDIA GeForce GTX 960M",
-    L"NVIDIA Quadro M2000M"};
```
dima-avdeev-jb, line 32: Are we skip this list of video adapters?
Or it moves to another place?

dima-avdeev-jb (approved):

skiko/src/jvmMain/cpp/windows/directXRedrawer.cc
igordmn: It is [here](https://github.com/JetBrains/skiko/blob/master/skiko/src/jvmMain/resources/not-supported-adapter.list)

dima-avdeev-jb: Also, we have broken build on android

dima-avdeev-jb:
Also, we have broken build on android

dima-avdeev-jb:
Now, we have 2 failed tests on Windows:
org.jetbrains.skiko.SkiaLayerPerformanceTest[awt] > FPS is near display refresh rate (multiple windows)[awt] FAILED
org.jetbrains.skiko.SkiaLayerTest[awt] > analytics[awt] FAILED

dima-avdeev-jb:
Cool, all CI checks work's Fine!

PR #579, opened 2022-08-25
Fix getting text from clipboard

igordmn:
Sometimes, internal AWT implementation can't properly parse the clipboard data.

getData's docs say that it can throw IOException

Fixes https://github.com/JetBrains/compose-jb/issues/2098


eymar (approved):
Does it need to catch IOException if it catches Exception first?

igordmn:
> Does it need to catch IOException if it catches Exception first?

Oh, I accidentally pushed wrong code  😅. It shouldn't be Exception.

eymar (approved):

PR #581, opened 2022-08-29
Expose SkiaLayer.redrawImmediately

igordmn:
Sometimes we need to update the current frame as fast as we can to avoid input lag.

For example, when we typed a character, or scrolled a list. If we render the new state only on the next frame, the user will experience input lag:
```
event -> change state -> needRedraw -> vsync -> draw -> vsync
```
If we would call this:
```
event -> change state -> redrawImmediately -> vsync
```
then there is no input lag.

We can't know in what condition we need to call it. The caller of SkiaLayer should decide it by itself. Usually it should be called when we change some state in response to an input event.

### DIFF
--- a/skiko/src/awtMain/kotlin/org/jetbrains/skiko/SkiaLayer.awt.kt
+++ b/skiko/src/awtMain/kotlin/org/jetbrains/skiko/SkiaLayer.awt.kt
@@ -484,6 +484,17 @@ actual open class SkiaLayer internal constructor(
         redrawer?.needRedraw()
     }
 
+    /**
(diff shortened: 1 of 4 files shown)


igordmn:
Fleet requested this change. No plans to use it in Compose yet.

igordmn:
It seems, that `redrawImmediately` still waits for vsync.

Because vsync block the rendering thread, and when we call `redrawImmediately`, we wait for the rendering thread (or drawLack, if we talk about Metal/Direct3D renderers).

To properly fix the issue, we probably need something like asynchronous non-blocking vsync.

PR #583, opened 2022-08-31
Text input multistage for China and Japan languages

dima-avdeev-jb:
Thanks @Rsedaikin for start this branch!

Add multistage text input for China and Japan lanuages.

Add platform specific SkikoInput interface as expect / actual.
So, we don't need to have all methods for all platfroms. I think, it's simpler to maintain.


skiko/src/commonMain/kotlin/org/jetbrains/skiko/SkikoView.kt
```diff
@@ -6,7 +6,10 @@ interface SkikoView {
     // Input
     fun onKeyboardEvent(event: SkikoKeyboardEvent) = Unit
     fun onPointerEvent(event: SkikoPointerEvent) = Unit
+
+    @Deprecated("This method will be removed. Use override val input:PlatformInput")
     fun onInputEvent(event: SkikoInputEvent) = Unit
+    val input: SkikoInput
```
dima-avdeev-jb, line 12: Breaking changes here

skiko/src/iosMain/kotlin/org/jetbrains/skiko/SkikoUIView.kt
```diff
@@ -117,57 +122,360 @@ class SkikoUIView : UIView, UIKeyInputProtocol {
...
+    /**
+     * The text position for the end of a document.
+     * https://developer.apple.com/documentation/uikit/uitextinput/1614555-endofdocument
+     */
+    override fun endOfDocument(): UITextPosition {
+        return IntermediateTextPosition(skiaLayer?.skikoView?.input?.endOfDocument() ?: 0)
+    }
+
+    /**
+     * Attention! fromPosition and toPosition may be null
```
eymar, line 257: Could the parameters type have `?`  - `UITextPosition?` ?

skiko/src/androidMain/kotlin/org/jetbrains/skiko/PlatformInput.android.kt
```diff
@@ -0,0 +1,5 @@
+package org.jetbrains.skiko
+
+actual interface PlatformInput {
```
eymar, line 3: Is there an actual `interface PlatformInput`? I can't find it in this branch.

skiko/src/iosMain/kotlin/org/jetbrains/skiko/SkikoUIView.kt
dima-avdeev-jb: Unfortunately, if I make the type nullable (UITextPosition?), then the compilation fails with **wrong override**.
May be we can you some annotations tricks to fix this compilation error...

skiko/src/commonMain/kotlin/org/jetbrains/skiko/SkikoView.kt
```diff
@@ -6,7 +6,10 @@ interface SkikoView {
     // Input
     fun onKeyboardEvent(event: SkikoKeyboardEvent) = Unit
     fun onPointerEvent(event: SkikoPointerEvent) = Unit
+
+    @Deprecated("This method will be removed. Use override val input:PlatformInput")
```
eymar, line 10: Is this message outdated? Does it mean `Use override val input: SkikoInput` ?

skiko/src/androidMain/kotlin/org/jetbrains/skiko/PlatformInput.android.kt
eymar: I guess PlatformInput was replaced by SkikoInput?

dima-avdeev-jb: Yeah, you right!
Fixed

skiko/src/commonMain/kotlin/org/jetbrains/skiko/SkikoView.kt
dima-avdeev-jb: I provide this warning to users of skiko. I they update skiko to the next version, they will see this warning.
Maybe it's unnecessary, because skiko still in pre-released versions, and we can break compatibility without warnings?

eymar: Do I understand correctly that SkikoInputEvent is not applicable for ios anymore (it's not used for ios)?

dima-avdeev-jb: Yes, for iOS it's redundant and not used anymore

eymar: Do IPADs support keyboards? Can we connect a keyboard to an IPAD?
If it's possible, does it mean that we need to add more methods to SkikoInput for ios? To handle not only text inputs, but keyboard events such as tab/arrows navigation?

dima-avdeev-jb: Yeah iPad support hardware keyboard. Our compose uikit samples already works with iPadOS.
iPadOS use same .ipa app packages like iOS.
We can add more methods to handle keyboard support right in actual SkikoInput interface.
This interface is supposed to be extensible

dima-avdeev-jb: And actually i want to add more methods in SkikoInput later to handle RTL text layout's

eymar: Got it! thank you :)

skiko/src/iosMain/kotlin/org/jetbrains/skiko/SkikoUIView.kt
```diff
@@ -89,9 +90,11 @@ class SkikoUIView : UIView, UIKeyInputProtocol {
                     uiPress.key?.let {
                         pressedKeycodes.add(it.keyCode)
                     }
-                    skiaLayer?.skikoView?.onKeyboardEvent(
-                        toSkikoKeyboardEvent(press, SkikoKeyboardEventKind.DOWN)
-                    )
+                    if (uiPress.key?.keyCode != SkikoKey.KEY_BACKSPACE.value) {
```
igordmn, line 93: Is it possible to know the source of the key - virtual or physical keyboard? In case of physical keyboard, probably we should send all keys, so users can handle them.

Also, there can be not only text. For example, we can want to go to the previous screen when we press Backspace.

skiko/src/iosMain/kotlin/org/jetbrains/skiko/SkikoUIView.kt
```diff
@@ -117,57 +122,360 @@ class SkikoUIView : UIView, UIKeyInputProtocol {
...
+     * https://developer.apple.com/documentation/uikit/uitextinput/1614558-replace
+     * @param range A range of text in a document.
+     * @param text A string to replace the text in range.
+     */
+    override fun replaceRange(range: UITextRange, withText: String) {
+        skiaLayer?.skikoView?.input?.replaceRange(range.toSkikoTextRange(), withText)
+    }
+
+    override fun setSelectedTextRange(selectedTextRange: UITextRange?) {
+        selectedTextRange?.let {
```
igordmn, line 182: We should use this code:
```
skiaLayer?.skikoView?.input?.setSelectedTextRange(selectedTextRange?.toSkikoTextRange())
```
Otherwise, if iOS sends null, we won't do anything.

igordmn (commented):
Awesome 👍

skiko/src/commonMain/kotlin/org/jetbrains/skiko/SkikoView.kt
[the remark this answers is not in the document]
dima-avdeev-jb: Thanks!

skiko/src/commonMain/kotlin/org/jetbrains/skiko/SkikoView.kt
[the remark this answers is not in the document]
dima-avdeev-jb: Yeah, good decision!

skiko/src/iosMain/kotlin/org/jetbrains/skiko/SkikoUIView.kt
dima-avdeev-jb: Sure, thanks!

skiko/src/iosMain/kotlin/org/jetbrains/skiko/SkikoUIView.kt
dima-avdeev-jb: Only physical keyboard events are received in this function. 
I made such a check to avoid double deletion.

Probably I should remove this check. And check for deletion elsewhere inside compose.

dima-avdeev-jb: this function is only called  when the physical keyboard pressed

dima-avdeev-jb: removed if condition

dima-avdeev-jb: For now it will work with double deletion of 2 last characters, when physical keyboard backspace pressed.
But, I think this bug is not a big problem and may be fixed later.

skiko/src/commonMain/kotlin/org/jetbrains/skiko/SkikoView.kt
dima-avdeev-jb: Backward compatibility for compilation was not changed, because we provide default val input: SkikoInput = get() Empty

skiko/src/iosMain/kotlin/org/jetbrains/skiko/SkikoUIView.kt
igordmn: > I think this bug is not a big problem and may be fixed later

If it is possible to fix this error at low-cost, better to make a fix now (in this PR on in the next one). Because without the fix we have physical keyboard unusable.

Will the previous approach, when we detect the backspace key, store pressed keys, and check it in `deleteBackward` will cause any issues?

PR #587, opened 2022-09-02
Fix crash on Windows when there is no compatible DirectX adapter

igordmn:
Fixes https://github.com/JetBrains/skiko/issues/582

### DIFF
--- a/skiko/src/awtMain/kotlin/org/jetbrains/skiko/redrawer/Direct3DRedrawer.kt
+++ b/skiko/src/awtMain/kotlin/org/jetbrains/skiko/redrawer/Direct3DRedrawer.kt
@@ -23,9 +23,12 @@ internal class Direct3DRedrawer(
 
     init {
         val adapter = chooseAdapter(properties.adapterPriority.ordinal)
-        adapterName = getAdapterName(adapter)
+        if (adapter == 0L) {
+            throw RenderException("Failed to choose DirectX12 adapter.")
+        }
+        adapterName = getAdapterName(0)
         adapterMemorySize = getAdapterMemorySize(adapter)
-        onDeviceChosen(getAdapterName(adapter))
+        onDeviceChosen(adapterName)
         device = createDirectXDevice(adapter, layer.contentHandle, layer.transparency)
         if (device == 0L) {
             throw RenderException("Failed to create DirectX12 device.")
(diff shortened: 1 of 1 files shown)


skiko/src/awtMain/kotlin/org/jetbrains/skiko/redrawer/Direct3DRedrawer.kt
```diff
@@ -23,9 +23,12 @@ internal class Direct3DRedrawer(
...
     init {
         val adapter = chooseAdapter(properties.adapterPriority.ordinal)
-        adapterName = getAdapterName(adapter)
+        if (adapter == 0L) {
+            throw RenderException("Failed to choose DirectX12 adapter.")
+        }
+        adapterName = getAdapterName(0)
         adapterMemorySize = getAdapterMemorySize(adapter)
-        onDeviceChosen(getAdapterName(adapter))
+        onDeviceChosen(adapterName)
```
dima-avdeev-jb, line 31: And adapterName = getAdapterName(0) will always take first adapter.
But adapterMemorySize = getAdapterMemorySize(adapter) takes from another one adapter.

I don't understand how it can fix bug?

igordmn: Yeah, I accidentally pushed the test code. it is fixed in the last commit.

`getAdapterName(0)` will crash, because 0 is not an index, it is pointer

igordmn: The fix in throwing exception. When we throw RenderException, fallback to OpenGL will happen

skiko/src/awtMain/kotlin/org/jetbrains/skiko/redrawer/Direct3DRedrawer.kt
```diff
@@ -23,9 +23,12 @@ internal class Direct3DRedrawer(
 
     init {
         val adapter = chooseAdapter(properties.adapterPriority.ordinal)
+        if (adapter == 0L) {
+            throw RenderException("Failed to choose DirectX12 adapter.")
+        }
         adapterName = getAdapterName(adapter)
         adapterMemorySize = getAdapterMemorySize(adapter)
-        onDeviceChosen(getAdapterName(adapter))
+        onDeviceChosen(adapterName)
```
dima-avdeev-jb, line 31: So, we protected from crash in getAdapterName(0)
But now we have another one crash:
```
if (adapter == 0L) {
            throw RenderException("Failed to choose DirectX12 adapter.")
        }
```

with same condition adapter == 0

Or exception RenderException will handled in another place?

igordmn: Yes, the logic catching RenderException is inside SkiaLayer. When we catch it, we fallback to the next render api.

Windows has this chain of API's: DirectX -> OpenGL -> Software Fast (via native code) -> Software Compat (via Swing)

dima-avdeev-jb (approved):

PR #589, opened 2022-09-03
Fix skiko sample on iosSimulatorArm64

dima-avdeev-jb:
Also now, launching from Xcode works on Apple M1 CPU

### DIFF
--- a/samples/SkiaMultiplatformSample/build.gradle.kts
+++ b/samples/SkiaMultiplatformSample/build.gradle.kts
@@ -1,3 +1,6 @@
+import org.jetbrains.kotlin.gradle.dsl.KotlinNativeBinaryContainer
+import org.jetbrains.kotlin.gradle.plugin.mpp.KotlinNativeTarget
+
 buildscript {
     repositories {
         mavenLocal()
@@ -18,7 +21,6 @@ plugins {
 }
 
 val coroutinesVersion = "1.5.2"
-val LAUNCH_ON_REAL_DEVICE = false // true, if you want to launch on real device from Xcode
 
 repositories {
     mavenLocal()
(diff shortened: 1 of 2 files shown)


samples/SkiaMultiplatformSample/build.gradle.kts
```diff
@@ -163,51 +167,22 @@ kotlin {
...
+            getByName(iosSourceSetName).dependsOn(iosMain)
+            if (LAUNCH_ON_REAL_DEVICE) {
+                getByName("iosArm64").dependsOn(iosMain)
             }
         }
     }
 }
 
-if (hostOs == "macos") {
-    project.tasks.register<Exec>("runIosSim") {
```
eymar, line 180: What would be the way to run it on simulator after this task is removed? Only via xcode or q
Appcode?

dima-avdeev-jb:
Sorry, I was in a hurry to open PullRequest. 
There are still problems with the X64 ios simulator.

dima-avdeev-jb:
Now works good from AppCode and Xcode on Intel and M1 CPU's.
Also ./gradlew runNative works on Intel and M1 from now.

samples/SkiaMultiplatformSample/build.gradle.kts
```diff
@@ -61,30 +64,20 @@ val unzipTask = tasks.register("unzipWasm", Copy::class) {
...
 
 kotlin {
-    val targets = mutableListOf<org.jetbrains.kotlin.gradle.plugin.mpp.KotlinNativeTarget>()
-
-    if (hostOs == "macos") {
-        val nativeHostTarget = when (host) {
-            "macos-x64" -> macosX64()
-            "macos-arm64" -> macosArm64()
-            else -> throw GradleException("Host OS is not supported yet")
+    if (hostOs == "macos") { //todo maybe redundant check
```
eymar, line 67: I think it's indeed redundant. And having it disabled on other OS might even be not desirable, because changing some common code API can affect disabled macOs source sets and IDE won't complain about it.

For example in compose project we have all targets declared unconditionally and the build seems to be working on mac os and windows.

eymar (approved):

PR #591, opened 2022-09-09
Fix text input in samples

dima-avdeev-jb:
After changing input, some samples became broken. (iOS and JS)
Fixed it!

Also, found and fixed NPE bug in iOS text input.

### DIFF
--- a/skiko/src/iosMain/kotlin/org/jetbrains/skiko/SkikoUIView.kt
+++ b/skiko/src/iosMain/kotlin/org/jetbrains/skiko/SkikoUIView.kt
@@ -254,7 +254,7 @@ class SkikoUIView : UIView, UIKeyInputProtocol, UITextInputProtocol, UITextPaste
      * Attention! position may be null
      */
     override fun positionFromPosition(position: UITextPosition, offset: NSInteger): UITextPosition? {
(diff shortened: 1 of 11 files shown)


skiko/src/iosMain/kotlin/org/jetbrains/skiko/SkikoUIView.kt
```diff
@@ -250,8 +250,11 @@ class SkikoUIView : UIView, UIKeyInputProtocol, UITextInputProtocol, UITextPaste
         )
     }
 
+    /**
+     * Attention! position may be null
+     */
     override fun positionFromPosition(position: UITextPosition, offset: NSInteger): UITextPosition? {
-        val p = (position as IntermediateTextPosition).position
+        val p = (position as? IntermediateTextPosition)?.position ?: 0
```
igordmn, line 257: shouldn't we return null if position isn't IntermediateTextPosition?

dima-avdeev-jb: Yeah, good idea!

dima-avdeev-jb: Done!

igordmn (approved):

PR #594, opened 2022-09-10
Add binding for SkSurface::draw() that takes sampling parameter

mcpiroman:
Question: why is SamplingMode mode being packed as two ints instead of one long?


mcpiroman:
OK, I see the message "Long can't be used because Long is an object in kotlin/js. Consider using _packedInt1 and _packedInt2"

PR #597, opened 2022-09-21
allow to change font rendering settings in paragraph

SergeevPavel:


skiko/src/commonMain/kotlin/org/jetbrains/skia/paragraph/FontRasterSettings.kt
```diff
@@ -0,0 +1,6 @@
+package org.jetbrains.skia.paragraph
+
+import org.jetbrains.skia.FontEdging
+import org.jetbrains.skia.FontHinting
+
+data class FontRasterSettings(val edging: FontEdging, val hinting: FontHinting, val subpixel: Boolean)
```
igordmn, line 6: It should be called `FontRastrSettings` to align with the Skia name

igordmn, line 6: Could you add a smoke test to `ParagraphStyleTests`? Just set and get the values.

Issue #601, opened 2022-10-12
Skiko failed to render a frame on Windows Direct3D Intel(R) HD Graphics 4600


SergeevPavel:
https://youtrack.jetbrains.com/issue/FL-14786/Blank-Window-on-User-Agreement

SergeevPavel:
related compose issue: https://github.com/JetBrains/compose-jb/issues/1771
As it pointed there, some Intel videocards doesn't support Direct3D 12.
I see two options here, we could try to check Direct3D 12 support programmatically, or just put the videocards into black list.
@igordmn what do you think?

igordmn:
> just put the videocards into black list

If we get the list from the official source (the intel site), it is fine as a fast fix, but it would be better to fix this issue the right way. 

As far as I remember, the check for Direct3D happens [here](https://github.com/JetBrains/skiko/blob/master/skiko/src/awtMain/cpp/windows/directXRedrawer.cc#L264). It seems it ether is not full or right, or we have bugs in drivers (in this case only the blacklist will help).

SergeevPavel:
should be fixed after https://github.com/JetBrains/skiko/pull/603

mahozad:
I have Intel HD Graphics 4600 and the issue resolved for me in v1.2.1.

The app prints `Failed to create DirectX12 device.` at std out at startup.  
Could you please log the fallback render API as well? I think it should probably be logged somewhere [here](https://github.com/JetBrains/skiko/blob/5616a1b18824a3aca6df7af17c5a409ad5bdfcde/skiko/src/awtMain/kotlin/org/jetbrains/skiko/SkiaLayer.awt.kt#L278).

Also, I think these issues are related:
  - https://github.com/JetBrains/compose-jb/issues/610#issuecomment-829075382
  - https://github.com/JetBrains/compose-jb/issues/449
  - https://github.com/JetBrains/compose-jb/issues/1771
  - https://github.com/JetBrains/skiko/issues/601

okushnikov:
Please check the <a href="https://youtrack.jetbrains.com/issue/SKIKO-601">following ticket</a> on YouTrack for follow-ups to this issue. GitHub issues will be closed in the coming weeks.

PR #603, opened 2022-10-14
limit supported D3D_FEATURE_LEVEL only to D3D12 (#601)

SergeevPavel:


igordmn:
Thanks!

Makes sense, as Skia only supports DirectX 12 (according to [this answer](https://groups.google.com/g/skia-discuss/c/TGBzojV7Ygs/m/dfeeb-1QCQAJ))

PR #613, opened 2022-11-04
allow to configure pixel geometry

SergeevPavel:


skiko/src/awtMain/kotlin/org/jetbrains/skiko/SkiaLayer.awt.kt
```diff
@@ -53,6 +54,7 @@ actual open class SkiaLayer internal constructor(
         isVsyncEnabled: Boolean = SkikoProperties.vsyncEnabled,
         isVsyncFramelimitFallbackEnabled: Boolean = SkikoProperties.vsyncFramelimitFallbackEnabled,
         renderApi: GraphicsApi = SkikoProperties.renderApi,
+        pixelGeometry: PixelGeometry = PixelGeometry.UNKNOWN,
```
igordmn, line 57: Let's move `pixelGeometry` below `analytics` to preserve backward compatibility. Technically we still don't need to do that, but it is better to start.

skiko/src/awtMain/kotlin/org/jetbrains/skiko/SkiaLayer.awt.kt
```diff
@@ -20,7 +20,8 @@ actual open class SkiaLayer internal constructor(
     externalAccessibleFactory: ((Component) -> Accessible)? = null,
     private val properties: SkiaLayerProperties,
     private val renderFactory: RenderFactory = RenderFactory.Default,
-    private val analytics: SkiaLayerAnalytics = SkiaLayerAnalytics.Empty,
+    actual val pixelGeometry: PixelGeometry = PixelGeometry.UNKNOWN,
```
igordmn, line 23: As far as I understand, this can be changed in real time, and in the future we need to somehow react on these changes.

For example, when we use a tablet, that can happen often.

SergeevPavel: Yep, I think that easiest way is to re-initialize a layer with different settings

PR #620, opened 2022-11-23
update info about dpi event if we came from `frameDispatcher`

SergeevPavel:


SergeevPavel:
issue in Fleet https://youtrack.jetbrains.com/issue/FL-14185/Fleet-window-gets-blury

SergeevPavel:
@igordmn what do you think?

skiko/src/awtMain/kotlin/org/jetbrains/skiko/SkiaLayer.awt.kt
```diff
@@ -494,6 +489,11 @@ actual open class SkiaLayer internal constructor(
         check(isEventDispatchThread()) { "Method should be called from AWT event dispatch thread" }
         check(!isDisposed) { "SkiaLayer is disposed" }
 
+        if (backedLayer.checkContentScale()) {
```
igordmn, line 492: Could you try to move this check not to `update`, but to 2 `paint`'s:
```
        backedLayer = object : HardwareLayer(externalAccessibleFactory) {
            override fun paint(g: java.awt.Graphics) {
                ...
                redrawer?.needRedraw()
                checkContent()
            }

 override fun paint(g: java.awt.Graphics) {
        super.paint(g)
        checkContent()
        ...
  }

  fun checkContent() {
        if (backedLayer.checkContentScale()) {
            notifyChange(PropertyKind.ContentScale)
        }
        redrawer?.syncSize()
  }
```

If it works, better to do it this way.

SergeevPavel: Yes, I have only precedent of absence of `SkiaLayer.paint` on scale change. And it's pretty unstable in terms of reproducibility. I brifly searched through the JBR codebase and it's not obvious is it guarantied or not to perform `paint` on every scale change.

Anyway, as far it's hard to reproduce and debug, I will follow your suggestion, add `checkContent()` in both of `paint` calls and add some more logs in Fleet. Maybe it will help to debug this case in fututre.

What about replacing this hack with tracking the change DPI events in OS, it sounds like a realy right solution, but I'm not sure how easy would be to integrate it in JBR.

igordmn: > What about replacing this hack with tracking the change DPI events in OS

We can write any native code we need (we can access to `windowHandle` for each platform), but it usually isn't easy, and we should do it for each platform separately.

> how easy would be to integrate it in JBR.

I think, if we implement reading/observing DPI via native code, it won't conflict with any functionality in JDK/JBR.

igordmn: >  is it guarantied or not to perform paint on every scale change.

If we use pure Swing, by design, it must call `paint` on every density change. Otherwise Swing content won't be valid, So it will be considered as a bug of Swing.

But here we don't use pure Swing, we mix it with AWT component `java.awt.Canvas`, which doesn't work well with Swing. Because of that we apply different hacks (override 2 paint's)

igordmn: >  it must call paint on every density change

Or maybe I mistaken 🤔, because `paint` is also is part of rendering pipeline, and there can be similar optimizations - like not calling `paint` if the window is minimized.

SergeevPavel: > like not calling paint if the window is minimized
in Fleet case it's somehow connected with waking laptop from a sleep. But after all we handle a plenty of user event a draw frames with incorrect scale and probably witout any `paint` call

igordmn (approved):

PR #622, opened 2022-11-25
Add keyboard customisation (iOS)

cl3m:
Add keyboard customisation variables on iOS to be used in compose-jb


cl3m:
The upcoming usage on compose-jb is in this commit : https://github.com/JetBrains/androidx/commit/4244f2d978447c4e9ebf8a0d93b10ade352fb51f

dima-avdeev-jb (approved):
LGTM
Code contains public mutable variables. (For now, I don't know how to do it better.)
But it's iOS only code. So, I decided to merge it as is.
In the future, we may change it.

PR #623, opened 2022-11-27
simplify sample usage

dima-avdeev-jb:
Users need an easy way to launch our sample (https://stackoverflow.com/questions/74393070/is-skiko-right-now-only-available-for-jvm-awt).
 - Set a fixed skiko version in sample.
 - Add options to use composite build in several samples.
 - Describe how to run samples with different dependencies.
 - Fix run configurations.
 
After this PR, users may run samples as is, without additional steps.
In the future, we should update a skiko version in samples/*/gradle.properties.


DEVELOPMENT.md
```diff
@@ -43,6 +44,78 @@ For example, if we want to include UI tests when we test JVM target, call this:
...
+./gradlew :SkiaAwtSample:run -Dskiko.transparency=true -Pskiko.version=0.0.0-SNAPSHOT
+```
+ - skiko.swing.interop=true
+```bash
+./gradlew :skiko:publishAllSkikoJvmRuntimeToMavenLocal &&\
+./gradlew :SkiaAwtSample:run -Dskiko.swing.interop=true -Pskiko.version=0.0.0-SNAPSHOT
+```
+
+### samples/SkiaJsSample with mavenLocal
+If page in browser will be empty, try to reload it.
```
igordmn, line 81:
```suggestion
If the page in the browser is empty, try to reload it.
```

Anyway, its seems like a bug, and better to not mention it in Readme? Because if the page is empty, reloading it usually is the first thing to try.

DEVELOPMENT.md
```diff
@@ -43,6 +44,78 @@ For example, if we want to include UI tests when we test JVM target, call this:
...
+```bash
+./gradlew publishToMavenLocal -p skiko -Pskiko.native.enabled=true -Pskiko.wasm.enabled=true &&\
+./gradlew jsBrowserDevelopmentRun -p samples/SkiaJsSample -Pskiko.version=0.0.0-SNAPSHOT
+```
+
+### samples/SkiaJsSample with skikoCompositeBuild=true
+It means: includeBuild("skiko") with dependency substitution. It compiles faster.
```
igordmn, line 88:
```suggestion
When you add `-DskikoCompositeBuild=true`, Skiko from [sources](https://github.com/JetBrains/skiko/tree/master/skiko) is used (without this property the maven artifact is used). Every time you change sources, they are automatically applied when you run `jsBrowserDevelopmentRun`.
```

`includeBuild` usually doesn't tell what this property does, even if people are familiar with the Compose Gradle build feature. Also, it is used not to build project faster, but to use skiko from sources, not as prepublished maven artifact.

It also means, that we don't need to call `publishSkikoWasmRuntimePublicationToMavenLocal` and add  `-Pskiko.version=0.0.0-SNAPSHOT`.

But see my other suggestion, if we do that way (use Composite build by default), we don't need this section at all.

samples/SkiaMultiplatformSample/settings.gradle.kts
```diff
@@ -8,7 +8,7 @@ pluginManagement {
 }
 rootProject.name = "SkiaMultiplatformSample"
 
-if (System.getenv("SKIKO_COMPOSITE_BUILD") == "1") {
+if (System.getenv("SKIKO_COMPOSITE_BUILD") == "1" || System.getProperty("skikoCompositeBuild") == "true") {
```
igordmn, line 11: Let's add this section to SkiaAwtSample too.

samples/SkiaJsSample/settings.gradle.kts
```diff
@@ -10,3 +10,10 @@ pluginManagement {
 }
 rootProject.name = "SkiaJsSample"
 
+if (System.getenv("SKIKO_COMPOSITE_BUILD") == "1" || System.getProperty("skikoCompositeBuild") == "true") {
```
igordmn, line 13:
```suggestion
// This section is needed to use Skiko from sources instead as a maven artifact. If you use Skiko in your own project outside of this repository, you don't this section.
if (File("../../skiko").exists()) {
```
I assume, when we develop Skiko - there is no point to use Skiko from maven repo. And if users just copy the example, it will work too.

This change will simplify the whole process:
- we can just run the example with a simple Gradle command without additional properties (`./gradlew run`). It will work as in our use case, and in user's use case (when they copy the example)
- we can open the example in IDEA without adding a global SKIKO_COMPOSITE_BUILD (I just checked the AWT sample, highlighting and debugging work)

If we do this change, we can (and should) simplify the Readme and the CI file - there will be no need to call `publishAllSkikoJvmRuntimeToMavenLocal` and define `skiko.version`.

Also, because of this, we can remove `skiko.version` property completely, and just use `0.7.41` constant instead of it.

(we can discuss this online, to sync on the final decision)

settings.gradle.kts
```diff
@@ -1,3 +1,7 @@
 rootProject.name = "skiko-all"
```
igordmn, line 1: This file (the root `setting.gradle.kts` was added in the old times, when we had only the AWT implementation, and needed a way to open the project in IDEA as whole. Let's remove it to avoid confusion. Now we have a better way (opening individual examples)

But after removing, we should change the CI file (call gradlew not from the root folder, but from the `skiko` folder.

DEVELOPMENT.md
dima-avdeev-jb: We still need to call `publishSkikoWasmRuntimePublicationToMavenLocal`.
It's a different jar, packaged by tricky script.

DEVELOPMENT.md
dima-avdeev-jb: Done

dima-avdeev-jb: Yeah, I think we can skip this comment. It's obviously.

samples/SkiaJsSample/settings.gradle.kts
dima-avdeev-jb: It's a good suggestion. But we have problems with this case.

`includeBuild("skiko")` provides only `org.jetbrains.skiko:skiko:1.2.3` artifact.

When we run task `publishAllSkikoJvmRuntimeToMavenLocal`, under the hood this task downloads and packs skia inside jar and saves to mavenLocal several additional dependencies `org.jetbrains.skiko:skiko-awt-runtime-macos-arm64:1.2.3`, `org.jetbrains.skiko:skiko-awt-runtime-macos-x64:0.0.0`, `org.jetbrains.skiko:skiko-android-runtime-arm64:1.2.3` and others.

Also, we have task `publishSkikoWasmRuntimePublicationToMavenLocal` to save `org.jetbrains.skiko:skiko-js-wasm-runtime:1.2.3`

I think we also can solve this problem, but I don't know how...
For this case, we need to rewrite publishing tasks and also provide modules instead of jars: https://github.com/JetBrains/skiko/blob/90fb5988e064d583daa920791418c3b7a69a154b/skiko/build.gradle.kts#L1232

DEVELOPMENT.md
```diff
@@ -1,6 +1,7 @@
 ## Building JVM bindings
 
 * Set JAVA_HOME to location of JDK, at least version 11
+* If you have **Apple silicon CPU (M1)**, better to use **X86** JDK architecture to compile JS target.
```
eymar, line 4: Doesn't it work otherwise? What happens otherwise?

dima-avdeev-jb: For me, on Arm64 JDK, sometimes I have error with yarn or webpack. (Like failed to install yarn)
I am also trying to use `kotlin.js.webpack.major.version=4`

With change JDK to X86 JDK, all works fine.
And for now, also Arm64 JDK works fine. Maybe it's a cache problem. I decided to mention JDK change as one of the possible solutions.

eymar: I see. Thanks!

dima-avdeev-jb: I changed a little bit this sentence to better understanding.

eymar (approved):

PR #627, opened 2022-12-01
Fix skiko declaration bridges

igoriakovlev:
Kotlin/C bridges synchronisation:

Removed Kotlin parameters (like kotlin.String) from skia bridges (would not work for js)
Fixed invalid int_64t result return from bridge (would not work for emscripten)
Added/removed missing arguments from bridges
Fixed invalid named bridge

### DIFF
--- a/skiko/src/jvmMain/cpp/common/Bitmap.cc
+++ b/skiko/src/jvmMain/cpp/common/Bitmap.cc
@@ -161,11 +161,18 @@ extern "C" JNIEXPORT jlong JNICALL Java_org_jetbrains_skia_BitmapKt__1nGetPixelR
     return reinterpret_cast<jlong>(pixelRef);
 }
 
-extern "C" JNIEXPORT jlong JNICALL Java_org_jetbrains_skia_BitmapKt__1nGetPixelRefOrigin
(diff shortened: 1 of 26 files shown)


skiko/src/jvmMain/cpp/common/Bitmap.cc
```diff
@@ -161,11 +161,18 @@ extern "C" JNIEXPORT jlong JNICALL Java_org_jetbrains_skia_BitmapKt__1nGetPixelR
     return reinterpret_cast<jlong>(pixelRef);
 }
 
-extern "C" JNIEXPORT jlong JNICALL Java_org_jetbrains_skia_BitmapKt__1nGetPixelRefOrigin
+extern "C" JNIEXPORT jint JNICALL Java_org_jetbrains_skia_BitmapKt__1nGetPixelRefOriginX
```
eymar, line 164: > Fixed invalid int_64t result return from bridge (would not work for emscripten)

Is that comment about this change?

eymar (approved):
Looks good to me. Thank you!
Let's wait for @igordmn to have a look

skiko/src/jvmMain/cpp/common/Bitmap.cc
igoriakovlev: Yes, it is.

igordmn (approved):

PR #628, opened 2022-12-08
Linux AWT test - fail fast

dima-avdeev-jb:
- split a big bash script into separate scripts
 - add timeout-minutes: 25 to :skiko:awtTest
 
We decided to not include bash script to print jstack.
But I want to save it in this PR for the history (or future usage):
```bash
          printJstackInBackground() {
            # Temp function to print jstack stacktrace every 60 seconds
            while [ 0 ]
              do
                sleep 60
                echo "--------------------------------------------------------------------------"
                echo "--------------------------jstack result:----------------------------------"
                echo "ps aux | grep java, result:"
                ps aux | grep java
                ps aux | grep java | grep jdk -i | awk '{print $2}' | while read pid; do
                   jstack $pid
                done
                echo "--------------------------------------------------------------------------"
              done
          }
          # & - means start background job:
          printJstackInBackground &

```

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -103,24 +103,26 @@ jobs:
       - run: |
           # Test Linux AWT
           printJstackInBackground() {
+            #TODO Remove when the issue https://github.com/JetBrains/skiko/issues/629 is fixed
             # Temp function to print jstack stacktrace every 60 seconds
             # Pay attention to org.jetbrains.skiko.redrawer.LinuxSoftwareRedrawer.dispose(LinuxSoftwareRedrawer.kt:27)
             while [ 0 ]
               do
+                sleep 60
                 echo "--------------------------------------------------------------------------"
+                echo "--------------------------jstack result:----------------------------------"
                 echo "ps aux | grep java, result:"
                 ps aux | grep java
                 ps aux | grep java | grep jdk -i | awk '{print $2}' | while read pid; do
                    jstack $pid
                 done
                 echo "--------------------------------------------------------------------------"
-                sleep 60
               done
           }
           printJstackInBackground &
           export DISPLAY=:0
           ./gradlew --no-daemon --stacktrace --info -Pskiko.native.enabled=true -Pkotlin.native.cacheKind.linuxX64=none -Pskiko.test.onci=true :skiko:awtTest
-        timeout-minutes: 15
+        timeout-minutes: 25
       - env:
           ANDROID_NDK_HOME: ${{ steps.setup-ndk.outputs.ndk-path }}
         run: |
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -33,6 +33,7 @@ jobs:
           ./gradlew --stacktrace --info -Pskiko.native.enabled=true :skiko:publishToMavenLocal
           ./gradlew --stacktrace --info :SkiaAwtSample:installDist # check AWT sample works
       - uses: actions/upload-artifact@v2
+        if: always()
         with:
           name: test-reports-macos
           path: ./skiko/build/reports/tests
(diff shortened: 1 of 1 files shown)


SergeevPavel:
I'm not realy sure, that it's related. But thing that I notice is that `window` here closed only on hapy test path https://github.com/JetBrains/skiko/blob/c08102bbf9742c920c4ce6e90bffc09c5a202772/skiko/src/awtTest/kotlin/org/jetbrains/skiko/SkiaLayerTest.kt#L362
I checked on macOS and for me it doesn't hang event if test finishes with time out.
But Linux may be a different story

dima-avdeev-jb:
> But thing that I notice is that `window` here closed only on hapy test path

I checked stacktrace with jstack, and found one function stuck in every failed test case:
```
org.jetbrains.skiko.redrawer.LinuxSoftwareRedrawer.dispose(LinuxSoftwareRedrawer.kt:27)
```
In successful cases, this fuction is not stuck.

dima-avdeev-jb:
@SergeevPavel 
We will try to fix the root problem too in the future.
For now, we decided to make fail cases go faster.

.github/workflows/ci.yml
```diff
@@ -94,12 +96,40 @@ jobs:
...
+                   jstack $pid
+                done
+                echo "--------------------------------------------------------------------------"
+                sleep 60
+              done
+          }
+          printJstackInBackground &
+          export DISPLAY=:0
+          ./gradlew --no-daemon --stacktrace --info -Pskiko.native.enabled=true -Pkotlin.native.cacheKind.linuxX64=none -Pskiko.test.onci=true :skiko:awtTest
+        timeout-minutes: 15
```
igordmn, line 123: 15 minutes is not enough, even if tests run as expected. See the previous runs. They are near 15 minutes.

dima-avdeev-jb: It's timeout only for :skiko:awtTest.
In the previous try, it was just 5 minutes.

.github/workflows/ci.yml
```diff
@@ -94,12 +96,40 @@ jobs:
...
-          ./gradlew --stacktrace --info -Pskiko.native.enabled=true -Pkotlin.native.cacheKind.linuxX64=none -Pskiko.test.onci=true :skiko:linuxX64Test :skiko:awtTest
-          ./gradlew --stacktrace --info -Pkotlin.native.cacheKind.linuxX64=none :skiko:publishToMavenLocal
-          ./gradlew --stacktrace --info :SkiaAwtSample:installDist # check jvm sample works
-          ./gradlew -Pskiko.android.enabled=true :skiko:publishSkikoJvmRuntimeAndroidX64PublicationToMavenLocal :skiko:publishSkikoJvmRuntimeAndroidArm64PublicationToMavenLocal :skiko:publishAndroidPublicationToMavenLocal
+          ./gradlew --no-daemon --stacktrace --info -Pskiko.native.enabled=true -Pkotlin.native.cacheKind.linuxX64=none -Pskiko.test.onci=true :skiko:linuxX64Test
+      - run: |
+          # Test Linux AWT
+          printJstackInBackground() {
+            # Temp function to print jstack stacktrace every 60 seconds
+            # Pay attention to org.jetbrains.skiko.redrawer.LinuxSoftwareRedrawer.dispose(LinuxSoftwareRedrawer.kt:27)
```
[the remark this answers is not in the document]
dima-avdeev-jb: loop is infinite.
`printJstackInBackground &`, where & meas background job.
Background Linux jobs do not block execution. When ./gradlew command completes this job will be over too.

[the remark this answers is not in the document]
igordmn: Let's add a comment then:
```
# Remove when the issue https://github.com/JetBrains/skiko/issues/629 is fixed 
```

.github/workflows/ci.yml
igordmn: 15 is still small, even if it takes 5 minutes. Let's set 25

igordmn (approved):

PR #630, opened 2022-12-08
align `nMakeFromBackendRenderTarget` signature with kotlin counterpart

SergeevPavel:


skiko/src/nativeJsMain/cpp/Surface.cc
```diff
@@ -75,22 +75,22 @@ SKIKO_EXPORT KNativePointer org_jetbrains_skia_Surface__1nMakeRasterN32Premul
...
+    // std::unique_ptr<SkSurfaceProps> surfaceProps = skija::SurfaceProps::toSkSurfaceProps(surfacePropsInts);
 
     sk_sp<SkSurface> surface = SkSurface::MakeFromBackendRenderTarget(
         static_cast<GrRecordingContext*>(context),
         *backendRenderTarget,
         grSurfaceOrigin,
         skColorType,
         colorSpace,
-        // "TODO: we silently ignore the surfacePropsObj arg for now.
+        // "TODO: we silently ignore the surfaceProps arg for now.
```
igoriakovlev, line 93: Do we still have to ignore it? I am not sure about the reason.

skiko/src/awtMain/cpp/windows/directXRedrawer.cc
```diff
@@ -354,7 +354,7 @@ extern "C"
     }
```
igoriakovlev, line 354: Are we allow such single word commit message as "rename"?

SergeevPavel: It will be squashed on merge anyway. And this commit is actually rename

skiko/src/nativeJsMain/cpp/Surface.cc
SergeevPavel: Don't know anything about it, let's try to enable it

igordmn (approved):

PR #633, opened 2022-12-13
Fix shaper segfault

SergeevPavel:


SergeevPavel:
As far as we get issues like https://youtrack.jetbrains.com/issue/FL-16855/Fleet-crashed-natively-on-skiko-ShapeLine
it's still possible for GC to collect a Native object before a JNI call, but after pointer is retrived.
I revisited all usages of `getPtr` function and check that when we pass the pointer to JNI call there is some reachability check for the Native object.

It's not all, there are many cases when we use _ptr directly, but I left it out of scope of this change.

SergeevPavel:
don't know why workflow isn't started. I run it manually https://github.com/JetBrains/skiko/actions/runs/3687869375

skiko/src/commonMain/kotlin/org/jetbrains/skia/shaper/TextBlobBuilderRunHandler.kt
```diff
@@ -73,6 +73,7 @@ class TextBlobBuilderRunHandler<T> internal constructor(
     }
 
     init {
+        // TODO [pavel.sergeev] what `manageText` == false is actually mean?
```
igordmn, line 76: I can only guess, but probably it was implemented before String has RefCnt. And now we can use acquire/release instead somehow (but I am not sure)

igordmn (approved):

PR #644, opened 2023-01-06
use kotlin 1.8.0

eymar:


eymar:
NOT GOING TO MERGE IT IN `master` now:

Just checked it in Compose project (mpp/demo) - darwin targets don't work:
```
e: Could not find "org.jetbrains.kotlin.native.platform.CoreFoundationBase"

...
Execution failed for task ':compose:ui:ui-graphics:compileKotlinMacosX64'
```

The error implies that skiko built with kotlin 1.8.0 is not compatible with ui-graphics (kotlin 1.7.10).
I was able to run SkiaMppSample (using koltin 1.7.10 + skiko on kotlin 1.8.0) +. I was able to run our falling-balls-mpp compose demo with compsoe 1.2.2 and forced skiko version. I'm going to find the source of the incompatibility.

eymar:
@igordmn I think it's ready for review now. The new skiko build will be supported only with kotlin 1.8.0 projects.

igordmn (approved):

PR #655, opened 2023-01-29
Fix GitHub Actions Linux mavenLocal test with Android step

dima-avdeev-jb:
CI was failed on platform name "android-33-ext4". Because of "33-ext4" fails to cast toInt(). Now it is fixed.


AlexeyTsvetkov (approved):
`name.removePrefix("android-").removeSuffix("-ext4").toInt()` would look simpler to me, but I don't mind the proposed version too

PR #657, opened 2023-02-01
UIKit Clipboard, simplify permission

dima-avdeev-jb:
UIKit require permission to get clipboard content.
But, we can access method `generalPasteboard.hasStrings()` to get clipboard content. without permission.
It will help in this PR: https://github.com/JetBrains/androidx/pull/393

### DIFF
--- a/skiko/src/iosMain/kotlin/org/jetbrains/skiko/Actuals.ios.kt
+++ b/skiko/src/iosMain/kotlin/org/jetbrains/skiko/Actuals.ios.kt
@@ -15,6 +15,8 @@ internal actual fun ClipboardManager_getText(): String? {
     return UIPasteboard.generalPasteboard.string
 }
 
+internal actual fun ClipboardManager_hasText(): Boolean = UIPasteboard.generalPasteboard.hasStrings()
+
 // TODO: not sure if correct.
 actual typealias Cursor = Any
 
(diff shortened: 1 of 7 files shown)


skiko/src/iosMain/kotlin/org/jetbrains/skiko/Actuals.ios.kt
```diff
@@ -15,6 +15,8 @@ internal actual fun ClipboardManager_getText(): String? {
     return UIPasteboard.generalPasteboard.string
 }
 
+internal actual fun ClipboardManager_hasText(): Boolean = UIPasteboard.generalPasteboard.hasStrings()
```
igordmn, line 18: Can we just return null in `open fun getText(): String?` if this is true?

igordmn: > But, we can access method generalPasteboard.hasStrings() to get clipboard content. without permission.

Also, wouldn't it more right, if we be checking for permission itself in Compose? Having `hasText` and returning null can mislead users, and hide the error.

igordmn: > if we be checking for permission itself in Compose

it can be provided via Skiko though, as you did. It seems we just need to rename the function to something like `hasPermissions`

igordmn (approved):

skiko/src/iosMain/kotlin/org/jetbrains/skiko/Actuals.ios.kt
igordmn: decided that reading text is overkill and not performant if we need to just check if the clipboard is empty

PR #661, opened 2023-02-07
Make device scaling reactivity more reliable and fix layout glitch

manu-unter:
…on Windows

First off, I don't think there was any need for the repeated checks for device pixel ratio changes via `checkContentScale` because we should be able to rely on the `graphicsContextScaleTransform` property from AWT.

I removed the corresponding code and changed `contentScale` to directly return the platform-specific implementations from the toolkit/device. There's also no need to cache the value because the device and the graphics configuration already does it for us.

Secondly, there was a layout glitch on Windows when moving windows between monitors with different device pixel ratios. A fix has to happen directly in the JBR, but for now I created a semi-cheap workaround. See this issue and the linked ones for details: https://youtrack.jetbrains.com/issue/JBR-5259/Canvas-is-painted-at-the-wrong-position-after-dragging-JFrame-to-a-monitor-with-different-scale

### DIFF
--- a/skiko/src/awtMain/kotlin/org/jetbrains/skiko/SkiaLayer.awt.kt
+++ b/skiko/src/awtMain/kotlin/org/jetbrains/skiko/SkiaLayer.awt.kt
@@ -2,6 +2,8 @@ package org.jetbrains.skiko
 
 import org.jetbrains.skia.*
 import org.jetbrains.skia.Canvas
+import org.jetbrains.skiko.hostOs
+import org.jetbrains.skiko.OS
 import org.jetbrains.skiko.redrawer.Redrawer
 import java.awt.Color
 import java.awt.Component
@@ -83,7 +85,6 @@ actual open class SkiaLayer internal constructor(
                 //    For example, on macOs when we resize window or change DPI
                 //
                 // 3. to avoid double paint in one single frame, use needRedraw instead of redrawImmediately
-                this@SkiaLayer.checkContentScale()
                 redrawer?.needRedraw()
             }
 
@@ -109,9 +110,23 @@ actual open class SkiaLayer internal constructor(
         }
         @Suppress("LeakingThis")
         add(backedLayer)
+
         backedLayer.addHierarchyListener {
             if (it.changeFlags and HierarchyEvent.SHOWING_CHANGED.toLong() != 0L) {
                 checkShowing()
+            }
+        }
+
+        addPropertyChangeListener("graphicsContextScaleTransform") {
+            redrawer?.syncSize()
+            notifyChange(PropertyKind.ContentScale)
+
+            // Workaround for JBR-5259
+            if (hostOs == OS.Windows) {
+                SwingUtilities.invokeLater {
+                    backedLayer.setLocation(1, 0)
+                    backedLayer.setLocation(0, 0)
+                }
             }
         }
     }
@@ -129,8 +144,6 @@ actual open class SkiaLayer internal constructor(
         super.addNotify()
         val window = SwingUtilities.getRoot(this) as Window
         window.addComponentListener(fullscreenAdapter)
-        backedLayer.defineContentScale()
-        checkContentScale()
         checkShowing()
         init(isInited)
     }
@@ -158,7 +171,7 @@ actual open class SkiaLayer internal constructor(
     }
 
     actual val contentScale: Float
-        get() = backedLayer.contentScale
+        get() = graphicsConfiguration.defaultTransform.scaleX.toFloat()
 
     /**
      * Returns the pointer to an OS specific handle (native resource) of the [SkiaLayer].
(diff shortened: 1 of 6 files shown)


skiko/src/awtMain/kotlin/org/jetbrains/skiko/SkiaLayer.awt.kt
```diff
@@ -109,11 +108,25 @@ actual open class SkiaLayer internal constructor(
...
             }
         }
+
+        addPropertyChangeListener("graphicsContextScaleTransform") {
+            redrawer?.syncSize()
+            notifyChange(PropertyKind.ContentScale)
+
+            // Workaround for JBR-5259
+            if (hostOs == OS.Windows) {
+                SwingUtilities.invokeLater {
```
igordmn, line 124: I don't have this issue on JBR 17.0.4 - Windows 11, 2 monitors 1.5 + 1.0 scales. Tried with the reproducer in the issue.

Could you check, if we have a bug on JDK as well?

igordmn (commented):
Thanks! This looks better.

skiko/src/awtMain/kotlin/org/jetbrains/skiko/SkiaLayer.awt.kt
```diff
@@ -340,21 +351,24 @@ actual open class SkiaLayer internal constructor(
...
-        var roundedWidth = width
-        var roundedHeight = height
-        if (isInited) {
-            roundedWidth = roundSize(width)
-            roundedHeight = roundSize(height)
-        }
-        super.setBounds(x, y, roundedWidth, roundedHeight)
-        backedLayer.setSize(roundedWidth, roundedHeight)
+    override fun doLayout() {
+        val (roundedWidth, roundedHeight) = if (isInited) roundSize(width) to roundSize(height) else width to height
```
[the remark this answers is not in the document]
manu-unter: Oh right, this was probably because the `contentScale` used to only be correct after `addNotify` was called. But that's not the case anymore. I'll remove the condition!

skiko/src/awtMain/kotlin/org/jetbrains/skiko/SkiaLayer.awt.kt
manu-unter: Will do tomorrow!

igordmn: > One thing that always works though is to change the scale on a single monitor 

Thanks! I reproduced it on JBR/JRE with the SkiaAwtSample.

But can't reproduce it in a Compose application. Investigating...

If we will keep this hack, we need to implement it properly - cancelling a scheduled resize (to avoid hypothetical memory leaks, crashes, and concurrent changes):
```
private val fixContentJob: Job?

fixContentJob?.cancel()
fixContentJob = GlobalScope.launch(MainUIDispatcher) {
   ...
}

fun dispose() {
   ...
   fixContentJob?.cancel()
}
```

P.S. ignore CI crash on iOS

igordmn: > But can't reproduce it in a Compose application. Investigating...

I was wrongly reproduced it. Now it reproduces everywhere in Compose.

It seems like a regression after some Windows update. It worked in Windows 10 and may be in some Windows 11.

It seems, the hack doesn't work when we change the scale from 200 to 125, and almost works when we change from 125 to 200 (we have 10-20px stripes)

manu-unter: I can in fact even reproduce the issue with any scale different from 100% in my VM. It looks to me like there is a third bug in the JBR which for some reason causes the backing buffer or the peer to be reallocated at 100% scale always when the scaling factor changes. Note that this is definitely different behavior from when the window is dragged between monitors, and I would consider this case a lot less frequent. As always, resizing the window helps, but this is not OK.

I have pushed the fix for the job scheduling, and I will continue the investigation about the 1x buffer tomorrow, when I have a proper Windows machine at hand which can compile custom JBRs etc.

manu-unter: I believe I found a way to work around the remaining issue. I filed an additional [JBR ticket](https://youtrack.jetbrains.com/issue/JBR-5305/Canvas-is-placed-in-the-wrong-location-and-cut-off-and-no-property-change-is-fired-after-changing-the-monitor-scale-on-Windows) to track it. It's similar but not the same as the one with two monitors.

I ended up bringing back very similar `checkContentScale` logic to the one that we had before, but this time it's marked with the relevant ticket numbers so we can remove the workarounds once it's fixed at the root. Even with the workarounds in place, proper fixes at the root would be very desirable to avoid flickers and intermediate invalid states.

@igordmn let me know if this works for you and if we can mere it then!

igordmn: Thanks! I will look at it tomorrow.

> but this time it's marked with the relevant ticket numbers so we can remove the workarounds once it's fixed at the root

We can do that only if it is fixed in OpenJDK. Skiko (and Compose) supports not only JBR.

igordmn: I still have this issue (glitches after we change the scale)

image

JDK 15, JBR 17

Does it work on your machine with the Clocks example?

manu-unter: I just tested it with both JBR 17.0.4.1 and corretto 11.0.18, and it worked with both. Are you sure you were using JBR 17? The screenshot lists a version of 11.0.11+9-b1475.1, which seems to be from May 21? https://github.com/JetBrains/JetBrainsRuntime/tree/jb11_0_11-b1475

skiko/src/awtMain/kotlin/org/jetbrains/skiko/SkiaLayer.awt.kt
```diff
@@ -71,19 +74,22 @@ actual open class SkiaLayer internal constructor(
     val canvas: java.awt.Canvas
         get() = backedLayer
 
+    var peerBufferSizeFixJob: Job? = null
```
igordmn, line 77:
```suggestion
    private var peerBufferSizeFixJob: Job? = null
```

igordmn (approved):

skiko/src/awtMain/kotlin/org/jetbrains/skiko/SkiaLayer.awt.kt
igordmn: On a new Windows configuration I can't reproduce it anymore. Maybe it is not always reproducible, maybe I was testing it wrongly.

igordmn (approved):

PR #662, opened 2023-02-07
Fix crash on pack after dispose

SergeevPavel:


SergeevPavel:
* FIxed crash https://youtrack.jetbrains.com/issue/FL-17739/Fleet-crashed-natively
* Replaced `println`s with simple `Logger` class just to mark places where we are logging something
* FPS test is unrelated with fixes, but still useful I think

skiko/src/awtMain/kotlin/org/jetbrains/skiko/SkiaLayer.awt.kt
```diff
@@ -291,7 +291,7 @@ actual open class SkiaLayer internal constructor(
                 redrawer = renderFactory.createRedrawer(this, renderApi, analytics, properties)
                 redrawer?.syncSize()
             } catch (e: RenderException) {
-                println(e.message)
+                Logger.warn("Fallback to next API: ${e.message}")
```
igordmn, line 294: This can happen often, actually. When we run an app on a VM, don't have drivers installed, and when we enable transparency on Windows (it supported only on OpenGL right now).

Let's change it to `info`

skiko/src/awtTest/kotlin/org/jetbrains/skiko/SkiaLayerPerformanceTest.kt
```diff
@@ -146,4 +199,30 @@ class SkiaLayerPerformanceTest {
             helpers.forEach { it.window.dispose() }
         }
     }
+
+    @Ignore("Doesn't pass with SOFTWARE API on macOS")
+    @Test
+    fun `FPS is near display refresh rate (multiple windows with clocks)`() = uiTest {
```
igordmn, line 205: This test throws AssertionError and it will be very flaky with the clocks content.

I suggest transform `printInfo` to `printInfo(assert = true)` and `printInfo(assert = false)`

skiko/src/commonMain/kotlin/org/jetbrains/skiko/Logging.kt
```diff
@@ -0,0 +1,15 @@
+package org.jetbrains.skiko
+
+object Logger {
```
igordmn, line 3: Can we make it internal?

skiko/src/awtMain/kotlin/org/jetbrains/skiko/SkiaLayer.awt.kt
igordmn: > when we enable transparency on Windows (it supported only on OpenGL right now).

On the other hand we should switch to OpenGL straight away, without a fallback. In 2 other cases, it is probably should be `warn`.

skiko/src/awtTest/kotlin/org/jetbrains/skiko/SkiaLayerPerformanceTest.kt
SergeevPavel: Good idea. I use this test locally to measure how changes affect rendering performance. What do you think, if we disable assertions, should we unignore it?

igordmn: yes, better to unignore

igordmn (approved):

skiko/src/awtMain/kotlin/org/jetbrains/skiko/SkiaLayer.awt.kt
SergeevPavel: It's not a big difference for now, because we have not implemented log filtering yet. But still, for me this message is quite important and unexpected, so I prefer to leave it as `warn` for now.

PR #666, opened 2023-02-22
prevent usage of dangling pointer to `MetalDevice`

SergeevPavel:
MetalRedrawer may be disposed during update which leads to use after free

### DIFF
--- a/skiko/src/awtMain/kotlin/org/jetbrains/skiko/redrawer/MetalRedrawer.kt
+++ b/skiko/src/awtMain/kotlin/org/jetbrains/skiko/redrawer/MetalRedrawer.kt
@@ -25,7 +25,11 @@ internal class MetalRedrawer(
     }
     private var drawLock = Any()
 
-    private val device: Long
+    private var device: Long
+        get() {
+            check(field != 0L) { "Device is not initialized" }
+            return field
+        }
     val adapterName: String
     val adapterMemorySize: Long
 
@@ -60,6 +64,7 @@ internal class MetalRedrawer(
         frameDispatcher.cancel()
         contextHandler.dispose()
         disposeDevice(device)
+        device = 0
         super.dispose()
     }
 
@@ -74,7 +79,9 @@ internal class MetalRedrawer(
             setVSyncEnabled(device, enabled = false)
             update(System.nanoTime())
             performDraw()
-            setVSyncEnabled(device, properties.isVsyncEnabled)
+            if (!isDisposed) {
+                setVSyncEnabled(device, properties.isVsyncEnabled)
+            }
         }
     }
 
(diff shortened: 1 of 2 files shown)


skiko/src/awtMain/kotlin/org/jetbrains/skiko/redrawer/MetalRedrawer.kt
```diff
@@ -74,7 +79,9 @@ internal class MetalRedrawer(
             setVSyncEnabled(device, enabled = false)
             update(System.nanoTime())
             performDraw()
-            setVSyncEnabled(device, properties.isVsyncEnabled)
+            if (!isDisposed) {
```
igordmn, line 82: If it will be easy, could you write a test in SkiaLayerTest where we dispose the layer in draw?

skiko/src/awtMain/kotlin/org/jetbrains/skiko/redrawer/MetalRedrawer.kt
```diff
@@ -74,7 +79,9 @@ internal class MetalRedrawer(
             setVSyncEnabled(device, enabled = false)
             update(System.nanoTime())
             performDraw()
```
igordmn, line 81: Shouldn't we check it here? `update` calls `render`, not `performDraw`

igordmn (commented):
Thanks!

skiko/src/awtMain/kotlin/org/jetbrains/skiko/redrawer/MetalRedrawer.kt
SergeevPavel: Not sure that I got what you mean. `perfromDraw` also contains the `isDisposed` check inside. So the only problem is  `setVSyncEnabled`

SergeevPavel:
related issue in Fleet https://youtrack.jetbrains.com/issue/FL-17255/Fleet-crashed-on-dragging-tab-out

skiko/src/awtMain/kotlin/org/jetbrains/skiko/redrawer/MetalRedrawer.kt
igordmn: But how it possible to layer be disposed? `performDraw` don't call user code.

If it is still disposed in `update`, then for safety reasons it is better to put check above, as reading the current code we can think that we added a check for a case when we dispose in `performDraw`, and people can miss this check when they add new code before `performDraw`

igordmn (approved):

skiko/src/awtMain/kotlin/org/jetbrains/skiko/redrawer/MetalRedrawer.kt
SergeevPavel: Ok, right, let's move it above

eymar:
merging it to build 0.7.52

PR #668, opened 2023-02-24
UIKit interop

dima-avdeev-jb:
- Vsync analogue for iOS with CADisplayLink
 - `opaque = false` to draw transparent holes


skiko/src/iosMain/kotlin/org/jetbrains/skiko/redrawer/MetalRedrawer.ios.kt
```diff
@@ -1,46 +1,64 @@
...
-        if (layer.isShowing()) {
-            draw()
+    private inline fun addFrameSubscription() {
+        if (!activeFrameSubscription) {
+            caDisplayLink.addToRunLoop(NSRunLoop.mainRunLoop, NSRunLoop.mainRunLoop.currentMode)
+        }
+    }
+
+    private inline fun removeFrameSubscription() {
+        if (activeFrameSubscription) {
```
MatkovIvan, line 60: It looks like it's always `false` (no assignments to `activeFrameSubscription` at all)

skiko/src/iosMain/kotlin/org/jetbrains/skiko/redrawer/MetalRedrawer.ios.kt
```diff
@@ -1,46 +1,64 @@
...
 import platform.QuartzCore.kCAGravityTopLeft
-import kotlin.system.getTimeNanos
-import platform.CoreGraphics.CGSizeMake
+import platform.darwin.NSObject
 
 internal class MetalRedrawer(
     private val layer: SkiaLayer
 ) : Redrawer {
-    private val contextHandler = MetalContextHandler(layer)
+    internal val contextHandler = MetalContextHandler(layer)
```
MatkovIvan, line 27: Is it required to change?

skiko/src/iosMain/kotlin/org/jetbrains/skiko/redrawer/MetalRedrawer.ios.kt
dima-avdeev-jb: Thanks, fixed!

skiko/src/iosMain/kotlin/org/jetbrains/skiko/redrawer/MetalRedrawer.ios.kt
dima-avdeev-jb: Thanks!
Yeah, it's not needed in this PR.
Reverted

skiko/src/iosMain/kotlin/org/jetbrains/skiko/redrawer/MetalRedrawer.ios.kt
```diff
@@ -83,14 +103,15 @@ internal class MetalRedrawer(
 
     private fun draw() {
         // TODO: maybe make flush async as in JVM version.
-        autoreleasepool {
+        autoreleasepool { //todo maybe autoreleasepool is redundant, make measurements and talk with Kotlin Native team
             if (!isDisposed) {
                 contextHandler.draw()
             }
         }
```
dima-avdeev-jb, line 110: Need advice here

MatkovIvan: One of the reasons why `autoreleasepool` is still needed nowadays is
> If you write a loop that creates many temporary objects

But it only applies to Objective-C objects that marked as "autorelease". However since it's hard to track what exactly will be allocated (ideally re-render the frame should have zero allocations if nothing changed), it's safe to keep it around drawing code.

dima-avdeev-jb: Yeah, thanks! I think so too.
Also, I asked this question to the Kotlin Native team.

dima-avdeev-jb: And inside `contextHandler.draw()` we have a lot of allocations and possible ObjectiveC interop.

dima-avdeev-jb: The Kotlin Native team advised to leave it as is. But measure performance.

skiko/src/iosMain/kotlin/org/jetbrains/skiko/redrawer/MetalRedrawer.ios.kt
```diff
@@ -33,14 +34,33 @@ internal class MetalRedrawer(
     private val queue = device.newCommandQueue()!!
     private var currentDrawable: CAMetalDrawableProtocol? = null
     private val metalLayer = MetalLayer()
+    private var activeFrameSubscription: AtomicBoolean = atomic(false)
```
eymar, line 37: I'm wondering if we need atomic.
we have `isDisposed` and it's just a boolean var.
But I'm not sure if we have some guarantees about the calling thread(s). 

Maybe we can add the display link in `init` block only once and then manage it by setting true/false to `isPaused` https://developer.apple.com/documentation/quartzcore/cadisplaylink/1621229-ispaused - it's thread-safe too (maybe it's cheaper than AtomicBoolean from kotlin)

Then  in `dispose` we will need to call `invalidate()` https://developer.apple.com/documentation/quartzcore/cadisplaylink/1621293-invalidate

eymar (approved):

skiko/src/iosMain/kotlin/org/jetbrains/skiko/redrawer/MetalRedrawer.ios.kt
```diff
@@ -83,14 +103,15 @@ internal class MetalRedrawer(
...
-        autoreleasepool {
+        autoreleasepool { //todo measure performance without autoreleasepool
             if (!isDisposed) {
                 contextHandler.draw()
             }
         }
     }
 
     fun finishFrame() {
+        removeFrameSubscription()
```
igordmn, line 114: Shouldn't we call this (or isPaused = false) in the end of `onDisplayLinkTick` for robustness? I.e. we be sure that every performed frame will be always the only one. In some conditions `finishFrame` can be not called (if `!layer.isShowing()`, for example)

skiko/src/iosMain/kotlin/org/jetbrains/skiko/redrawer/MetalRedrawer.ios.kt
igordmn: If we can implement `isPaused` approach, it looks better.

igordmn (approved):

skiko/src/iosMain/kotlin/org/jetbrains/skiko/redrawer/MetalRedrawer.ios.kt
dima-avdeev-jb: Done

skiko/src/iosMain/kotlin/org/jetbrains/skiko/redrawer/MetalRedrawer.ios.kt
dima-avdeev-jb: Good suggestion!
Done.

skiko/src/iosMain/kotlin/org/jetbrains/skiko/redrawer/MetalRedrawer.ios.kt
```diff
@@ -33,15 +32,21 @@ internal class MetalRedrawer(
...
-    init {
-        metalLayer.init(this.layer, contextHandler, device)
-    }
-
-    private val frameDispatcher = FrameDispatcher(SkikoDispatchers.Main) {
+    private val frameListener: NSObject = FrameTickListener {
         if (layer.isShowing()) {
             draw()
         }
+        caDisplayLink.setPaused(true)
```
igordmn, line 39: Actually, after looking at this code, will it work with animations?

I see this chain of calls now:

```
draw
  ...
    composeScene.render
      ...
      change state
          need redraw
          caDisplayLink.setPaused(true)

caDisplayLink.setPaused(false)
```
That leads to drawing of only the first frame. It may work with animations like CircularProgressIIndicator (they may dispatch new frames in a coroutine). But may not work with the case, when we change some state in `SideEffect`

igordmn: > when we change some state in SideEffect

probably no in a side effect, but in this case:
```
@Composable
fun Test() {
  val state by remember { mutableStateOf(false) }

  if (state) {
    println("rerendered")
  }
  Canvas(Modifier.size(100.dp)) {
     state = true
  }
}
```

dima-avdeev-jb: CircularProgressIndicator works good:


https://user-images.githubusercontent.com/99798741/222785701-9b84d812-b829-4716-a4af-d8b5fb9ac3e0.mov

dima-avdeev-jb: `@Composable Test()` function prints log only once:
`rerendered`

igordmn: It works how expected. But how does it work, is the chain I described above is wrong🤔. `caDisplayLink.setPaused(false)` can be called during `draw`

dima-avdeev-jb: > caDisplayLink.setPaused(false) can be called during draw

Actually, it's a good comment! We should expect this case. 
A made new commit: move setPause(true) before draw() call. This will allow usage like you described!

igordmn: Looks good!

PR #670, opened 2023-03-01
Mention more prerequisites in DEVELOPMENT.md

MatkovIvan:


igordmn (commented):
Thanks!

It is required now, but in the future we should support just `publishAwtToMavenLocal` that shouldn't require all that.

Also, it is required because of this line:
```
internal val Project.isInIdea: Boolean
    get() {
        return System.getProperty("idea.active")?.toBoolean() == true
    }
```
It was introduced to include all sources into the idea project, but it messes with requirements for building from IDE.

igordmn (approved):

PR #674, opened 2023-03-14
set to device layer the same background color as the window have

SergeevPavel:
We expect that this change will fix [FL-12113](https://youtrack.jetbrains.com/issue/FL-12113) White background shown on window resize

### DIFF
--- a/skiko/src/awtMain/objectiveC/macos/MetalRedrawer.mm
+++ b/skiko/src/awtMain/objectiveC/macos/MetalRedrawer.mm
@@ -216,14 +216,16 @@ JNIEXPORT jlong JNICALL Java_org_jetbrains_skiko_redrawer_MetalRedrawer_createMe
         device.layer.pixelFormat = MTLPixelFormatBGRA8Unorm;
         device.layer.contentsGravity = kCAGravityTopLeft;
 
-        CGFloat transparent[] = { 0.0f, 0.0f, 0.0f, 0.0f };
-        device.layer.backgroundColor = CGColorCreate(CGColorSpaceCreateDeviceRGB(), transparent);
-        device.layer.opaque = NO;
+        NSWindow* window = (__bridge NSWindow*) (void *) windowPtr;
 
+        device.layer.opaque = NO;
         if (transparency)
         {
-            NSWindow* window = (__bridge NSWindow*) (void *) windowPtr;
+            CGFloat transparent[] = { 0.0f, 0.0f, 0.0f, 0.0f };
+            device.layer.backgroundColor = CGColorCreate(CGColorSpaceCreateDeviceRGB(), transparent);
             window.hasShadow = NO;
+        } else {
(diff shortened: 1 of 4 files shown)


skiko/src/awtMain/objectiveC/macos/MetalRedrawer.mm
```diff
@@ -216,14 +216,16 @@ JNIEXPORT jlong JNICALL Java_org_jetbrains_skiko_redrawer_MetalRedrawer_createMe
...
 
+        device.layer.opaque = NO;
         if (transparency)
         {
-            NSWindow* window = (__bridge NSWindow*) (void *) windowPtr;
+            CGFloat transparent[] = { 0.0f, 0.0f, 0.0f, 0.0f };
+            device.layer.backgroundColor = CGColorCreate(CGColorSpaceCreateDeviceRGB(), transparent);
             window.hasShadow = NO;
+        } else {
+            device.layer.backgroundColor = window.backgroundColor.CGColor;
```
igordmn, line 228: Compose uses SkiaLayer not only in the whole window, but also in individual parts (like `ComposePanel` that can be embedded into IDEA).

This will not allow to use transparent `ComposePanel`'s.

The fact that MetalRedrawer was coupled with `window` for transparency was unfortunate, and we should fix that in the future.

Can we use `SkiaLayer`'s color instead?

SergeevPavel: Yes, we probably can, but this change would be not so trivial. So I would prefer to postpone it until that decoupling refactoring.

igordmn: > until that decoupling refactoring.

It wasn't planned, and we shouldn't wait for it to implement this feature. In anyway, we will need to read a background of SkiaLayer, and pass it to Redrawer, and then to the native code

igordmn: The PR will look like [this](https://github.com/JetBrains/skiko/pull/677/files). Except, we will read and set background

skiko/src/awtMain/objectiveC/macos/MetalRedrawer.mm
```diff
@@ -268,6 +290,18 @@ JNIEXPORT void JNICALL Java_org_jetbrains_skiko_redrawer_MetalRedrawer_setLayerV
...
 
+JNIEXPORT void JNICALL Java_org_jetbrains_skiko_redrawer_MetalRedrawer_setLayerBackground(
+    JNIEnv *env, jobject redrawer, jlong devicePtr, jint color)
+{
+    @autoreleasepool {
+        MetalDevice *device = (__bridge MetalDevice *) (void *) devicePtr;
+        if (!device || !device.layer) {
+            return;
+        }
+        device.layer.backgroundColor = decodeColor(color);
```
igordmn, line 301: We should check `transparency` here as well

PR #683, opened 2023-03-26
Add configurable logger, solidify workaround for JBR-5274

SergeevPavel:


manu-unter (approved):
LGTM!

skiko/src/awtMain/kotlin/org/jetbrains/skiko/SkiaLayer.awt.kt
```diff
@@ -524,6 +530,8 @@ actual open class SkiaLayer internal constructor(
         check(isEventDispatchThread()) { "Method should be called from AWT event dispatch thread" }
         check(!isDisposed) { "SkiaLayer is disposed" }
 
+        checkContentScale()
```
igordmn, line 533: `update` will be only called in response on `needRedraw` and `redrawImmediately`. So, we will have the wrong density until something is changed. For example, until we hover a button.

igordmn (approved):
Thanks!

Logger looks good to me.

Regarding contentScale - it helps us to not miss the callback completely, but in case when we don't have animations/changes, we still have an issue. See my comment.

skiko/src/awtMain/kotlin/org/jetbrains/skiko/SkiaLayer.awt.kt
SergeevPavel: I know, but it's still better than having the wrong density until a window resize. The root cause of the problem is JBR-5274, and we are going to remove this workaround when it's fixed.

igordmn: ok, let's merge

PR #684, opened 2023-03-26
lateinit var _skiaLayer

dima-avdeev-jb:
Simplify usage of skiaLayer, because it's always not null.


skiko/src/iosMain/kotlin/org/jetbrains/skiko/SkikoUIView.kt
```diff
@@ -19,12 +19,16 @@ import kotlin.math.min
 class SkikoUIView : UIView, UIKeyInputProtocol, UITextInputProtocol,
     org.jetbrains.skiko.objc.UIViewExtensionProtocol {
     @OverrideInit
-    constructor(frame: CValue<CGRect>) : super(frame)
+    constructor(frame: CValue<CGRect>) {
+        error("use SkikoUIView constructor with skiaLayer")
```
MatkovIvan, line 23: Can we add some annotation to throw an error in compile time?

dima-avdeev-jb: Added Deprecated annotation with error level. It will fail if accidentally used on compilation.

MatkovIvan (approved):

dima-avdeev-jb:
Found some problems. Let's postpone it.

PR #685, opened 2023-04-04
fix iOS pointer event timestamp

dima-avdeev-jb:
Fix little mistake


igordmn (approved):

MatkovIvan (commented):
Right. It make sense if `timestamp` is double (not `Int` as I assumed on previous PR)

PR #686, opened 2023-04-05
Fix JVM invalid unicode strings interop

MatkovIvan:
### The Problem
`utfToUtf8` function (that converted ["modified UTF-8"](https://docs.oracle.com/javase/1.5.0/docs/guide/jni/spec/types.html#wp16542) to standard UTF-8) accesses out-of-bounds memory in case of invalid unicode and produces invalid UTF-8 output that causes future layouting issues.

### What's changed
`utfToUtf8` was replaced to UTF-16 -> UTF-8 conversion from skia library (patched version).

Fixes https://github.com/JetBrains/compose-multiplatform/issues/2963

### DIFF
--- a/skiko/src/jvmMain/cpp/common/interop.cc
+++ b/skiko/src/jvmMain/cpp/common/interop.cc
@@ -960,116 +960,25 @@ std::unique_ptr<SkM44> skM44(JNIEnv* env, jfloatArray matrixArray) {
     }
 }
 
-// bytes  range             bits  byte 1      byte 2      byte 3      byte 4      byte 5      byte 6
-//
-// Normal UTF-8:
-//
-// 1      U+     0..    7F  7     0xxxxxxx
-// 2      U+    80..   7FF  11    110xxxxx    10xxxxxx
-// 3      U+   800..  FFFF  16    1110xxxx    10xxxxxx    10xxxxxx
-// 4      U+ 10000..10FFFF  21    11110xxx    10xxxxxx    10xxxxxx    10xxxxxx
-//
-// Modified UTF-8 (Java):
-//
-// 1      U+     1..    7F  7     0xxxxxxx
-// 2      U+     0                11000000    10000000
-// 2      U+    80..   7FF  11    110xxxxx    10xxxxxx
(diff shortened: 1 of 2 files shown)


skiko/src/jvmMain/cpp/common/interop.cc
```diff
@@ -960,116 +960,90 @@ std::unique_ptr<SkM44> skM44(JNIEnv* env, jfloatArray matrixArray) {
...
-        if ((byte1 & 0b11100000) == 0b11000000) {
-            SkASSERT((byte2 & 0b11000000) == 0b10000000);
-            data[write_offset] = byte1;
-            data[write_offset + 1] = byte2;
-            read_offset += 2;
-            write_offset += 2;
-            continue;
-        }
-
-        SkASSERT(read_offset + 2 < len);
```
MatkovIvan, line 1015: `SkASSERT` wasn't work in release mode

skiko/src/jvmMain/cpp/common/interop.cc
```diff
@@ -960,116 +960,90 @@ std::unique_ptr<SkM44> skM44(JNIEnv* env, jfloatArray matrixArray) {
...
-            continue;
+static constexpr inline bool utf16_is_surrogate(uint16_t c) { return (c - 0xD800U) < 2048U; }
+static constexpr inline bool utf16_is_high_surrogate(uint16_t c) { return (c & 0xFC00) == 0xD800; }
+static constexpr inline bool utf16_is_low_surrogate(uint16_t c) { return (c & 0xFC00) == 0xDC00; }
+
+// U+FFFD REPLACEMENT CHARACTER used to replace an unknown, unrecognized, or unrepresentable character.
+#define REPLACEMENT_CHARACTER 0xFFFD
+
+// Replacement of SkUTF::NextUTF16 to carefully handle invalid unicode characters.
+static SkUnichar NextUTF16(const uint16_t** ptr, const uint16_t* end) {
```
MatkovIvan, line 971: TODO apply this fix in our skia fork instead of making duplicate here

MatkovIvan: Since it's documented behaviour in skia, it's better do not touch it there

dima-avdeev-jb (approved):

PR #687, opened 2023-04-06
Avoid duplicated scale transform updates

manu-unter:
This is an improvement in general, but specifically with the fix from [JBR-5274](https://youtrack.jetbrains.com/issue/JBR-5274) in mind. Previously, we didn't pay attention to actual updates that did come in and we didn't update our remembered value from there. Especially since those updates come in reliably on macOS now when the display scaling changes (as soon as the change lands in our JBR builds), we would accidentally duplicate `graphicsContextScaleTransform` updates. With these changes, it should again only fire if such an update was missing.


SergeevPavel:
LGTM!

SergeevPavel:
I found that after this change `swingSkia` example stopped to work correctly. What exactly goes wrong is unclear though. Investigating

Issue #676, opened 2023-03-15
implement ClipboardManager_getText for js


julianhille:
Currently getting text from the clipboard is not implemented in js.

navigator.clipboard.getText is async might be the root cause why it is not implemented until now.

julianhille:
Wouldn't it be feasible to just return in the Javascript case a promise and the jsmain impl on the client then needs to take care of awaiting it?

Just thinking loud. This sounds doable but kinda ugly.

okushnikov:
Please check the <a href="https://youtrack.jetbrains.com/issue/SKIKO-676">following ticket</a> on YouTrack for follow-ups to this issue. GitHub issues will be closed in the coming weeks.

PR #690, opened 2023-04-17
Implement ClipboardManager_getText for js (#676)

tmdh:


dima-avdeev-jb (approved):

dima-avdeev-jb:
@tmdh Thanks!

eymar:
Unfortunately, this implementation doesn't work.

Here is its js equivalent:
```js
function getClipboardContent() { 
   var ctext = null; 
   window.navigator.clipboard.readText().then((result) => { 
       console.log("setting the result to - " + result); 
      ctext = result }
    ).catch((e) => console.log("Error = " + e));
   console.log("returning - " + ctext);
   return ctext; 
}
```
It can be tested in Chrome console with "Emulate a focus paged" enabled.

Here is a screenshot with the output:
Screenshot 2023-04-18 at 11 11 18

It can be seen that "return" happens before the callback with the result.
It works like that because "readText" return a Promise - https://developer.mozilla.org/en-US/docs/Web/API/Clipboard/readText

And the Promise won't be executed synchronously/immediately. 

So we need to revert this PR.

dima-avdeev-jb:
@eymar  Yeah, Thanks!

tmdh:
Sorry, guys. I messed up.

PR #692, opened 2023-04-19
fix `SwingSkia` example

SergeevPavel:
component listeners are performed to late so for absolute positioning we need to stick with `doLayout`


m-sasha (approved):
It's also more robust because it doesn't leak listeners if the component is added/removed several times.

PR #693, opened 2023-04-19
Use background layer size and position for CALayer; make default logger configurable

SergeevPavel:


SergeevPavel:
@manu-unter could you take a look?

skiko/src/awtMain/kotlin/org/jetbrains/skiko/redrawer/MetalRedrawer.kt
```diff
@@ -125,13 +125,13 @@ internal class MetalRedrawer(
     override fun syncSize() = synchronized(drawLock) {
         check(isEventDispatchThread()) { "Method should be called from AWT event dispatch thread" }
         val rootPane = getRootPane(layer)
-        val globalPosition = convertPoint(layer, layer.x, layer.y, rootPane)
+        val globalPosition = convertPoint(layer.backedLayer, 0, 0, rootPane)
```
manu-unter, line 128: So now we use the HardwareLayer (Canvas) origin from the Swing side, and update the CALayer to it, right? Seems about right to me

I don't even understand what the math from before should have done. It used the SkiaLayer (JPanel) origin _as coordinates whithin its own frame of reference_ which cannot be what we need. That's applying the offset of itself with respect to the content panel twice, if anything.

👍

manu-unter (approved):
Thanks for digging into this, I hope it fixed the issue!

skiko/src/awtMain/kotlin/org/jetbrains/skiko/redrawer/MetalRedrawer.kt
SergeevPavel: As I understand, this code is intended to calculate position of Skia CALayer relative to window. Seems like in examples and in our case the `layer`s origin is (0, 0) anyway, so it did work accidentally.

SergeevPavel:
@m-sasha I'm trying to dig this issue https://youtrack.jetbrains.com/issue/FL-18508/Window-layout-is-broken
The only noticeable change in this pr was mentioned above. Will try to figure out the test

PR #694, opened 2023-04-20
Expose ICU's `u_charDirection` function

MatkovIvan:
Required for https://github.com/JetBrains/compose-multiplatform/issues/2875

See https://unicode-org.github.io/icu-docs/apidoc/dev/icu4c/uchar_8h.html#aab8b83cf1d986e3726a7177ec70a21a7

It should be the same as `java.lang.Character.getDirectionality()`


skiko/src/commonMain/kotlin/org/jetbrains/skia/icu/Unicode.kt
```diff
@@ -0,0 +1,43 @@
+package org.jetbrains.skia.icu
+
+import org.jetbrains.skia.ExternalSymbolName
+import org.jetbrains.skia.impl.Library.Companion.staticLoad
+
+object CharDirection {
+    init {
+        staticLoad()
```
dima-avdeev-jb, line 8: why do we need staticLoad() ?

MatkovIvan: In case of JNI we need to make sure that library is loaded before calling any `external` methods. To see what problem exactly it cause (missing method to call bla bla) - comment it and run single test from `UnicodeTest`

skiko/src/commonMain/kotlin/org/jetbrains/skia/icu/Unicode.kt
```diff
@@ -0,0 +1,43 @@
+package org.jetbrains.skia.icu
+
+import org.jetbrains.skia.ExternalSymbolName
+import org.jetbrains.skia.impl.Library.Companion.staticLoad
+
+object CharDirection {
```
dima-avdeev-jb, line 6:
```suggestion
/**
 * CharDirection specifies the language directional property of a character set.
 * Represent UCharDirection from ICU 73.1
 * https://unicode-org.github.io/icu-docs/apidoc/dev/icu4c/uchar_8h.html
 */
object CharDirection {
```

m-sasha (approved):

dima-avdeev-jb (approved):

PR #698, opened 2023-05-02
minor note about `matchFamilyStyle` that doesn't work

SergeevPavel:


skiko/src/commonTest/kotlin/org/jetbrains/skia/FontMgrTest.kt
```diff
@@ -81,6 +81,15 @@ class FontMgrTest {
                 assertEquals(jbMono, styleSet.matchStyle(FontStyle.ITALIC))
             }
 
+            /*
+            when `TypefaceFontProvider` is used
+            `matchFamilyStyle` always returns null,
+            even if we have an appropriate styles in styleSet returned by `matchFamily`
+            It should be fixed in skia at some point
```
MatkovIvan, line 88: Let's add TODO mark here

PR #699, opened 2023-05-03
Refactor awt metal code and provide some basic docs

Walingar:
During Metal investigation, I decided to make small refactorings, so I understand code more. I hope that it also makes it more easier to read

Refactorings include:
* Moving of some native code from Redrawer to ContextHandler, since it is needed only there. And before "back" reference was used. Now ContextHandler is kinda separate class
* Provided simple k docs they can be improved later. I think I'll write more during later investigation

### DIFF
--- a/skiko/src/awtMain/kotlin/org/jetbrains/skiko/context/MetalContextHandler.kt
+++ b/skiko/src/awtMain/kotlin/org/jetbrains/skiko/context/MetalContextHandler.kt
@@ -4,16 +4,16 @@ import org.jetbrains.skia.*
 import org.jetbrains.skiko.Logger
 import org.jetbrains.skiko.RenderException
 import org.jetbrains.skiko.SkiaLayer
-import org.jetbrains.skiko.redrawer.MetalRedrawer
 
-internal class MetalContextHandler(layer: SkiaLayer) : JvmContextHandler(layer) {
-    val metalRedrawer: MetalRedrawer
-        get() = layer.redrawer!! as MetalRedrawer
-
+internal class MetalContextHandler(
+    layer: SkiaLayer,
+    private val device: Long,
+    private val adapterInfo: AdapterInfo? = null
+) : JvmContextHandler(layer) {
     override fun initContext(): Boolean {
         try {
             if (context == null) {
-                context = metalRedrawer.makeContext()
+                context = makeContext()
                 if (System.getProperty("skiko.hardwareInfo.enabled") == "true") {
                     Logger.info { "Renderer info:\n ${rendererInfo()}" }
                 }
@@ -33,7 +33,7 @@ internal class MetalContextHandler(layer: SkiaLayer) : JvmContextHandler(layer) 
         val height = (layer.backedLayer.height * scale).toInt().coerceAtLeast(0)
 
         if (width > 0 && height > 0) {
-            renderTarget = metalRedrawer.makeRenderTarget(width, height)
+            renderTarget = makeRenderTarget(width, height)
 
             surface = Surface.makeFromBackendRenderTarget(
                 context!!,
@@ -55,12 +55,31 @@ internal class MetalContextHandler(layer: SkiaLayer) : JvmContextHandler(layer) 
     override fun flush() {
         super.flush()
         surface?.flushAndSubmit()
-        metalRedrawer.finishFrame()
+        finishFrame()
     }
 
     override fun rendererInfo(): String {
         return super.rendererInfo() +
-            "Video card: ${metalRedrawer.adapterName}\n" +
-            "Total VRAM: ${metalRedrawer.adapterMemorySize / 1024 / 1024} MB\n"
+                if (adapterInfo != null) {
+                    "Video card: ${adapterInfo.adapterName}\n" + "Total VRAM: ${adapterInfo.adapterMemorySize / 1024 / 1024} MB\n"
+                } else {
+                    ""
+                }
     }
+
+    private fun makeRenderTarget(width: Int, height: Int) = BackendRenderTarget(
+        makeMetalRenderTarget(device, width, height)
+    )
+
+    private fun makeContext() = DirectContext(
+        makeMetalContext(device)
+    )
+
+    private fun finishFrame() = finishFrame(device)
+
+    private external fun makeMetalContext(device: Long): Long
+    private external fun makeMetalRenderTarget(device: Long, width: Int, height: Int): Long
+    private external fun finishFrame(device: Long)
+
+    data class AdapterInfo(val adapterName: String, val adapterMemorySize: Long)
 }
--- a/skiko/src/awtMain/kotlin/org/jetbrains/skiko/context/MetalContextHandler.kt
+++ b/skiko/src/awtMain/kotlin/org/jetbrains/skiko/context/MetalContextHandler.kt
@@ -4,10 +4,11 @@ import org.jetbrains.skia.*
 import org.jetbrains.skiko.Logger
 import org.jetbrains.skiko.RenderException
 import org.jetbrains.skiko.SkiaLayer
+import org.jetbrains.skiko.redrawer.MetalDevice
 
 internal class MetalContextHandler(
     layer: SkiaLayer,
-    private val device: Long,
+    private val device: MetalDevice,
     private val adapterInfo: AdapterInfo? = null
 ) : JvmContextHandler(layer) {
     override fun initContext(): Boolean {
@@ -68,14 +69,14 @@ internal class MetalContextHandler(
     }
 
     private fun makeRenderTarget(width: Int, height: Int) = BackendRenderTarget(
-        makeMetalRenderTarget(device, width, height)
+        makeMetalRenderTarget(device.ptr, width, height)
     )
 
     private fun makeContext() = DirectContext(
-        makeMetalContext(device)
+        makeMetalContext(device.ptr)
     )
 
-    private fun finishFrame() = finishFrame(device)
+    private fun finishFrame() = finishFrame(device.ptr)
 
     private external fun makeMetalContext(device: Long): Long
     private external fun makeMetalRenderTarget(device: Long, width: Int, height: Int): Long
--- a/skiko/src/awtMain/kotlin/org/jetbrains/skiko/context/MetalContextHandler.kt
+++ b/skiko/src/awtMain/kotlin/org/jetbrains/skiko/context/MetalContextHandler.kt
@@ -6,6 +6,14 @@ import org.jetbrains.skiko.RenderException
 import org.jetbrains.skiko.SkiaLayer
 import org.jetbrains.skiko.redrawer.MetalDevice
 
+/**
+ * Provides a way to draw on Skia canvas created in [layer] bounds using Metal GPU acceleration.
+ *
+ * For each [ContextHandler.draw] request it initializes Skia Canvas with Metal context and
+ * draws [SkiaLayer.draw] content in this canvas.
+ *
+ * @see "src/awtMain/objectiveC/macos/MetalContextHandler.mm" -- native implementation
+ */
 internal class MetalContextHandler(
     layer: SkiaLayer,
     private val device: MetalDevice,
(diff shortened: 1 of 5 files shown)


skiko/src/awtMain/kotlin/org/jetbrains/skiko/redrawer/MetalRedrawer.kt
```diff
@@ -63,8 +84,8 @@ internal class MetalRedrawer(
     override fun dispose() = synchronized(drawLock) {
         frameDispatcher.cancel()
         contextHandler.dispose()
-        disposeDevice(device)
-        device = 0
+        disposeDevice(device.ptr)
+        device = MetalDevice(0)
```
dima-avdeev-jb, line 88: Is it something like NULL pointer ?

Walingar: I think so, don't know why null wasn't used

m-sasha: Maybe `null` makes more sense, indeed.

skiko/src/awtMain/kotlin/org/jetbrains/skiko/context/MetalContextHandler.kt
```diff
@@ -55,12 +64,31 @@ internal class MetalContextHandler(layer: SkiaLayer) : JvmContextHandler(layer)
...
-        metalRedrawer.finishFrame()
+        finishFrame()
     }
 
     override fun rendererInfo(): String {
         return super.rendererInfo() +
-            "Video card: ${metalRedrawer.adapterName}\n" +
-            "Total VRAM: ${metalRedrawer.adapterMemorySize / 1024 / 1024} MB\n"
+                if (adapterInfo != null) {
+                    "Video card: ${adapterInfo.adapterName}\n" + "Total VRAM: ${adapterInfo.adapterMemorySize / 1024 / 1024} MB\n"
```
MatkovIvan, line 73: Let's override `adapterInfo.toString()` instead. (or revert line break after +)

skiko/src/awtMain/objectiveC/macos/MetalContextHandler.mm
```diff
@@ -0,0 +1,64 @@
...
+        MetalDevice *device = (__bridge MetalDevice *) (void*) devicePtr;
+        GrMtlBackendContext backendContext = {};
+        backendContext.fDevice.retain((__bridge GrMTLHandle) device.adapter);
+        backendContext.fQueue.retain((__bridge GrMTLHandle) device.queue);
+        return (jlong) GrDirectContext::MakeMetal(backendContext).release();
+    }
+}
+
+JNIEXPORT jlong JNICALL Java_org_jetbrains_skiko_context_MetalContextHandler_makeMetalRenderTarget(
+    JNIEnv * env, jobject contextHandler, jlong devicePtr, jint width, jint height)
```
MatkovIvan, line 29:
```suggestion
    JNIEnv* env, jobject contextHandler, jlong devicePtr, jint width, jint height)
```

skiko/src/awtMain/objectiveC/macos/MetalContextHandler.mm
```diff
@@ -0,0 +1,64 @@
...
+#import <mtl/GrMtlTypes.h>
+
+#import "MetalDevice.h"
+
+extern "C"
+{
+JNIEXPORT jlong JNICALL Java_org_jetbrains_skiko_context_MetalContextHandler_makeMetalContext(
+    JNIEnv* env, jobject contextHandler, jlong devicePtr)
+{
+    @autoreleasepool {
```
MatkovIvan, line 19: I do not see the reason for `autoreleasepool` here. Usually you need to put it explicitly inside loop or so, but not sure why you need it here

See https://developer.apple.com/library/archive/documentation/Cocoa/Conceptual/MemoryMgmt/Articles/mmAutoreleasePools.html

skiko/src/awtMain/objectiveC/macos/MetalContextHandler.mm
```diff
@@ -0,0 +1,64 @@
...
+
+JNIEXPORT jlong JNICALL Java_org_jetbrains_skiko_context_MetalContextHandler_makeMetalRenderTarget(
+    JNIEnv * env, jobject contextHandler, jlong devicePtr, jint width, jint height)
+{
+    @autoreleasepool {
+        MetalDevice *device = (__bridge MetalDevice *) (void *) devicePtr;
+        GrBackendRenderTarget* renderTarget = NULL;
+
+        id<CAMetalDrawable> currentDrawable = [device.layer nextDrawable];
+        if (!currentDrawable) return 0;
```
MatkovIvan, line 36:
```suggestion
        if (!currentDrawable) {
            return NULL;
        }
```

skiko/src/awtMain/objectiveC/macos/MetalContextHandler.mm
```diff
@@ -0,0 +1,64 @@
...
+        return (jlong) GrDirectContext::MakeMetal(backendContext).release();
+    }
+}
+
+JNIEXPORT jlong JNICALL Java_org_jetbrains_skiko_context_MetalContextHandler_makeMetalRenderTarget(
+    JNIEnv * env, jobject contextHandler, jlong devicePtr, jint width, jint height)
+{
+    @autoreleasepool {
+        MetalDevice *device = (__bridge MetalDevice *) (void *) devicePtr;
+        GrBackendRenderTarget* renderTarget = NULL;
```
MatkovIvan, line 33: `NULL` is just `define 0`, is C++ `nullptr` available here? if not Objective-C's `nil` will also work

skiko/src/awtMain/kotlin/org/jetbrains/skiko/redrawer/MetalRedrawer.kt
dima-avdeev-jb: we can create const val NULL: Long = 0 for this cases

dima-avdeev-jb: Or NULL_PTR: Long = 0

skiko/src/awtMain/objectiveC/macos/MetalContextHandler.mm
```diff
@@ -0,0 +1,64 @@
...
+
+        id<CAMetalDrawable> currentDrawable = [device.layer nextDrawable];
+        if (!currentDrawable) return 0;
+        device.drawableHandle = currentDrawable;
+        GrMtlTextureInfo info;
+        info.fTexture.retain((__bridge GrMTLHandle) currentDrawable.texture);
+        renderTarget = new GrBackendRenderTarget(width, height, 0, info);
```
MatkovIvan, line 40: I do not see deleting (freeing memory) of this object

skiko/src/awtMain/objectiveC/macos/MetalContextHandler.mm
```diff
@@ -0,0 +1,64 @@
...
+            id<MTLCommandBuffer> commandBuffer = [device.queue commandBuffer];
+            commandBuffer.label = @"Present";
+            [commandBuffer presentDrawable:currentDrawable];
+            [commandBuffer commit];
+            device.drawableHandle = nil;
+        }
+    }
+}
+
+} // extern C
```
MatkovIvan, line 63: You can use `extern "C"` as part of function signature to avoid such constructions/confusions

skiko/src/awtMain/kotlin/org/jetbrains/skiko/redrawer/MetalRedrawer.kt
Walingar: Changed it to null, please check the latest commit

skiko/src/awtMain/objectiveC/macos/MetalContextHandler.mm
Walingar: I haven't changed how code works. This PR is only about refactorings, if you think that memory leak can be here, I would be glad if you fix it in the separate PR

skiko/src/awtMain/objectiveC/macos/MetalContextHandler.mm
Walingar: I don't know answer for that, since this code was moved from another file, I didn't change semantic here.

skiko/src/awtMain/kotlin/org/jetbrains/skiko/context/MetalContextHandler.kt
Walingar: Not sure that it is ok to have toString with \n. Break after + will cause auto-formatting to format it in a weird way.

skiko/src/awtMain/objectiveC/macos/MetalContextHandler.mm
Walingar: Looking at the nearest files, it is done the same way with extern "C" block, so I'd keep it consistent

dima-avdeev-jb (approved):
LGTM!

Because it is just moving code from one file to another - better to not change ObjecticeC implementation. Just copy / paste as @Walingar already did.

If we want to change it - worth doing it in another PR.

skiko/src/awtMain/objectiveC/macos/MetalContextHandler.mm
Walingar: As I understand we provide this reference to jvm and there is `renderTarget = null` that will clear this reference. Doesn't it work?

skiko/src/awtMain/objectiveC/macos/MetalContextHandler.mm
MatkovIvan: Ok, keeping it here is safe, I meant it can be just redundant

Walingar: BTW, from the link that you mentioned: 
> Cocoa always expects code to be executed within an autorelease pool block, otherwise autoreleased objects do not get released and your application leaks memory.

skiko/src/awtMain/objectiveC/macos/MetalContextHandler.mm
MatkovIvan: No, operator `new` MUST be paired with `delete` (it might be not explicit in case of using smart pointer classes from std lib, but it's not a case here.
For not it looks like a leak

Walingar:
Some tests started to fail, investigating

skiko/src/awtMain/objectiveC/macos/MetalContextHandler.mm
MatkovIvan: OK, all of it are technically the same, but semantics and compiler checks a bit different. So, I'm OK to keep old C `NULL` macros here

Walingar:
Looks like it was a random flakiness. Local runs and re-run on GHA works fine

skiko/src/awtMain/objectiveC/macos/MetalContextHandler.mm
SergeevPavel: I think this `delete` is here https://github.com/JetBrains/skiko/blob/60ccb70a4cf8b7db8a1dbeaf7a9ecd905d27a6cb/skiko/src/jvmMain/cpp/common/BackendRenderTarget.cc#L7

skiko has quite complicated management of native resources though.
This particular piece is managed by `class BackendRenderTarget(...) : Managed(ptr, _FinalizerHolder.PTR)`

skiko/src/awtMain/objectiveC/macos/MetalContextHandler.mm
MatkovIvan: Yep, this is why it usually placed in `main` function. Not in every function like here.

skiko/src/awtMain/objectiveC/macos/MetalContextHandler.mm
MatkovIvan: Ok, the type is the same, but are we sure that this particular object will go through that path? Shouldn't we declare another finalizer for MetalContextHandler?

PR #704, opened 2023-05-10
Throw exception instead of force unwrapping Metal handles when initialising MetalRedrawer.

elijah-semyonov:
Sometimes people crash on iOS simulator which glitched. Throwing an explicit exception stating the problem instead of force-unwrapping and wondering where NullPointerException in crash stacktrace came from should help us to discard unjustified issue reports.


MatkovIvan (commented):
You don't need to write it in `init` and create additional local variables

pjBooms (approved):

kocheick:
Well this problem should be taken care of
 I've been developing in the same MacBook pro for 2 years until the latest OS update with Senora and sequoia, and I've been dealing with this issue but never had it before. 

How do you explain this?
It worked before until OS updates then I started getting Metal is not supported.

Now for more reference, I've actually had this issue happening but once I reboot the laptop, all goes to normal but after these updates ~ I haven't been able to do any iOS debugging I'm just relying on Android for my Compose Multiplatform app.

PR #714, opened 2023-05-25
Draft: iOS frames scheduling fix

elijah-semyonov:
Start processing pointer event : 17035.92367720834
needRedraw : 17035.92496145834
schedule for next frame : 17035.924975958336
Finish processing pointer event : 17035.924986083337
displayLink callback : 17035.925012208336
dispatch draw : 17035.925722250005

Start processing pointer event : 17037.975465708336
Finish processing pointer event : 17037.975889250003
displayLink callback : 17037.975940083335
skip : 17037.975952958335
needRedraw : 17037.97625058334
dispatch immediately : 17037.978999875002

### DIFF
--- a/skiko/src/iosMain/kotlin/org/jetbrains/skiko/SkikoUIView.kt
+++ b/skiko/src/iosMain/kotlin/org/jetbrains/skiko/SkikoUIView.kt
@@ -3,6 +3,7 @@ package org.jetbrains.skiko
 import kotlinx.cinterop.*
 import org.jetbrains.skia.Point
 import org.jetbrains.skia.Rect
+import org.jetbrains.skiko.redrawer.logEventTime
 import platform.CoreGraphics.*
 import platform.Foundation.*
 import platform.UIKit.*
@@ -181,13 +182,26 @@ class SkikoUIView : UIView, UIKeyInputProtocol, UITextInputProtocol,
         return _pointInside(skiaPoint, withEvent)
     }
 
+    private var touchesCount = 0
+        set(value) {
+            field = value
+
+            skiaLayer?.redrawer?.needsProactiveDisplayLink = value != 0
+        }
+
     override fun touchesBegan(touches: Set<*>, withEvent: UIEvent?) {
         super.touchesBegan(touches, withEvent)
+
+        touchesCount += touches.size
+
         sendTouchEventToSkikoView(withEvent!!, SkikoPointerEventKind.DOWN)
     }
 
     override fun touchesEnded(touches: Set<*>, withEvent: UIEvent?) {
         super.touchesEnded(touches, withEvent)
+
+        touchesCount -= touches.size
+
         sendTouchEventToSkikoView(withEvent!!, SkikoPointerEventKind.UP)
     }
 
@@ -198,6 +212,9 @@ class SkikoUIView : UIView, UIKeyInputProtocol, UITextInputProtocol,
 
     override fun touchesCancelled(touches: Set<*>, withEvent: UIEvent?) {
         super.touchesCancelled(touches, withEvent)
+
+        touchesCount -= touches.size
+
         sendTouchEventToSkikoView(withEvent!!, SkikoPointerEventKind.UP)
     }
 
@@ -215,6 +232,7 @@ class SkikoUIView : UIView, UIKeyInputProtocol, UITextInputProtocol,
             )
         }
 
+        logEventTime("Start processing pointer event ${NSThread.currentThread}")
         skiaLayer?.skikoView?.onPointerEvent(
             SkikoPointerEvent(
                 x = pointers.centroidX,
@@ -225,6 +243,11 @@ class SkikoUIView : UIView, UIKeyInputProtocol, UITextInputProtocol,
                 platform = event
             )
         )
+        /**
+         * If invalidation doesn't happen while onPointerEvent is processed, it's too late to schedule any work for this frame.
+         */
+        skiaLayer?.redrawer?.preventDrawDispatchDuringCurrentVsync()
+        logEventTime("Finish processing pointer event")
     }
 
     private val UITouch.isPressed get() =
--- a/skiko/src/iosMain/kotlin/org/jetbrains/skiko/redrawer/MetalRedrawer.ios.kt
+++ b/skiko/src/iosMain/kotlin/org/jetbrains/skiko/redrawer/MetalRedrawer.ios.kt
@@ -16,10 +16,7 @@ import platform.Metal.MTLCreateSystemDefaultDevice
 import platform.Metal.MTLDeviceProtocol
 import platform.Metal.MTLPixelFormatBGRA8Unorm
 import platform.QuartzCore.*
-import platform.UIKit.UIScreen
-import platform.UIKit.UIView
 import platform.UIKit.window
-import platform.darwin.NSInteger
 import platform.darwin.NSObject
 
 internal class MetalRedrawer(
@@ -101,7 +98,7 @@ internal class MetalRedrawer(
     }
 
     private fun topUpRundownCounter() {
-        rundownCounter = 3
+        rundownCounter = RUNDOWN_COUNTER_TOP_VALUE
     }
 
     override fun needRedraw() {
@@ -136,6 +133,16 @@ internal class MetalRedrawer(
                 currentDrawable = null
             }
         }
+    }
+
+    companion object {
+        /**
+         * This value indicates how many frames CADisplayLink will continue to draw after the last [needRedraw] call
+         * until it pauses itself. The value is arbitrarily chosen. Must be at least 2 to avoid incorrect scheduling.
+         * It doesn't affect correctness, and the actual optimal value (if this machinery is needed at all)
+         * will be set once [rundownCounter] _todo_ investigation is finished.
+         */
+        private const val RUNDOWN_COUNTER_TOP_VALUE = 3
     }
 }
 
--- a/skiko/src/iosMain/kotlin/org/jetbrains/skiko/redrawer/MetalRedrawer.ios.kt
+++ b/skiko/src/iosMain/kotlin/org/jetbrains/skiko/redrawer/MetalRedrawer.ios.kt
@@ -51,8 +51,7 @@ internal class MetalRedrawer(
     private val frameListener: NSObject = FrameTickListener {
         rundownCounter -= 1
 
-        if (rundownCounter <= 0) {
-            rundownCounter = 0
+        if (rundownCounter == 0) {
             caDisplayLink.setPaused(true)
         }
 
--- a/skiko/src/iosMain/kotlin/org/jetbrains/skiko/redrawer/MetalRedrawer.ios.kt
+++ b/skiko/src/iosMain/kotlin/org/jetbrains/skiko/redrawer/MetalRedrawer.ios.kt
@@ -12,12 +12,17 @@ import platform.CoreGraphics.CGContextRef
 import platform.CoreGraphics.CGSizeMake
 import platform.Foundation.NSRunLoop
 import platform.Foundation.NSSelectorFromString
+import platform.Foundation.NSThread
 import platform.Metal.MTLCreateSystemDefaultDevice
 import platform.Metal.MTLDeviceProtocol
 import platform.Metal.MTLPixelFormatBGRA8Unorm
 import platform.QuartzCore.*
 import platform.UIKit.window
 import platform.darwin.NSObject
+
+fun logEventTime(eventName: String) {
+    println("$eventName : ${CACurrentMediaTime()}")
+}
 
 internal class MetalRedrawer(
     private val layer: SkiaLayer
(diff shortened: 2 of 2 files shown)


skiko/src/iosMain/kotlin/org/jetbrains/skiko/SkikoUIView.kt
```diff
@@ -181,13 +182,26 @@ class SkikoUIView : UIView, UIKeyInputProtocol, UITextInputProtocol,
         return _pointInside(skiaPoint, withEvent)
     }
 
+    private var touchesCount = 0
+        set(value) {
+            field = value
+
+            skiaLayer?.redrawer?.needsProactiveDisplayLink = value != 0
```
MatkovIvan, line 189: A few suggestions to improve readability:
1. Move logic from setter to separate call to be more explicit
2. Wrap `value != 0` to separate getter
3. Add a comment with explanation why touches affect display link

skiko/src/iosMain/kotlin/org/jetbrains/skiko/redrawer/MetalRedrawer.ios.kt
```diff
@@ -12,16 +12,18 @@ import platform.CoreGraphics.CGContextRef
...
 import platform.Metal.MTLDeviceProtocol
 import platform.Metal.MTLPixelFormatBGRA8Unorm
 import platform.QuartzCore.*
-import platform.UIKit.UIScreen
-import platform.UIKit.UIView
 import platform.UIKit.window
-import platform.darwin.NSInteger
 import platform.darwin.NSObject
 
+fun logEventTime(eventName: String) {
```
MatkovIvan, line 23: It doesn't seem as public API

MatkovIvan (dismissed):
Blocking because `logEventTime` should not be public

skiko/src/iosMain/kotlin/org/jetbrains/skiko/SkikoUIView.kt
```diff
@@ -181,13 +182,26 @@ class SkikoUIView : UIView, UIKeyInputProtocol, UITextInputProtocol,
...
+
+        touchesCount += touches.size
+
         sendTouchEventToSkikoView(withEvent!!, SkikoPointerEventKind.DOWN)
     }
 
     override fun touchesEnded(touches: Set<*>, withEvent: UIEvent?) {
         super.touchesEnded(touches, withEvent)
+
+        touchesCount -= touches.size
```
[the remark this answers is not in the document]
pjBooms: We need to check if newly introduced variables can be accessed from multiple threads. If they can, it is better to declare them as Atomic to avoid race conditions

skiko/src/iosMain/kotlin/org/jetbrains/skiko/redrawer/MetalRedrawer.ios.kt
pjBooms: Moreover it prints something unconditionally that should not be in the final code

elijah-semyonov: Hence _DRAFT_, will be removed.

skiko/src/iosMain/kotlin/org/jetbrains/skiko/SkikoUIView.kt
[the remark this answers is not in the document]
elijah-semyonov: @MatkovIvan 
Yes, it's safe. iOS guarantees lifecycle of touches intercepted by UIView. It begins and either ends or cancels.

@pjBooms 
Touch events on iOS are dispatched on RunLoop.main, which is bound to main thread, no data races.

skiko/src/iosMain/kotlin/org/jetbrains/skiko/redrawer/MetalRedrawer.ios.kt
```diff
@@ -82,14 +134,47 @@ internal class MetalRedrawer(
...
+            if (layer.isShowing()) {
+                draw()
+            }
+
+            canDrawOnCurrentVsync = false
+
+            return true
+        }
+
+        return false
```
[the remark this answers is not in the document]
elijah-semyonov: If it's more idiomatic, sure.

skiko/src/iosMain/kotlin/org/jetbrains/skiko/redrawer/MetalRedrawer.ios.kt
```diff
@@ -33,12 +35,62 @@ internal class MetalRedrawer(
...
+            caDisplayLink.setPaused(true)
+        }
+
+        if (hasScheduledDrawForNextVsync) {
+            if (drawIfNeededDuringCurrentVsyncInterval()) {
+                hasScheduledDrawForNextVsync = false
+            } else {
+                logEventTime("discard draw")
+
+                caDisplayLink.setPaused(false)
```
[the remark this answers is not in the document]
elijah-semyonov: Rephrased it.

skiko/src/iosMain/kotlin/org/jetbrains/skiko/redrawer/MetalRedrawer.ios.kt
```diff
@@ -33,12 +35,62 @@ internal class MetalRedrawer(
...
+
+            if (value) {
+                caDisplayLink.setPaused(false)
+            }
+        }
+
+    /**
+     * Indicates that there is update that needs redraw, which will be executed once new vsync is coming
+     */
+    private var hasScheduledDrawForNextVsync = false
```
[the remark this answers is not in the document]
elijah-semyonov: Is implicit logic in setter considered a bad practice in Kotlin? In my opinion it prevents impossible states (we can't change amount of touches without causing side effects which are mandatory). Another way would be to it would be to extract it to a separate class and make it explicit and still enforce the invariant within class. But that way we can create a class for every possible invariant existing.

3 bools flags are ok here in my opinion, because there is no impossible combination of them.
Extracting a state machine to some sort of sealed interface would be just more verbose.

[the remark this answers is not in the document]
elijah-semyonov:
```
/**
 * Explanation of frame scheduling logic on iOS:
 * `frame` means an interval between two vsyncs
 *
 * needsProactiveDisplayLink
 * |     canDrawOnCurrentVsync
 * |     |     hasScheduledDrawForNextVsync
 * |     |     |
 * false false false  | DispLink is off, there is no draw scheduled, we can't draw anything immediately not to drift timings, because those are not known to us
 * false false true   | DispLink is on, we can't draw right now, but there is scheduled draw for next frame.
 * false true  false  | DispLink is on, but will be off after this frame, we can draw something, if [needRedraw] comes, or just do nothing
 * false true  true   | Impossible state, [needRedraw] would immediately dispatch draw and set both [canDrawOnCurrentVsync] and [hasScheduledDrawForNextVsync] to false
 * true  false false  | DispLink is on, but it's too late in the frame to dispatch draw, no draw is scheduled for next frame
 * true  false true   | DispLink is on, draw was either dispatched or it's too late to dispatch it this frame, draw is scheduled for next frame
 * true  true  false  | DispLink is on, we can draw during this frame immediately, no draw is scheduled on next frame
 * true  true  true   | Impossible state, [needRedraw] would immediately dispatch draw and set both [canDrawOnCurrentVsync] and [hasScheduledDrawForNextVsync] to false
 */
```

Actually I was a bit wrong, some states _are_ impossible, will try to refactor it now.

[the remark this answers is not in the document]
elijah-semyonov: Refactored in latest commit.

skiko/src/iosMain/kotlin/org/jetbrains/skiko/redrawer/MetalRedrawer.ios.kt
elijah-semyonov: Removed

skiko/src/iosMain/kotlin/org/jetbrains/skiko/redrawer/MetalRedrawer.ios.kt
```diff
@@ -16,12 +16,15 @@ import platform.Metal.MTLCreateSystemDefaultDevice
 import platform.Metal.MTLDeviceProtocol
 import platform.Metal.MTLPixelFormatBGRA8Unorm
 import platform.QuartzCore.*
-import platform.UIKit.UIScreen
-import platform.UIKit.UIView
 import platform.UIKit.window
-import platform.darwin.NSInteger
 import platform.darwin.NSObject
 
+internal enum class DrawSchedulingState(val canDrawNow: Boolean, val hasDrawScheduledForNextFrame: Boolean) {
```
pjBooms: canDrawNow is not accessed,  hasDrawScheduledForNextFrame can be replaced with == SCHEDULED_ON_NEXT_FRAME 
So I  believe this enum properties can be removed

skiko/src/iosMain/kotlin/org/jetbrains/skiko/redrawer/MetalRedrawer.ios.kt
```diff
@@ -82,14 +126,44 @@ internal class MetalRedrawer(
...
+                drawSchedulingState = DrawSchedulingState.SCHEDULED_ON_NEXT_FRAME
+            }
+
+            DrawSchedulingState.AVAILABLE_ON_CURRENT_FRAME -> {
+                drawIfLayerIsShowing()
+
+                drawSchedulingState = DrawSchedulingState.AVAILABLE_ON_NEXT_FRAME
+            }
+
+            DrawSchedulingState.SCHEDULED_ON_NEXT_FRAME -> Unit /** already scheduled, do nothing */
```
pjBooms, line 157: Remove Unit?

pjBooms (changes requested):

skiko/src/iosMain/kotlin/org/jetbrains/skiko/redrawer/MetalRedrawer.ios.kt
elijah-semyonov: Oh, it's an old commit, it was already fixed.

elijah-semyonov: Noticed it at the same time, fixed

skiko/src/iosMain/kotlin/org/jetbrains/skiko/redrawer/MetalRedrawer.ios.kt
elijah-semyonov: What's an idiomatic way to do nothing on empty `when` cases in Kotlin?

elijah-semyonov: Done.

pjBooms: I think there is no idiomatic way. We need to agree about this within our team. I would use "{}" instead of Unit. @igordmn?

elijah-semyonov: Done.

pjBooms: I would also use "//" styled comment within {}

MatkovIvan: I guess `Unit` is OK, but it can be replaced with `{}`.

Regarding having this case at all, it's better to keep it explicit instead of just `else`

elijah-semyonov: Agree with @MatkovIvan about being explicit on all enum cases.

Will replace /** */ for // inside function scopes.

pjBooms: BTW, "/**" are used only for comments that are going to be reflected in public documentation. Two stars mean that it is a public comment.

elijah-semyonov: Aha, got it! Do we have some linting tooling, I'm not aware of?

pjBooms: I believe that IDE's are configured for code styles (so when you do format code with IDE, it adjusts for code styles) but it does not catch all cases.

skiko/src/iosMain/kotlin/org/jetbrains/skiko/SkikoUIView.kt
```diff
@@ -181,13 +181,32 @@ class SkikoUIView : UIView, UIKeyInputProtocol, UITextInputProtocol,
         return _pointInside(skiaPoint, withEvent)
     }
 
+    /**
```
pjBooms, line 184: Use /* comments for private things

skiko/src/iosMain/kotlin/org/jetbrains/skiko/redrawer/MetalRedrawer.ios.kt
```diff
@@ -16,12 +16,15 @@ import platform.Metal.MTLCreateSystemDefaultDevice
 import platform.Metal.MTLDeviceProtocol
 import platform.Metal.MTLPixelFormatBGRA8Unorm
 import platform.QuartzCore.*
-import platform.UIKit.UIScreen
-import platform.UIKit.UIView
 import platform.UIKit.window
-import platform.darwin.NSInteger
 import platform.darwin.NSObject
 
+internal enum class DrawSchedulingState {
```
pjBooms, line 22: I believe it is a private enum

skiko/src/iosMain/kotlin/org/jetbrains/skiko/SkikoUIView.kt
elijah-semyonov: It has a side effect of property description not being shown on hover in IDE.

skiko/src/iosMain/kotlin/org/jetbrains/skiko/redrawer/MetalRedrawer.ios.kt
elijah-semyonov: Indeed, fixed

skiko/src/iosMain/kotlin/org/jetbrains/skiko/SkikoUIView.kt
elijah-semyonov: I guess, it's a question to IDE developers, fixed.

pjBooms (approved):

skiko/src/iosMain/kotlin/org/jetbrains/skiko/redrawer/MetalRedrawer.ios.kt
```diff
@@ -33,12 +36,55 @@ internal class MetalRedrawer(
     private val queue = device.newCommandQueue() ?: throw IllegalStateException("Couldn't create Metal command queue")
     private var currentDrawable: CAMetalDrawableProtocol? = null
     private val metalLayer = MetalLayer()
+
+    private var drawSchedulingState = DrawSchedulingState.AVAILABLE_ON_NEXT_FRAME
```
MatkovIvan, line 40: Why initial value is not `AVAILABLE_ON_CURRENT_FRAME`?

elijah-semyonov: Because precise vsync timing is not accessible without CADisplayLink callbacks. Voluntarily dispatching a frame outside of its scheduling loop can cause frame drifting in worst cases.

MatkovIvan: Please write it as a comment to this line

elijah-semyonov: Done

MatkovIvan (approved):

PR #720, opened 2023-06-06
Introduce separate SkiaSwingLayer for swing interop

Walingar:
This PR is kinda in progress of refactoring, but I would like to discuss the main idea of the interop with Swing.

Before the PR, interop was implemented using SkiaLayer that is based on awt.Canvas -- heavyweight component. It introduces some amount of bugs such as resizing, missing z ordering. Also, most of the code that is needed for standalone Compose application is not needed for interop and vice versa.

So, I've introduced `SkiaSwingLayer` that extends `SkiaSwingLayerComponent` where all needed from Compose API are presented. It reuses most of the `SkiaLayer` functionality, I've already extracted it, please take a look.

Regarding rendering, seems like we will need to have a separate bunch of native redrawers and contexthandlers, since they will work differently with some code reuse. What do you think about this approach?


Walingar:
I think, I've finished with refactoring, so folks can you please take a look at it.
I removed context handlers and moved all the logic to redrawers instead as we discussed with @igordmn. 

Also, I've copied from intellij code a part that is used for JCEF rendering, it significantly speeds up the rendering process.  (see org.jetbrains.skiko.swing.SwingOffscreenDrawer).

skiko/src/awtMain/kotlin/org/jetbrains/skiko/redrawer/RedrawerManager.kt
```diff
@@ -0,0 +1,50 @@
...
+        do {
+            thrown = false
+            try {
+                _renderApi = fallbackRenderApiQueue.removeAt(0)
+                _redrawer = redrawerFactory(_renderApi, redrawer)
+            } catch (e: RenderException) {
+                Logger.warn(e) { "Fallback to next API" }
+                thrown = true
+            }
+        } while (thrown && fallbackRenderApiQueue.isNotEmpty())
```
MatkovIvan, line 36: I'd wrap iterating logic to separate sequence/yield function

skiko/src/awtMain/kotlin/org/jetbrains/skiko/redrawer/RedrawerManager.kt
```diff
@@ -0,0 +1,50 @@
...
+
+        if (thrown) {
+            throw RenderException("Cannot fallback to any render API")
+        }
+    }
+
+    fun forceRenderApi(renderApi: GraphicsApi) {
```
MatkovIvan, line 43: `_renderApi` assigned in multiple places here, so the expected behaviour of "forcing" is not clear.

skiko/src/awtMain/kotlin/org/jetbrains/skiko/swing/SkiaSwingLayer.kt
```diff
@@ -0,0 +1,123 @@
...
+ */
+@Suppress("unused") // used in Compose Multiplatform
+@ExperimentalSkikoApi
+open class SkiaSwingLayer(
+    skikoView: SkikoView,
+    analytics: SkiaLayerAnalytics = SkiaLayerAnalytics.Empty,
+) : SkiaSwingLayerComponent() {
+    internal companion object {
+        init {
+            Library.load()
```
MatkovIvan, line 29: I do not see any `external` functions in this class. What is it for?

skiko/src/awtMain/kotlin/org/jetbrains/skiko/swing/SkiaSwingLayer.kt
```diff
@@ -0,0 +1,123 @@
...
+        }
+    }
+
+    private val properties = SkiaLayerProperties()
+
+    private var isInited = false
```
MatkovIvan, line 35: `isInitiated`?

skiko/src/awtMain/kotlin/org/jetbrains/skiko/swing/SwingOffscreenDrawer.kt
```diff
@@ -0,0 +1,274 @@
...
+        image: Image,
+        srcBounds: Rectangle?,
+        observer: ImageObserver?
+    ) {
+        var g1 = g
+        var scale1 = scale
+        var dx1 = dx
+        var dy1 = dy
+        var delta = 0.0
+        // Calculate the delta based on the image size. The bigger the size - the smaller the delta.
```
MatkovIvan, line 152: It's a good candidate to extract it to a separate function

skiko/src/awtMain/kotlin/org/jetbrains/skiko/swing/SwingOffscreenDrawer.kt
```diff
@@ -0,0 +1,274 @@
...
+        dy: Int,
+        dw: Int,
+        dh: Int,
+        hasDestinationSize: Boolean,
+        op: BufferedImageOp?,
+        image: Image,
+        srcBounds: Rectangle?,
+        observer: ImageObserver?
+    ) {
+        var g1 = g
```
MatkovIvan, line 147: I'd revisit naming in this function

skiko/src/awtMain/kotlin/org/jetbrains/skiko/swing/SwingOffscreenDrawer.kt
```diff
@@ -0,0 +1,274 @@
...
+                var dw = dw
+                var dh = dh
+                if (!hasDestinationSize) {
+                    dw = size(userWidth)
+                    dh = size(userHeight)
+                }
+                g.drawImage(/* img = */ image,
+                    /* dx1 = */ dx, /* dy1 = */ dy, /* dx2 = */ dx + dw, /* dy2 = */ dy + dh,
+                    /* sx1 = */ sx, /* sy1 = */ sy, /* sx2 = */ sx + sw, /* sy2 = */ sy + sh,
+                    /* observer = */ observer
```
MatkovIvan, line 257: Why are labels commented out?

skiko/src/awtMain/kotlin/org/jetbrains/skiko/swing/SwingOffscreenDrawer.kt
Walingar: As written in the beginning code in this file is a copy-paste from intellij sources, so I made as few changes as possible

skiko/src/awtMain/kotlin/org/jetbrains/skiko/swing/SwingOffscreenDrawer.kt
Walingar: This is a way how we can use naming parameters when calling java function

skiko/src/awtMain/kotlin/org/jetbrains/skiko/swing/SkiaSwingLayer.kt
elijah-semyonov: `isInitialized`?

skiko/src/awtMain/kotlin/org/jetbrains/skiko/swing/SkiaSwingLayer.kt
MatkovIvan: > This is the same like 


It's required ONLY in wrapper classes that have native calls

Walingar:
@elijah-semyonov @MatkovIvan as I see you also look at this PR, so can you please pay attention on skiko/src/awtMain/objectiveC/macos/MetalSwingRedrawer.mm

Here I introduced offscreen drawing, and may be I did it wrong

skiko/src/awtMain/kotlin/org/jetbrains/skiko/swing/SkiaSwingLayer.kt
Walingar: Where did you get this information? I see that no `ContextHandler`s have this call, it is only inside SkiaLayer

Walingar: I mean, for sure, I can put it in every file with an external call, but I can put it once where it is most probably will be needed.
I guess folk that wrote SkiaLayer also thought this way

Walingar: https://github.com/JetBrains/skiko/commit/55d72a8fdceea51a951f6cdc58b2c8b9c73f48e7

MatkovIvan: Checking/loading native library makes sense only if you need to call native method from it. It's placed in `companion object/init` block for making sure that native calls might be made from here (that class)

MatkovIvan: Okay, placing it in "umbrella" class works too, I was not sure that it's that case

Walingar: It is the case because we call skia and even some metal (will call directx, opengl) in nested classes

m-sasha:
Is `SkiaSwingLayerComponent` necessary? Will it have implementations other than `SkiaSwingLayer`?

Walingar:
@m-sasha it is not really necessary, it just represents the contract that is used by Compose. Otherwise, I need to suppress "not-used" inspection on public methods

skiko/src/awtMain/kotlin/org/jetbrains/skiko/swing/SkiaSwingLayer.kt
```diff
@@ -0,0 +1,123 @@
...
+ * so all Swing functionality like z-ordering, double-buffering etc. will be taken into account during rendering.
+ *
+ * But if no interop with Swing is needed, it is better to use [SkiaLayer] instead.
+ */
+@Suppress("unused") // used in Compose Multiplatform
+@ExperimentalSkikoApi
+open class SkiaSwingLayer(
+    skikoView: SkikoView,
+    analytics: SkiaLayerAnalytics = SkiaLayerAnalytics.Empty,
+) : SkiaSwingLayerComponent() {
```
igordmn, line 26: Do we need `SkiaSwingLayerComponent`? Can we extend `JComponent` directly?

skiko/src/awtMain/kotlin/org/jetbrains/skiko/redrawer/MetalRedrawer.kt
```diff
@@ -58,16 +62,13 @@ internal class MetalRedrawer(
...
     init {
-        val adapter = chooseAdapter(properties.adapterPriority.ordinal)
-        val adapterName = getAdapterName(adapter)
-        val adapterMemorySize = getAdapterMemorySize(adapter)
-        onDeviceChosen(adapterName)
+        val adapter = chooseMetalAdapter(properties.adapterPriority)
+        onDeviceChosen(adapter.name)
         val initDevice = layer.backedLayer.useDrawingSurfacePlatformInfo {
-            MetalDevice(createMetalDevice(layer.windowHandle, layer.transparency, adapter, it))
+            MetalDevice(createMetalDevice(layer.windowHandle, layer.transparency, adapter.ptr, it))
```
igordmn, line 68: `createMetalDevice` releases the hard reference to `adapter.ptr` (calling __bridge_transfer), and after this function it is not valid accessing `adapter`.

Does calling `adapter.name` work after `createMetalDevice`?

To fix the issue, we need to replace `__bridge_transfer` by `__bridge` inside `createMetalDevice`, and create `MetalAdapter.dispose` that releases the pointer via `__bridge_transfer`.

__bridge_retained - acquires the ownership, we use it to pass to java
__bridge_transfer - releases the ownership, we use it to pass from java
__bridge - no changing the ownership, just convert it for passing from/to java. the object itself usually stored in an Objective C field.

(more in [this article](https://medium.com/@notestomyself/bridge-vs-bridge-transfer-vs-bridge-retained-f84e6b6406d1))

igordmn (commented):
Looked at the new API and the old API implementation changes.

Can look at the new API implementation after the merge.

samples/SkiaAwtSample/src/main/kotlin/SkiaAwtSample/SkiaSwingPanel.kt
```diff
@@ -0,0 +1,36 @@
+package SkiaAwtSample
+
+import org.jetbrains.skiko.ClipComponent
+import org.jetbrains.skiko.SkikoView
+import org.jetbrains.skiko.swing.SkiaSwingLayer
+import java.awt.Color
+import java.awt.Component
+import javax.swing.JLayeredPane
+
+class SkiaSwingPanel(skikoView: SkikoView) : JLayeredPane() {
```
igordmn, line 10: Let's add @ExperimentalSkikoApi before merging

skiko/src/awtMain/kotlin/org/jetbrains/skiko/swing/SkiaSwingLayer.kt
Walingar: As I told here https://github.com/JetBrains/skiko/pull/720#issuecomment-1594757816
We don't need it's just a contract of what Compose uses from SkiaSwingLayer to not to have unused declarations.
But I think to reduce complexity I will remove it and suppress errors

igordmn: Ah, missed this comment. I would remove it.

> I need to suppress "not-used" inspection on public methods

This is for all public class/functions. There should be another way to suppress this inspection (but @Suppress isn't the solution as well).

skiko/src/awtMain/kotlin/org/jetbrains/skiko/redrawer/MetalRedrawer.kt
Walingar: > Does calling adapter.name work after createMetalDevice?

But it works fine and this code was written before, or do I miss something?

igordmn: We didn't use `adapter.ptr` after `createMetalDevice`. We used before it.

Walingar: But this is the same, since `adapter` is just a wrapper for `ptr`, `name`, and `memorySize`

skiko/src/awtMain/kotlin/org/jetbrains/skiko/swing/SwingRedrawerBase.kt
```diff
@@ -0,0 +1,138 @@
...
+                "OS: ${hostOs.id} ${hostArch.id}\n"
+    }
+
+    private fun initDirectContext(): Boolean {
+        try {
+            if (context == null) {
+                context = createDirectContext()
+                onContextInit()
+                if (System.getProperty("skiko.hardwareInfo.enabled") == "true") {
+                    Logger.info { "Renderer info:\n ${rendererInfo()}" }
```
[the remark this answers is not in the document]
igordmn: I made a bigger refactoring [here](https://github.com/JetBrains/skiko/pull/734). What do you think?

igordmn (approved):

igordmn:
In the future, we need to write tests, similar to SkiaLayerTest (maybe even combine them)

PR #721, opened 2023-06-07
Generate compile_commands.json from Gradle C++ compilation tasks

orangy:


skiko/buildSrc/src/main/kotlin/CompileSkikoCppTask.kt
```diff
@@ -178,6 +183,17 @@ abstract class CompileSkikoCppTask() : AbstractSkikoNativeToolTask() {
                 RunExternalProcessWork.workResults.remove(work)
             }
         }
+
+        if (compiler.get().startsWith("clang")) {
```
igordmn, line 187:
```suggestion
        // Create compile_commands.json to be able to open the project in Fleet. It is CLang convention (https://clang.llvm.org/docs/JSONCompilationDatabase.html#build-system-integration)
        if (compiler.get().startsWith("clang")) {
```

igordmn (approved):

PR #724, opened 2023-06-19
Reimplement the Canvas transformation functions 

m-sasha:
via the corresponding Skia functions directly, rather than by concatenating with a transformation matrix.

This appears to be significantly faster. Specifically in Compose, the canvas is translated for every node twice (by its position, and then back). Changing the `translate` function to perform a direct call to Skia transform makes drawing simple nodes much faster.

### DIFF
--- a/skiko/src/commonMain/kotlin/org/jetbrains/skia/Canvas.kt
+++ b/skiko/src/commonMain/kotlin/org/jetbrains/skia/Canvas.kt
@@ -1080,11 +1080,19 @@ open class Canvas internal constructor(ptr: NativePointer, managed: Boolean, int
     }
 
     fun translate(dx: Float, dy: Float): Canvas {
-        return concat(Matrix33.makeTranslate(dx, dy))
+        interopScope {
+            Stats.onNativeCall()
+            _nTranslate(_ptr, dx, dy)
+        }
+        return this
     }
 
     fun scale(sx: Float, sy: Float): Canvas {
-        return concat(Matrix33.makeScale(sx, sy))
+        interopScope {
+            Stats.onNativeCall()
+            _nScale(_ptr, sx, sy)
+        }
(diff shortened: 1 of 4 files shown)


igordmn (approved):

MatkovIvan (commented):
Could you please post your benchmark / measurements for both JVM and native?

m-sasha:
> Could you please post your benchmark / measurements for both JVM and native?

The improvement I saw was in the "Color Sparks" benchmark from the Fleet team. It's just a huge amount of small nodes that redraw each frame.

I don't think it's particularly important to "prove" the performance improvement in this case because this particular change doesn't really have a chance to make it worse. Plus, it makes sense regardless of performance.

MatkovIvan:
> this particular change doesn't really have a chance to make it worse

It does. Native calls might be expensive [1](https://stackoverflow.com/a/13977914/1703290) [2](https://www.javamex.com/tutorials/jni/overhead.shtml)


> Compared to a pure Java method call, calling a user-written [native method](https://www.javamex.com/tutorials/jni/) usually has a significant overhead. The reasons for this are as much to do with optimisations that the JVM can't make compared to regular Java methods:
>- the JVM can't inline the native method;
>- the JVM doesn't know enough about the method to make optimisations that it could make when compiling a regular Java method (for example, it has to assume that all of the parameters passed in are always used);
>- the JVM can't make other optimisations that it could make if it were dynamically compiling the code (e.g. compiling a constant parameter is a constant operand to a machine instruction rather than placing it on the stack and reading it off again);
in order to make the call into the DLL or library, the JVM may have to perform extra work, such as rearranging items on the stack.

image

m-sasha:
@MatkovIvan Indeed, but this PR replaces a call to `toInterop` + a native call with just one native call.

PR #728, opened 2023-06-23
Fix complex blending on iOS

MatkovIvan:
Skia performs complex blending via generaged shaders that require reading already drawn pixels. Setting [`framebufferOnly`](https://developer.apple.com/documentation/quartzcore/cametallayer/1478168-framebufferonly) to `false` allows it.
Google's similar issue: https://groups.google.com/g/skia-discuss/c/FTO0NSR_09Q

Partially fixes https://github.com/JetBrains/compose-multiplatform/issues/3222

PS Metal on desktop is still broken. Adding this flag is not enough there.

### DIFF
--- a/skiko/src/iosMain/kotlin/org/jetbrains/skiko/redrawer/MetalRedrawer.ios.kt
+++ b/skiko/src/iosMain/kotlin/org/jetbrains/skiko/redrawer/MetalRedrawer.ios.kt
@@ -224,6 +224,7 @@ internal class MetalLayer : CAMetalLayer {
             this.backgroundColor =
                 CGColorCreate(CGColorSpaceCreateDeviceRGB(), it.addressOf(0))
         }
+        this.framebufferOnly = false
         this.opaque = false // For UIKit interop through a "Hole"
         skiaLayer.view?.let {
             this.frame = it.frame


elijah-semyonov (approved):

skiko/src/iosMain/kotlin/org/jetbrains/skiko/redrawer/MetalRedrawer.ios.kt
```diff
@@ -224,6 +224,7 @@ internal class MetalLayer : CAMetalLayer {
             this.backgroundColor =
                 CGColorCreate(CGColorSpaceCreateDeviceRGB(), it.addressOf(0))
         }
+        this.framebufferOnly = false
```
igordmn, line 227: How will it affect FPS for UI that doesn't use blending?

Could you check on some not simple example (run1 on iOS, for example)?

MatkovIvan: Sure I'll check, I'd say some (<1ms) drop is expected. But anyway correctness is more important than that

igordmn: > correctness is more important than that

Sure, but if it would significantly affect FPS, we have alternatives.

MatkovIvan: During my tests on iPad mini 2, I do not see the difference - frame time avg is bouncing about the same values

PR #736, opened 2023-06-29
Add `configureSignAndPublishDependencies` to make sign tasks work with gradle 8.1.1

eymar:
After gradle version was updated to 8.1.1 (https://github.com/JetBrains/skiko/pull/717) CI builds for web and android started to fail.

```
Reason: Task ':publishSkikoWasmRuntimePublicationToComposeRepoRepository' uses this output of task ':signJsPublication' without declaring an explicit or implicit dependency. This can lead to incorrect results being produced, depending on what order the tasks are executed.
    
    Possible solutions:
      1. Declare task ':signJsPublication' as an input of ':publishSkikoWasmRuntimePublicationToComposeRepoRepository'.
      2. Declare an explicit dependency on ':signJsPublication' from ':publishSkikoWasmRuntimePublicationToComposeRepoRepository' using Task#dependsOn.
      3. Declare an explicit dependency on ':signJsPublication' from ':publishSkikoWasmRuntimePublicationToComposeRepoRepository' using Task#mustRunAfter.
```

```
Reason: Task ':publishAndroidPublicationToComposeRepoRepository' uses this output of task ':skikoAndroidJar' without declaring an explicit or implicit dependency. This can lead to incorrect results being produced, depending on what order the tasks are executed.
    
    Possible solutions:
      1. Declare task ':skikoAndroidJar' as an input of ':publishAndroidPublicationToComposeRepoRepository'.
      2. Declare an explicit dependency on ':skikoAndroidJar' from ':publishAndroidPublicationToComposeRepoRepository' using Task#dependsOn.
      3. Declare an explicit dependency on ':skikoAndroidJar' from ':publishAndroidPublicationToComposeRepoRepository' using Task#mustRunAfter.
            ```
            
The error message contains a suggestion how to fix the issue. This PR adds explicit dependencies between `sign` tasks and publication tasks.

___

Note: it's a bit unexpected behaviour IMO. I'm not so familiar with gradle and signing tasks. Apparantly new gradle version introduced more strict checks for tasks using outputs of other tasks. Maybe there's a way to configure it better - please let me know.

### DIFF
--- a/skiko/build.gradle.kts
+++ b/skiko/build.gradle.kts
@@ -1303,6 +1303,53 @@ if (skiko.isCIBuild || mavenCentral.signArtifacts) {
         sign(publishing.publications)
         useInMemoryPgpKeys(mavenCentral.signArtifactsKey.get(), mavenCentral.signArtifactsPassword.get())
     }
+    configureSignAndPublishDependencies()
+}
+
+
+fun configureSignAndPublishDependencies() {
+    if (supportWasm) {
+        tasks.forEach { task ->
+            val name = task.name
+            val publishJs = "publishJsPublicationTo"
+            val publishWasm = "publishSkikoWasmRuntimePublicationTo"
+            val signWasm = "signSkikoWasmRuntimePublication"
+            val signJs = "signJsPublication"
+
+            when {
+                name.startsWith(publishJs) -> task.dependsOn(signWasm)
+                name.startsWith(publishWasm) -> task.dependsOn(signJs)
+            }
+        }
+    }
+    if (supportAndroid) {
+        tasks.forEach { task ->
+            val name = task.name
+            val signAndroid = "signAndroidPublication"
+            val generateMetadata = "generateMetadataFileForAndroidPublication"
+            val publishAndroid = "publishAndroidPublicationTo"
+            val publishX64 = "publishSkikoJvmRuntimeAndroidX64PublicationTo"
+            val publishArm64 = "publishSkikoJvmRuntimeAndroidArm64PublicationTo"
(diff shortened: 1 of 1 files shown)

### CODE REVIEW

**Comment by Walingar** [comment]
@AlexeyTsvetkov maybe you know/can suggest some good alternatives?

**Review by Walingar — APPROVED** [review]

**Comment by AlexeyTsvetkov** [comment]
I'm not sure, what is the root cause of all this ([KT-46466?](https://youtrack.jetbrains.com/issue/KT-46466)). 
However, I suggest to try a different approach to configuring explicit dependencies between the tasks:
```
publishing.publications.configureEach {
    val publicationName = toTitleCase(name)
    publishing.repositories.configureEach {
        val repoName = toTitleCase(name)
        tasks.named("publish${publicationName}PublicationTo${repoName}Repository").configure {
            dependsOn("sign${publicationName}Publication")
        }
    }
}
``` 
This way is compact, more generic, and uses [recommended lazy APIs](https://docs.gradle.org/current/userguide/task_configuration_avoidance.html).
I have not tested the real publication, though.

**Comment by AlexeyTsvetkov** [comment]
The current approach is also OK!

**Comment by eymar** [comment]
@AlexeyTsvetkov I tried the approach that you suggested. It doesn't consider such a case:

```
2. Declare an explicit dependency on ':signJsPublication' from ':publishSkikoWasmRuntimePublicationToBuildRepoRepository' using Task#dependsOn.
```
`publishSkikoWasmRuntimePublicationToBuildRepoRepository` would have a dependency only on `signSkikoWasmRuntimePublication` but it has to depend on `signJsPublication.

Making the workaround more generic seems to be not straightforward. 
I'll proceed with the initial workaround, as you mentioned it's okay too.

PR #737, opened 2023-06-29
Add test for swing interop

Walingar:
I've added test for z-ordering with SkiaSwingLayer. Also, I checked that test fails with heavy-weight SkiaLayer component


Walingar:
API can be improved when more tests will be written. May be you know some other usecases that we can handle in tests? I thought about resizing, but cannot imagine how to write test for such dynamic behaviour

igordmn:
> May be you know some other usecases

I would say, we need to check all cases from SkiaLayerTest. They also applicable to SkiaSwingLayer

Walingar:
Not all, but I agree we can check smth like multiple windows and multiple components. Thank you, will write!

igordmn:
> resizing

Resizing needs to be tested for 2 cases:
- is it able to resize the content (the example of this test is in SkiaLayerTest).
- if the size of the panel can react on the layout. This test can be tested in integration with Compose in the Compose sources.

skiko/src/awtTest/kotlin/org/jetbrains/skiko/SkiaSwingLayerTest.kt
```diff
@@ -0,0 +1,87 @@
...
+                    isOpaque = false
+                    setBounds(50, 50, 50, 50)
+                }
+
+                override fun paintComponent(g: Graphics?) {
+                    val scratchGraphics = g?.create() as? Graphics2D ?: return
+                    try {
+                        scratchGraphics.setRenderingHint(
+                            RenderingHints.KEY_ANTIALIASING,
+                            RenderingHints.VALUE_ANTIALIAS_ON
```
igordmn, line 53: Better to disable antialising, it can behave differently on different systems and the screenshots will be different.

It is the case with Skia, not sure about Graphics2D. It still succeeds on Windows.

skiko/src/awtTest/kotlin/org/jetbrains/skiko/SkiaSwingLayerTest.kt
```diff
@@ -0,0 +1,87 @@
...
+                        scratchGraphics.color = Color.GREEN
+                        scratchGraphics.fillRoundRect(5, 5, 40, 40, 16, 16)
+                    } finally {
+                        scratchGraphics.dispose()
+                    }
+                }
+            }
+
+            val layeredPane = JLayeredPane()
+            layeredPane.add(popup, Integer.valueOf(1))
```
igordmn, line 64: Why in the code we have the popup behind the skia layer (the index 1 < 2), but it draws over it?

skiko/src/awtTest/kotlin/org/jetbrains/skiko/SkiaSwingLayerTest.kt
```diff
@@ -0,0 +1,87 @@
...
+            delay(1000)
+            screenshots.assert(window.bounds)
+        } finally {
+            window.close()
+        }
+    }
+
+    private fun fillSkikoView(color: Color = Color.RED): SkikoView = object : SkikoView {
+        override fun onRender(canvas: Canvas, width: Int, height: Int, nanoTime: Long) {
+            canvas.drawRect(Rect(0f, 0f, width.toFloat(), height.toFloat()), Paint().apply {
```
igordmn, line 82: Worth add a semi-transparent green background around the rect, to check blending (to the Swing rect as well)

skiko/src/awtTest/kotlin/org/jetbrains/skiko/SkiaSwingLayerTest.kt
```diff
@@ -0,0 +1,87 @@
...
+import kotlin.test.Test
+
+@OptIn(ExperimentalSkikoApi::class)
+class SkiaSwingLayerTest {
+
+    @get:Rule
+    val screenshots = ScreenshotTestRule()
+
+    @Test
+    fun `overlapped popup`() = uiTest {
```
igordmn, line 29: Worth to check also the case `overlapped skia layer`, when we draw skia content over the swing content.

skiko/src/awtTest/kotlin/org/jetbrains/skiko/SkiaSwingLayerTest.kt
Walingar: Will use `setLayer` here

skiko/src/awtTest/kotlin/org/jetbrains/skiko/SkiaSwingLayerTest.kt
Walingar: I think it is worth adding it only for swing here. Will do

skiko/src/awtTest/kotlin/org/jetbrains/skiko/SkiaSwingLayerTest.kt
Walingar: Added!

PR #738, opened 2023-06-29
Implement hostOs detection for web target

eymar:
It's going to be used in compose to propertly configure KeyMapping and other os-specific features/configurations (potentially for scroll physics, etc).

### DIFF
--- a/skiko/src/jsMain/kotlin/org/jetbrains/skiko/OsArch.js.kt
+++ b/skiko/src/jsMain/kotlin/org/jetbrains/skiko/OsArch.js.kt
@@ -47,6 +47,6 @@ internal fun detectHostOs(): OS {
         platformInfo.contains("Linux", true) -> OS.Linux
         platformInfo.contains("Mac", true) -> OS.MacOS
         platformInfo.contains("Win", true) -> OS.Windows
-        else -> OS.JS
+        else -> OS.Unknown
     }
 }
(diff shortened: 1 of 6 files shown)


skiko/src/jsMain/kotlin/org/jetbrains/skiko/OsArch.js.kt
```diff
@@ -1,6 +1,10 @@
 package org.jetbrains.skiko
 
-actual val hostOs: OS = OS.JS
+import kotlinx.browser.window
+
+actual val hostOs: OS by lazy {
```
MatkovIvan, line 5: Why should we store (by lazy) the value?

MatkovIvan: i thought about `get() =`, but all variants will work here

skiko/src/jsMain/kotlin/org/jetbrains/skiko/OsArch.js.kt
```diff
@@ -1,6 +1,8 @@
 package org.jetbrains.skiko
 
-actual val hostOs: OS = OS.JS
+import kotlinx.browser.window
+
+actual val hostOs: OS = detectHostOs()
```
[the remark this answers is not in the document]
eymar: Done. Also added a depracation for Arch.JS and Arch.WASM.
I decided to not detect the host Arch value for web for now, as I see no use cases for now.

skiko/src/jsMain/kotlin/org/jetbrains/skiko/OsArch.js.kt
eymar: I decided to not add a getter to make sure the value is calculated only once. 
The value can potentially change when the app runs (by a client), but it's not a valid case to support. We (in compose) expect this value never changes.

igordmn (approved):

PR #682, opened 2023-03-25
UIKit interop touches

dima-avdeev-jb:
It needs to interop with UIKit and touch events.

PR #752, opened 2023-07-06
iOS Remove workaround with ObjectiveC protocol.

dima-avdeev-jb:
After Kotlin 1.8.20, we can remove redundant workaround with ObjectiveC protocol. Simplify SkikoUIView.
Now, Kotlin can override extension functions like `pointInside`
```
/**
* https://developer.apple.com/documentation/uikit/uiview/1622533-point
*/
override fun pointInside(point: CValue<CGPoint>, withEvent: UIEvent?): Boolean {
```
Workaround was added in this PR: https://github.com/JetBrains/skiko/pull/682


eymar (approved):

MatkovIvan (commented):
Do we force usage of new Kotlin? Where?

dima-avdeev-jb:
> Do we force usage of new Kotlin? Where?

It was changed in previous commits to 1.8.20 https://github.com/JetBrains/skiko/blob/master/skiko/build.gradle.kts#L10

PR #758, opened 2023-07-10
Added SkikoInternalApi annotation

dima-avdeev-jb:


elijah-semyonov (approved):

skiko/src/commonMain/kotlin/org/jetbrains/skiko/Annotations.kt
```diff
@@ -27,3 +27,12 @@ annotation class DelicateSkikoApi
...
 annotation class ExperimentalSkikoApi
+
+/**
+ * Marks declarations that we did not intend to make public.
+ */
+@RequiresOptIn(
+    level = RequiresOptIn.Level.WARNING,
+    message = "This is an internal API. Please don't use it."
+)
+annotation class SkikoInternalApi
```
MatkovIvan, line 38: Why `ExperimentalSkikoApi` But `SkikoInternalApi`? I'd suggest `InternalSkikoApi` instead.

PS Please do not merge changes that changes some public things so fast

MatkovIvan: Oh, it's not to master. I see

PR #763, opened 2023-07-17
Fix dynamic framework support and fix binary compatibility of SkikoUI…

AlexeyTsvetkov:
…View

This change configures Kotlin Gradle
plugin to include a CInterop def file
(https://kotlinlang.org/docs/native-c-interop.html) to fix linking a dynamic framework for iOS

Also helps with https://youtrack.jetbrains.com/issue/KT-60399 (at least in some cases). Internal discussion https://jetbrains.slack.com/archives/G017NLN12D8/p1689594898607469


skiko/buildSrc/src/main/kotlin/utils.kt
```diff
@@ -43,3 +43,17 @@ fun Task.projectDirs(vararg relativePaths: String): List<Directory> {
     val projectDir = project.layout.projectDirectory
     return relativePaths.map { path -> projectDir.dir(path) }
 }
+
+fun listOfFrameworks(vararg frameworks: String): List<String> =
+    frameworks.flatMap { listOf("-framework", it) }
+
+fun mutableListOfLinkerOptions(options: Collection<String>): MutableList<String> =
```
igordmn, line 50: Can we also use `flatMap` here to simplify the code? (otherwise a reader should understand the difference between these two)

skiko/build.gradle.kts
```diff
@@ -427,6 +428,32 @@ kotlin {
     }
 }
 
+fun configureCinterop(
```
igordmn, line 431: Is the property `enableCInteropCommonization` for something else? (docs doesn't say its purpose clearly)

skiko/build.gradle.kts
```diff
@@ -427,6 +428,32 @@ kotlin {
...
+    arch: Arch,
+    target: KotlinNativeTarget,
+    targetString: String,
+    linkerOpts: List<String>,
+) {
+    val tasks = target.project.tasks
+    val taskNameSuffix = joinToTitleCamelCase(os.idWithSuffix(isIosSim = target.isIosSimArm64()), arch.id)
+    val writeCInteropDef = tasks.register("writeCInteropDef$taskNameSuffix", WriteCInteropDefFile::class.java) {
+        this.linkerOpts.set(linkerOpts)
+        outputFile.set(project.layout.buildDirectory.file("cinterop/$targetString/skiko.def"))
```
igordmn, line 443: Is this the reason why dynamic frameworks didn't work?

AlexeyTsvetkov: The root issue is that the linker does not know it needs to look up macOS frameworks, when linking a dynamic framework. The linker opts can be configured by our Gradle plugin or via synthetic cinterop def file. We discussed available options with @pjBooms and decided that providing the cinterop file is a better solution.

igordmn (approved):

PR #764, opened 2023-07-19
Fix memory leak because of missing autoreleasepool in offscreen rendering for Swing interop

Walingar:


elijah-semyonov (approved):

igordmn (approved):

skiko/src/awtMain/objectiveC/macos/MetalApi.mm
```diff
@@ -26,6 +26,21 @@
 extern "C"
 {
 
+extern "C" void* objc_autoreleasePoolPush(void);
+extern "C" void objc_autoreleasePoolPop(void*);
```
MatkovIvan, line 30: There is no need for `extern "C"` inside `extern "C"` block

PR #774, opened 2023-07-26
Fix segfault of color space gradient makers upon pos=null

LoadingByte:
Passing null into the "positions" parameter of a gradient shader factory that accepts a custom ColorSpace caused a segfault prior to this fix.


elijah-semyonov (approved):
Looks good to me.

igordmn (approved):

PR #776, opened 2023-07-27
Introduce Swing interop with offscreen rendering on Linux with OpenGL

Walingar:
This PR provides offscreen rendering on Graphics with GPU acceleration for Linux using OpenGL

There is still room for improvement from the performance side, like we can reuse created textures and so on. But performance is comparable with Metal (Intel) on my Intel laptop

### DIFF
--- a/skiko/src/awtMain/cpp/linux/swingRedrawer.cc
+++ b/skiko/src/awtMain/cpp/linux/swingRedrawer.cc
@@ -49,6 +49,7 @@ public:
 class OffScreenBuffer
 {
 public:
+    Display* display;
     GLXPbuffer pbuffer;
     int width;
     int height;
@@ -62,6 +63,7 @@ public:
 
         width = _width;
         height = _height;
+        display = context->display;
 
         pbuffer = glXCreatePbuffer(context->display, context->fbConfigs[0], pbufferAttribs);
     }
--- a/skiko/src/awtMain/cpp/linux/swingRedrawer.cc
+++ b/skiko/src/awtMain/cpp/linux/swingRedrawer.cc
@@ -141,8 +141,6 @@ extern "C"
     {
         OffScreenContext* context = fromJavaPointer<OffScreenContext *>(contextPtr);
         glXMakeCurrent(context->display, None, nullptr);
-        glBindFramebuffer(GL_FRAMEBUFFER, 0);
-        glBindTexture(GL_TEXTURE_2D, 0);
     }
 
     JNIEXPORT jlong JNICALL Java_org_jetbrains_skiko_swing_LinuxOpenGLSwingRedrawer_createTexture(
@@ -177,6 +175,8 @@ extern "C"
     JNIEXPORT void JNICALL Java_org_jetbrains_skiko_swing_LinuxOpenGLSwingRedrawer_disposeTexture(
         JNIEnv *env, jobject redrawer, jlong texturePtr)
     {
+        glBindFramebuffer(GL_FRAMEBUFFER, 0);
+        glBindTexture(GL_TEXTURE_2D, 0);
         OffScreenTexture *texture = fromJavaPointer<OffScreenTexture *>(texturePtr);
 
         delete texture;
--- a/skiko/src/awtMain/cpp/linux/swingRedrawer.cc
+++ b/skiko/src/awtMain/cpp/linux/swingRedrawer.cc
@@ -1,0 +1,64 @@
+#include <jawt_md.h>
+#include <GL/gl.h>
+#include <GL/glx.h>
+#include <X11/X.h>
+#include <X11/Xlib.h>
+#include <X11/Xresource.h>
+#include <cstdlib>
+#include <unistd.h>
+#include <stdio.h>
+#include "jni_helpers.h"
+
+class TextureHolder
+{
+public:
+    GLuint tex;
+    GLuint fbo;
+
+    ~TextureHolder() {
+        glDeleteFramebuffers(1, &fbo);
+        glDeleteTextures(1, &tex);
+    }
+};
+
+extern "C"
+{
+    JNIEXPORT jlong JNICALL Java_org_jetbrains_skiko_swing_LinuxOpenGLSwingRedrawer_createAndBindFrameBuffer(
+        JNIEnv *env, jobject redrawer, jint width, jint height)
+    {
+        GLuint tex;
+        glGenTextures(1, &tex);
+        glBindTexture(GL_TEXTURE_2D, tex);
+        glTexImage2D(GL_TEXTURE_2D, 0, GL_RGBA8, width, height, 0, GL_RGBA, GL_UNSIGNED_BYTE, NULL);
+
+        GLuint fbo;
+        glGenFramebuffers(1, &fbo);
+        glBindFramebuffer(GL_FRAMEBUFFER, fbo);
+
+        glFramebufferTexture2D(GL_FRAMEBUFFER, GL_COLOR_ATTACHMENT0, GL_TEXTURE_2D, tex, 0);
+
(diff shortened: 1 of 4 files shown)


skiko/src/awtMain/kotlin/org/jetbrains/skiko/swing/LinuxOpenGLSwingRedrawer.kt
```diff
@@ -0,0 +1,130 @@
...
+                flush(surface, g)
+                disposeTexture(texturePtr)
+            }
+        } finally {
+            finishRendering(offScreenContextPtr)
+        }
+    }
+
+    private fun flush(surface: Surface, g: Graphics2D) {
+        surface.flushAndSubmit(syncCpu = true)
```
Walingar, line 78: When directX will be implemented, we can extract common code here

skiko/src/awtMain/cpp/linux/swingRedrawer.cc
```diff
@@ -0,0 +1,185 @@
...
+        JNIEnv *env, jobject redrawer, jlong contextPtr)
+    {
+        OffScreenContext* context = fromJavaPointer<OffScreenContext *>(contextPtr);
+        glFinish();
+        glXMakeCurrent(context->display, None, nullptr);
+        glBindFramebuffer(GL_FRAMEBUFFER, 0);
+        glBindTexture(GL_TEXTURE_2D, 0);
+    }
+
+    JNIEXPORT jlong JNICALL Java_org_jetbrains_skiko_swing_LinuxOpenGLSwingRedrawer_createTexture(
```
igordmn, line 149: Better name it as `bindTexture`/`unbindTexture` instead of create/dispose.

It not only creates a texture, but also changes the global buffer via `glBindFramebuffer`, to which OpenGl will draw.

Also, we should move `glBindFramebuffer(GL_FRAMEBUFFER, 0)` to `unbindTexture` from `finishRendering` to be symmetric

skiko/src/awtMain/cpp/linux/swingRedrawer.cc
```diff
@@ -0,0 +1,185 @@
...
+        OffScreenBuffer* buffer = fromJavaPointer<OffScreenBuffer *>(bufferPtr);
+
+        glXMakeCurrent(context->display, buffer->pbuffer, context->context);
+    }
+
+    JNIEXPORT void JNICALL Java_org_jetbrains_skiko_swing_LinuxOpenGLSwingRedrawer_finishRendering(
+        JNIEnv *env, jobject redrawer, jlong contextPtr)
+    {
+        OffScreenContext* context = fromJavaPointer<OffScreenContext *>(contextPtr);
+        glFinish();
```
igordmn, line 143: We don't need this. It should be called by `flushAndSubmit(syncCpu = true)`.

igordmn (commented):
I only checked the code, looks good to me, thanks!

If someone else will also check on how it works on Linux, it is good to be merged.

skiko/src/awtMain/cpp/linux/swingRedrawer.cc
```diff
@@ -0,0 +1,185 @@
...
+
+class OffScreenTexture
+{
+public:
+    GLuint tex;
+    GLuint fbo;
+
+    ~OffScreenTexture() {
+        glDeleteFramebuffers(1, &fbo);
+        glDeleteTextures(1, &tex);
```
MatkovIvan, line 84: If it's not just a struct, let's move initialization to class too. Let's make `class` is set of data and functions to operate with them. Encapsulation, right?

skiko/src/awtMain/cpp/linux/swingRedrawer.cc
```diff
@@ -0,0 +1,185 @@
...
+    }
+};
+
+extern "C"
+{
+    JNIEXPORT jlong JNICALL Java_org_jetbrains_skiko_swing_LinuxOpenGLSwingRedrawer_makeOffScreenContext(
```
MatkovIvan, line 90: Don't need for indent inside `extern "C"` block

skiko/src/awtMain/cpp/linux/swingRedrawer.cc
Walingar: Here we need to bind texture and fbo to the current GL context for initialization. I don't want to have such side effects in constructor. But I'll provide constructor with passed fbo and tex

skiko/src/awtMain/cpp/linux/swingRedrawer.cc
Walingar: We have indents in all these classes

skiko/src/awtMain/cpp/linux/swingRedrawer.cc
```diff
@@ -0,0 +1,187 @@
...
+        };
+
+        display = XOpenDisplay(nullptr);
+
+        int numConfigs = 0;
+        fbConfigs = glXChooseFBConfig(display, DefaultScreen(display), glxContextAttribs, &numConfigs);
+
+        XVisualInfo* visual = glXGetVisualFromFBConfig(display, fbConfigs[0]);
+
+        context = glXCreateContext(display, visual, nullptr, True);
```
elijah-semyonov, line 37: Should we do sanity check during initialising so we could fall back to software render instead of crashing (or UB of some sort)? 
[XOpenDisplay](https://www.x.org/releases/X11R7.5/doc/man/man3/XOpenDisplay.3.html), [glXChooseFBConfig](https://registry.khronos.org/OpenGL-Refpages/gl2.1/xhtml/glXChooseFBConfig.xml), [glXGetVisualFromFBConfig](https://registry.khronos.org/OpenGL-Refpages/gl2.1/xhtml/glXGetVisualFromFBConfig.xml), [glXCreateContext](https://registry.khronos.org/OpenGL-Refpages/gl2.1/xhtml/glXCreateContext.xml) all can return null.

skiko/src/awtMain/cpp/linux/swingRedrawer.cc
```diff
@@ -0,0 +1,187 @@
...
+            GLX_PBUFFER_WIDTH, width,
+            GLX_PBUFFER_HEIGHT, height,
+            None  
+        };
+
+        width = _width;
+        height = _height;
+        display = context->display;
+
+        pbuffer = glXCreatePbuffer(context->display, context->fbConfigs[0], pbufferAttribs);
```
elijah-semyonov, line 68: https://registry.khronos.org/OpenGL-Refpages/gl2.1/xhtml/glXCreatePbuffer.xml can generate error. Should we process that?

skiko/src/awtMain/cpp/linux/swingRedrawer.cc
```diff
@@ -0,0 +1,187 @@
...
+        delete buffer;
+    }
+
+    JNIEXPORT void JNICALL Java_org_jetbrains_skiko_swing_LinuxOpenGLSwingRedrawer_startRendering(
+        JNIEnv *env, jobject redrawer, jlong contextPtr, jlong bufferPtr)
+    {
+        OffScreenContext* context = fromJavaPointer<OffScreenContext *>(contextPtr);
+        OffScreenBuffer* buffer = fromJavaPointer<OffScreenBuffer *>(bufferPtr);
+
+        glXMakeCurrent(context->display, buffer->pbuffer, context->context);
```
elijah-semyonov, line 141: [Similarly, can fail](https://registry.khronos.org/OpenGL-Refpages/gl2.1/xhtml/glXMakeCurrent.xml)
Did you test it on two screens?

Walingar: This error appear even with one screen but on window resize

PR #777, opened 2023-07-28
DirectX offscreen rendering for Swing Graphics2D

elijah-semyonov:


samples/SkiaAwtSample/src/main/kotlin/SkiaAwtSample/ClocksAwt.kt
```diff
@@ -40,8 +40,8 @@ open class ClocksAwt(private val scaleProvider: () -> Float) : SkikoView {
           strokeWidth = 1f
         }
         val watchFillHover = Paint().apply { color = 0xFFE4FF01.toInt() }
-        for (x in 0 .. (width - 50) step 50) {
-            for (y in 20 .. (height - 50) step 50) {
+        for (x in 0 .. (width - 50) step 500) {
```
Walingar, line 43: Let's revert these changes

samples/SkiaAwtSample/src/main/kotlin/SkiaAwtSample/ResizableSwingSkia.kt
```diff
@@ -11,6 +11,7 @@ fun swingSkiaResize() = SwingUtilities.invokeLater {
     window.defaultCloseOperation = WindowConstants.DISPOSE_ON_CLOSE
     window.title = "Swing Window with Compose"
 
+//    var skiaPanel: SkiaPanel? = null
```
Walingar, line 14: Let's revert these changes as well

skiko/src/awtMain/cpp/windows/directXSwingRedrawer.cc
```diff
@@ -0,0 +1,400 @@
+#ifdef SK_DIRECT3D
```
Walingar, line 1: @igordmn @MatkovIvan please take a look at this code

skiko/src/awtMain/kotlin/org/jetbrains/skiko/swing/Direct3DSwingRedrawer.kt
```diff
@@ -0,0 +1,128 @@
...
+    override fun onRender(g: Graphics2D, width: Int, height: Int, nanoTime: Long) {
+        val alignedWidth = if (width % widthSizeAlignment != 0) {
+            width + widthSizeAlignment - (width % widthSizeAlignment);
+        } else {
+            width
+        }
+
+        val newTexturePtr = getRenderTargetTexture(device, alignedWidth, height)
+
+        if (newTexturePtr == 0.toLong()) {
```
Walingar, line 60: Minor: 0.toLong() -> 0L

skiko/src/awtMain/kotlin/org/jetbrains/skiko/swing/Direct3DSwingRedrawer.kt
```diff
@@ -0,0 +1,128 @@
...
+            throw RenderException("Can't allocate DirectX resources")
+        }
+
+        if (newTexturePtr != texturePtr) {
+            texturePtr = newTexturePtr
```
Walingar, line 65: We should close renderTarget and surface in this case.
BTW can we create surface and renderTarget on each render as we do with Metal? Does it significantly affect performance?

Walingar (changes requested):

skiko/src/awtMain/cpp/windows/directXSwingRedrawer.cc
```diff
@@ -0,0 +1,400 @@
...
+        d3dDevice->backendContext.fQueue = queue;
+        d3dDevice->backendContext.fProtectedContext = GrProtected::kNo;
+
+        auto& textureDesc = d3dDevice->textureDesc;
+        textureDesc.Dimension = D3D12_RESOURCE_DIMENSION_TEXTURE2D;
+        textureDesc.Alignment = 0;
+        textureDesc.Width = 0;
+        textureDesc.Height = 0;
+        textureDesc.DepthOrArraySize = 1;
+        textureDesc.MipLevels = 1;
```
MatkovIvan, line 212: I don't think that we need it

skiko/src/awtMain/cpp/windows/directXSwingRedrawer.cc
```diff
@@ -0,0 +1,400 @@
...
+        textureDesc.Layout = D3D12_TEXTURE_LAYOUT_UNKNOWN;
+        textureDesc.Flags = D3D12_RESOURCE_FLAG_ALLOW_RENDER_TARGET;
+
+        auto& readbackBufferDesc = d3dDevice->readbackBufferDesc;
+        readbackBufferDesc.Dimension = D3D12_RESOURCE_DIMENSION_BUFFER;
+        readbackBufferDesc.Alignment = 0;
+        readbackBufferDesc.Width = 0;
+        readbackBufferDesc.Height = 1;
+        readbackBufferDesc.DepthOrArraySize = 1;
+        readbackBufferDesc.MipLevels = 1;
```
MatkovIvan, line 225: Same

skiko/src/awtMain/cpp/windows/directXSwingRedrawer.cc
```diff
@@ -0,0 +1,400 @@
...
+        }
+
+        memcpy(bytesPtr, readbackBufferBytesPtr, rangeLength);
+
+        D3D12_RANGE emptyRange{ 0, 0 };
+        device->readbackBuffer->Unmap
+        (
+            0,
+            &emptyRange
+        );
```
MatkovIvan, line 380:
```suggestion
        device->readbackBuffer->Unmap(0, &emptyRange);
```

skiko/src/awtMain/kotlin/org/jetbrains/skiko/swing/Direct3DSwingRedrawer.kt
```diff
@@ -0,0 +1,128 @@
...
+
+            val format = 28 // DXGI_FORMAT_R8G8B8A8_UNORM
```
MatkovIvan, line 67: Add a const for that

samples/SkiaAwtSample/src/main/kotlin/SkiaAwtSample/ResizableSwingSkia.kt
```diff
@@ -22,6 +23,10 @@ fun swingSkiaResize() = SwingUtilities.invokeLater {
         }
     }
 
+//    skiaPanel = SkiaPanel().also {
```
igordmn, line 26: Checked it on my machine, FPS of SkiaSwingPanel is the same as with SkiaPanel 🎉

skiko/src/awtMain/kotlin/org/jetbrains/skiko/swing/Direct3DSwingRedrawer.kt
elijah-semyonov: Yes, indeed, they leak now on each resize.  Recreating those makes sense when their wrapping entity or usage options change, which is not the case. I don't think it affects performance since those are lightweight logical entities (assuming Skia doesn't do something dumb or too smart because of this), but why would we opt-in into that way?

skiko/src/awtMain/cpp/windows/directXSwingRedrawer.cc
```diff
@@ -0,0 +1,400 @@
...
+            return 0;
+        }
+        if (adapterPtr == 0) {
+            return 0;
+        }
+        gr_cp<IDXGIAdapter1> adapter((IDXGIAdapter1 *) adapterPtr);
+
+        D3D_FEATURE_LEVEL maxSupportedFeatureLevel = D3D_FEATURE_LEVEL_12_0;
+        D3D_FEATURE_LEVEL featureLevels[] = {
+            // TODO add D3D_FEATURE_LEVEL_12_2
```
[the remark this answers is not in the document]
elijah-semyonov: @Walingar do you remember why this line was included in the scratch version?

skiko/src/awtMain/cpp/windows/directXSwingRedrawer.cc
```diff
@@ -0,0 +1,400 @@
...
+#include "../common/interop.hh"
+
+#include "d3d/GrD3DTypes.h"
+#include <d3d12sdklayers.h>
+#include "d3d/GrD3DBackendContext.h"
+#include <d3d12.h>
+#include <dxgi1_4.h>
+#include <dxgi1_6.h>
+
+class DirectXOffscreenDevice
```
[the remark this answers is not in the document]
elijah-semyonov: does it matter?

skiko/src/awtMain/cpp/windows/directXSwingRedrawer.cc
elijah-semyonov: I'd better keep it in explicit initial state to display that they are set somewhere else.

elijah-semyonov: It is needed 

https://learn.microsoft.com/en-us/windows/win32/api/d3d12/ns-d3d12-d3d12_resource_desc

skiko/src/awtMain/cpp/windows/directXSwingRedrawer.cc
elijah-semyonov: > Height, DepthOrArraySize, and MipLevels must be 1.

skiko/src/awtMain/cpp/windows/directXSwingRedrawer.cc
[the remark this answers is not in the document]
MatkovIvan: In case of class I'd not put everything to public and add methods to work with them (encapsulation, right?)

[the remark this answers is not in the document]
elijah-semyonov: You imply semantics which aren't declared somewhere in the project. Moreover, on one of previous PRs I was asked not to put any logic outside of JNI methods, because it makes it harder to navigate between JNI methods bodies and code (without a lot of tinkering).

skiko/src/awtMain/kotlin/org/jetbrains/skiko/swing/Direct3DSwingRedrawer.kt
elijah-semyonov: I think it's even better to obtain it via JNI call and avoid magic numbers

skiko/src/awtMain/cpp/windows/directXSwingRedrawer.cc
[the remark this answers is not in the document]
elijah-semyonov: Writing this code properly (using smart pointers, encapsulating all within methods) would turn this into code I was specifically asked not to produce.

skiko/src/awtMain/cpp/windows/directXSwingRedrawer.cc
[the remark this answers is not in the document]
Walingar: I don't remember :) We can remove it I think

Walingar (approved):

Walingar:
@elijah-semyonov I've refactored the code, can you please take a look at the changes?

skiko/src/awtMain/cpp/windows/directXSwingRedrawer.cc
```diff
@@ -0,0 +1,424 @@
...
+        }
+
+        if (readbackBufferResource) {
+            readbackBufferResource->Release();
+        }
+    }
+
+    int readbackBufferWidth() {
+        return readbackBufferWidth(width, height);
+    }
```
elijah-semyonov, line 131: Isn't it redundant?

Walingar: It is used in readPixels

skiko/src/awtMain/cpp/windows/directXSwingRedrawer.cc
```diff
@@ -0,0 +1,424 @@
...
+        ID3D12Resource* resource = texture->resource;
+
+        GrD3DTextureResourceInfo texResInfo = {};
+        texResInfo.fResource.retain(resource);
+        texResInfo.fResourceState = D3D12_RESOURCE_STATE_COMMON;
+        texResInfo.fFormat = DXGI_FORMAT_R8G8B8A8_UNORM;
+        texResInfo.fSampleCount = 1;
+        texResInfo.fLevelCount = 1;
+        GrBackendRenderTarget* renderTarget = new GrBackendRenderTarget(texture->width, texture->height, texResInfo);
+        return reinterpret_cast<jlong>(renderTarget);
```
elijah-semyonov, line 275: Both `resource` and `readbackBufferResource` allocations can fail inside DirectXOffScreenTexture constructor, we need to check that both are not nullptr or fallback otherwise (delete DirectXOffScreenTexture and return zero).

skiko/src/awtMain/cpp/windows/directXSwingRedrawer.cc
```diff
@@ -0,0 +1,424 @@
...
+        return toJavaPointer(d3dDevice);
+    }
+
+    JNIEXPORT jlong JNICALL Java_org_jetbrains_skiko_swing_Direct3DSwingRedrawer_makeDirectXRenderTargetOffScreen(
+            JNIEnv *env, jobject redrawer, jlong texturePtr) {
+        DirectXOffScreenTexture *texture = fromJavaPointer<DirectXOffScreenTexture *>(texturePtr);
+        ID3D12Resource* resource = texture->resource;
+
+        GrD3DTextureResourceInfo texResInfo = {};
+        texResInfo.fResource.retain(resource);
```
elijah-semyonov, line 269: If `resource` failed to allocated, it's nullptr here. Is it safe?

Walingar: If texture is created, it means that resource is allocated

Issue #784, opened 2023-08-04
Crash on Mac without Metal


avently:
I used Compose Multiplatform 1.4.0 on Mac installed in virtual machine. There is only **virtual** GPU (no Metal) which was working fine. Now I updated to 1.4.3 and started to see this after start:
```
Exception in thread "main" java.lang.NullPointerException: Parameter specified as non-null is null: method org.jetbrains.skiko.context.MetalContextHandler$AdapterInfo.<init>, parameter adapterName
        at org.jetbrains.skiko.context.MetalContextHandler$AdapterInfo.<init>(MetalContextHandler.kt)
        at org.jetbrains.skiko.redrawer.MetalRedrawer.<init>(MetalRedrawer.kt:70)
        at org.jetbrains.skiko.Actuals_awtKt$makeDefaultRenderFactory$1.createRedrawer(Actuals.awt.kt:26)
        at org.jetbrains.skiko.SkiaLayer.findNextWorkingRenderApi(SkiaLayer.awt.kt:327)
        at org.jetbrains.skiko.SkiaLayer.init(SkiaLayer.awt.kt:347)
        at org.jetbrains.skiko.SkiaLayer.addNotify(SkiaLayer.awt.kt:160)
        at androidx.compose.ui.awt.ComposeLayer$ComponentImpl.addNotify(ComposeLayer.desktop.kt:182)
        at java.desktop/java.awt.Container.addNotify(Container.java:2804)
        at java.desktop/javax.swing.JComponent.addNotify(JComponent.java:4846)
        at androidx.compose.ui.awt.ComposeWindowDelegate$_pane$1.addNotify(ComposeWindowDelegate.desktop.kt:81)
        at java.desktop/java.awt.Container.addNotify(Container.java:2804)
        at java.desktop/javax.swing.JComponent.addNotify(JComponent.java:4846)
        at java.desktop/java.awt.Container.addNotify(Container.java:2804)
        at java.desktop/javax.swing.JComponent.addNotify(JComponent.java:4846)
        at java.desktop/java.awt.Container.addNotify(Container.java:2804)
        at java.desktop/javax.swing.JComponent.addNotify(JComponent.java:4846)
        at java.desktop/javax.swing.JRootPane.addNotify(JRootPane.java:721)
        at java.desktop/java.awt.Container.addNotify(Container.java:2804)
        at java.desktop/java.awt.Window.addNotify(Window.java:791)
[...]

PR #786, opened 2023-08-08
Fix crash during MetalAdapter initialisation

elijah-semyonov:
## Changes
Correctly handle nulls when working with Obj-C API.
Throw RenderException if MTLDevice is null to allow RedrawerManager fallback to software renderer.
Minor refactor and typos fixes.

## Fixes
https://github.com/JetBrains/compose-multiplatform/issues/3460
https://github.com/JetBrains/skiko/issues/784

### DIFF
--- a/skiko/src/awtMain/kotlin/org/jetbrains/skiko/MetalApi.kt
+++ b/skiko/src/awtMain/kotlin/org/jetbrains/skiko/MetalApi.kt
@@ -8,6 +8,11 @@ package org.jetbrains.skiko
  */
 internal fun chooseMetalAdapter(adapterPriority: GpuPriority): MetalAdapter {
     val adapter = chooseAdapter(adapterPriority.ordinal)
+
+    if (adapter == 0L) {
+        throw RenderException("MetalApi.chooseAdapter returned null")
+    }
+
     val adapterName = getAdapterName(adapter)
     val adapterMemorySize = getAdapterMemorySize(adapter)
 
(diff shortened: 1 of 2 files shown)


skiko/src/awtMain/kotlin/org/jetbrains/skiko/MetalApi.kt
```diff
@@ -8,6 +8,11 @@ package org.jetbrains.skiko
  */
 internal fun chooseMetalAdapter(adapterPriority: GpuPriority): MetalAdapter {
     val adapter = chooseAdapter(adapterPriority.ordinal)
+
+    if (adapter == 0L) {
+        throw RenderException("MetalApi.chooseAdapter returned null")
```
MatkovIvan, line 13: Do we have more appropriate exception type? It's not about rendering but about initialization

MatkovIvan (approved):
LGTM

skiko/src/awtMain/kotlin/org/jetbrains/skiko/MetalApi.kt
elijah-semyonov: `RenderException` is used throughout entire skiko for initialisation and is expected in
https://github.com/JetBrains/skiko/blob/38090dc7aef0a8ac83a2cac0f9f8730b40cbf6f7/skiko/src/awtMain/kotlin/org/jetbrains/skiko/redrawer/RedrawerManager.kt#L29-L35
To fallback to software renderer.

MatkovIvan: I think that we need to subclass it for such cases at some point, not a blocker for this PR

elijah-semyonov: Probably, out of scope for this fix

PR #793, opened 2023-08-22
Accessibility focus on wrong component fix

Walingar:
Introduce nativeInitializeAccessible method that should be called on Accessible creation. Read doc for this function for details and see releated [issue](https://youtrack.jetbrains.com/issue/COMPOSE-176) for details

### DIFF
--- a/skiko/src/awtMain/objectiveC/macos/Accessibility.mm
+++ b/skiko/src/awtMain/objectiveC/macos/Accessibility.mm
@@ -1,0 +1,29 @@
+#import <jawt.h>
+#import <jawt_md.h>
+
+extern "C" {
+JNIEXPORT void JNICALL Java_org_jetbrains_skiko_AccessibilityKt_initializeCAccessible(
+    JNIEnv *env, jobject obj, jobject accessible)
+{
+    @autoreleasepool {
+        jclass accClass = env->FindClass("sun/lwawt/macosx/CAccessible");
+        if (accClass == NULL) {
+            return;
+        }
+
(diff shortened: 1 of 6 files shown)


skiko/src/awtMain/objectiveC/macos/Accessibility.mm
```diff
@@ -0,0 +1,29 @@
...
+            accClass, "getCAccessible",
+            "(Ljavax/accessibility/Accessible;)Lsun/lwawt/macosx/CAccessible;"
+        );
+        if (getCAccessibleMethod == NULL) {
+            return;
+        }
+
+        jobject result = env->CallStaticObjectMethod(accClass, getCAccessibleMethod, accessible);
+        if (result == NULL) {
+            return;
```
MatkovIvan, line 24: it looks like it's already last instruction in this function, so extra check doesn't make sense here

skiko/src/awtMain/objectiveC/macos/Accessibility.mm
```diff
@@ -0,0 +1,29 @@
...
+        if (accClass == NULL) {
+            return;
+        }
+
+        jmethodID getCAccessibleMethod = env->GetStaticMethodID(
+            accClass, "getCAccessible",
+            "(Ljavax/accessibility/Accessible;)Lsun/lwawt/macosx/CAccessible;"
+        );
+        if (getCAccessibleMethod == NULL) {
+            return;
```
MatkovIvan, line 19: Should we print some log/warning?

PR #798, opened 2023-08-29
Set kotlinx.coroutines to 1.7.3; Use limited parallelism instead of `Dispatchers.IO`

SergeevPavel:


skiko/src/awtMain/kotlin/org/jetbrains/skiko/redrawer/MetalRedrawer.kt
```diff
@@ -41,6 +38,8 @@ internal class MetalRedrawer(
     private val contextHandler: MetalContextHandler
 
     companion object {
+        @OptIn(ExperimentalCoroutinesApi::class)
+        private val dispatcherToBlockOn = Dispatchers.IO.limitedParallelism(64)
```
igordmn, line 42: Why we name it "toBlockOn"? Sounds that we block something, but we don't block in coroutines, we suspend.

skiko/src/awtMain/kotlin/org/jetbrains/skiko/redrawer/WindowsOpenGLRedrawer.kt
```diff
@@ -73,6 +74,8 @@ internal class WindowsOpenGLRedrawer(
     private fun swapBuffers() = swapBuffers(device)
 
     companion object {
+        @OptIn(ExperimentalCoroutinesApi::class)
+        private val dispatcherToBlockOn = Dispatchers.IO.limitedParallelism(64)
```
igordmn, line 78: We can't use experiemental API in skiko, as it isn't backward compatible.

This configuration will throw an exception in runtime:

App -> Kotlin 1.20 -> non-compatible-time API
App -> Compose -> Skiko -> Kotlin 1.8 -> non-compatible-time API

igordmn: Why we use limitedParallelism? the default IO has the limit 64, if I am not mistaken.

skiko/src/jvmMain/kotlin/org/jetbrains/skiko/FrameLimiter.kt
```diff
@@ -1,28 +1,28 @@
...
-private const val NanosecondsPerMillisecond = 1_000_000L
+import kotlin.time.TimeSource
 
 /**
  * Limit the duration of the frames (to avoid high CPU usage) to [frameMillis].
  * The actual delay depends on the precision of the system timer
  * (Windows has ~15ms precision by default, Linux/macOs ~2ms).
  * FrameLimiter will try to delay frames as close as possible to [frameMillis], but not greater
  */
+@OptIn(ExperimentalTime::class, ExperimentalCoroutinesApi::class)
```
igordmn, line 14: The same, we shouldn't use TimeSource here because it is experimental. At least it is in 1.8 (it is stabilized in 1.9). Not sure, if it is the same in 1.9, we need to compare, or just use the old `nanoTime`.

We can't upgrade Skiko to Kotlin 1.8, until we switch to it in Compose (this year approximately).

skiko/src/jvmMain/kotlin/org/jetbrains/skiko/FrameLimiter.kt
```diff
@@ -1,28 +1,28 @@
...
 class FrameLimiter(
     private val coroutineScope: CoroutineScope,
     private val frameMillis: () -> Long,
-    private val nanoTime: () -> Long = System::nanoTime
+    private val dispatcherToBlockOn: CoroutineDispatcher = Dispatchers.IO.limitedParallelism(64),
```
igordmn, line 18: We can pass it as part of `coroutineScope`

skiko/src/jvmMain/kotlin/org/jetbrains/skiko/FrameLimiter.kt
```diff
@@ -1,28 +1,28 @@
...
+    private val unpreciseDelay: suspend (Long) -> Unit = ::delay,
```
igordmn, line 19: Maybe "impreciseDelay"?

skiko/src/awtMain/kotlin/org/jetbrains/skiko/redrawer/MetalRedrawer.kt
SergeevPavel: in this case we actually use this dispatcher for waiting for vsync. Which means that thread from it will be blocked for some time. That the reason why we switch to `Dispatcher.IO`. The same is true for the others `dispatcherToBlockOn `

skiko/src/jvmMain/kotlin/org/jetbrains/skiko/FrameLimiter.kt
SergeevPavel: We override this dispatcher only in tests with test dispatcher that support virtual time. So I think now it's better

skiko/src/awtMain/kotlin/org/jetbrains/skiko/redrawer/WindowsOpenGLRedrawer.kt
igordmn: > Isn't it safe to use internally, if we wont expose it in public api?

It isn't safe, as an application can use different Kotlin. When it uses newer Kotlin, Skiko automatically uses it too. And if there are incompatible changes (for example, limitedParallelism is renamed), there will be a crash in skiko.

skiko/src/jvmMain/kotlin/org/jetbrains/skiko/FrameLimiter.kt
igordmn: We could make it internal, but it still seems to me redundant, as the usual Kotlin pattern is to pass dispatcher is by changing coroutineContext. I don't have big preference though, internal is also will be fine

skiko/src/awtMain/kotlin/org/jetbrains/skiko/redrawer/WindowsOpenGLRedrawer.kt
igordmn: We can add an argument to `SkiaLayer(dispatcherToBlockOn = Executors.newFixedThreadPool(64).asCoroutineDispatcher())` (or maybe even `newCachedThreadPool`) and pass it down. In Fleet we can pass `limitedParallelism`.

igordmn: After thinking more, it seems `newCachedThreadPool` is the better option for vsync for all clients (Compose/Fleet)? It doesn't reuse threads from the common coroutines pool though, but not sure if it adds significant overhead.

SergeevPavel: Maybe the ability to reuse threads from the common coroutines pool is not a big problem. I think in most cases we switch from UI thread to this executor and UI thread is not in the common pool too.

skiko/src/jvmMain/kotlin/org/jetbrains/skiko/FrameLimiter.kt
igordmn: > we don't block on anything in this class

Actually, I am not completely sure, callin `currentTime()` 15x can be slow, I added `dispatcherToBlockOn` to HardwareLayer.

We have failed test, I fixed it in the last commit. Please, review it.

igordmn (approved):

PR #799, opened 2023-08-29
Use xdg-open to open URLs on Linux if AWT does not support it

Nohus:
Fixes https://github.com/JetBrains/compose-multiplatform/issues/3583

### DIFF
--- a/skiko/src/awtMain/kotlin/org/jetbrains/skiko/Actuals.awt.kt
+++ b/skiko/src/awtMain/kotlin/org/jetbrains/skiko/Actuals.awt.kt
@@ -48,7 +48,7 @@ internal actual fun URIHandler_openUri(uri: String) {
         desktop.browse(URI(uri))
     } else when (hostOs) {
         OS.Linux -> {
-            URI(uri) // Validate URI
+            URI(uri) // Validate URI for exception behavior consistent with the Desktop.browse() case (throwing URISyntaxException)
             Runtime.getRuntime().exec(arrayOf("xdg-open", URL(uri).toString()))
         }
         OS.Android, OS.Windows, OS.MacOS, OS.Ios, OS.JS, OS.Unknown -> {
--- a/skiko/src/awtMain/kotlin/org/jetbrains/skiko/Actuals.awt.kt
+++ b/skiko/src/awtMain/kotlin/org/jetbrains/skiko/Actuals.awt.kt
@@ -46,7 +46,7 @@ internal actual fun URIHandler_openUri(uri: String) {
         desktop.browse(URI(uri))
     } else when (hostOs) {
         OS.Linux -> {
-            Runtime.getRuntime().exec(arrayOf("xdg-open", uri))
+            Runtime.getRuntime().exec(arrayOf("xdg-open", URI(uri).toString()))
         }
         OS.Android, OS.Windows, OS.MacOS, OS.Ios, OS.JS, OS.Unknown -> {
             throw UnsupportedOperationException("AWT does not support the BROWSE action on this platform")
--- a/skiko/src/awtMain/kotlin/org/jetbrains/skiko/Actuals.awt.kt
+++ b/skiko/src/awtMain/kotlin/org/jetbrains/skiko/Actuals.awt.kt
@@ -41,7 +41,17 @@ internal actual fun makeDefaultRenderFactory(): RenderFactory =
     }
 
 internal actual fun URIHandler_openUri(uri: String) {
-    Desktop.getDesktop().browse(URI(uri))
+    val desktop = Desktop.getDesktop()
+    if (desktop.isSupported(Desktop.Action.BROWSE)) {
+        desktop.browse(URI(uri))
+    } else when (hostOs) {
+        OS.Linux -> {
+            Runtime.getRuntime().exec("xdg-open $uri")
+        }
+        OS.Android, OS.Windows, OS.MacOS, OS.Ios, OS.JS, OS.Unknown -> {
+            throw UnsupportedOperationException("AWT does not support the BROWSE action on this platform")
+        }
+    }
 }
 
 private val systemClipboard by lazy {
--- a/skiko/src/awtMain/kotlin/org/jetbrains/skiko/Actuals.awt.kt
+++ b/skiko/src/awtMain/kotlin/org/jetbrains/skiko/Actuals.awt.kt
@@ -49,11 +49,7 @@ internal actual fun URIHandler_openUri(uri: String) {
     } else when (hostOs) {
         OS.Linux -> {
             URI(uri) // Validate URI
-            try {
-                Runtime.getRuntime().exec(arrayOf("xdg-open", URL(uri).toString()))
-            } catch (e: MalformedURLException) {
-                throw IllegalArgumentException("URI is not a valid URL", e)
-            }
+            Runtime.getRuntime().exec(arrayOf("xdg-open", URL(uri).toString()))
         }
         OS.Android, OS.Windows, OS.MacOS, OS.Ios, OS.JS, OS.Unknown -> {
             throw UnsupportedOperationException("AWT does not support the BROWSE action on this platform")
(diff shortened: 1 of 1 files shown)


Nohus:
@m-sasha Sorry for the ping, but I saw you reviewed a newer PR recently, but this one is still unacknowledged. Totally understandable if more time is needed on this, just wanted to make sure someone it didn't get lost.

m-sasha:
Hi. The code looks ok, but I don't know if we want to hardcode the use of a specific commandline tool to open URLs. Especially since it's easy for developers to do this themselves.

m-sasha:
Also you should probably use the version of Runtime.exec that accepts an array of commands. Otherwise this is a potential injection attack vulnerability.

Nohus:
The current behavior is to crash with an `UnsupportedOperationException`, and `xdg-open` is a [standard utility](https://www.freedesktop.org/wiki/Software/xdg-utils/) that is present on virtually all Linux systems with a graphical desktop environment, and it's commonly used for this purpose. As an example, Google Chrome calls `xdg-open` on Linux to open URLs with a scheme it cannot open itself.

> Also you should probably use the version of Runtime.exec that accepts an array of arguments

Fair point, fixed.

m-sasha:
Hi, a few changes we'd like to request.  The function should behave the same as `Desktop.browse`:
 - It should throw `IOException` it it failed to open the URL. You need to wait for the return code of `xdg-open` and respond accordingly.
 - It should throw `IllegalArgumentException` if the argument string is not a valid URL.

Nohus:
> It should throw IOException it it failed to open the URL. You need to wait for the return code of xdg-open and respond accordingly.

My worry with this is that waiting for the process to finish takes around 150-200ms (on my computer), and so executing this from the UI thread would visibly hang the application for a moment when a user clicks a link.

> It should throw IllegalArgumentException if the argument string is not a valid URL.

Done. Though it throws `URISyntaxException`, which is what the existing code does.

m-sasha:
Any idea why it takes so long?200ms seems like a lot for this.

Nohus:
image

I have measured it with `strace` and looks like `xdg-open` spends 99% of that time waiting for `kde-open5` to which it delegates opening the URL on KDE. In turn `kde-open5` is a QT application and it takes it's time to load fonts, languages, QT libraries, MIME type files, etc. `strace` tells me it opens 432 files in it's executing, which are all filesystem operations that take time. Definitely seems like there is room for improvement in `kde-open5`, but in any case, on different systems `xdg-open` may delegate opening the URL to different utilities, and so how long it takes will vary.

m-sasha:
In theory, it shouldn't be a problem, because it's easy to launch this action in a coroutine on Dispatcher.IO, but in practice people may not expect it to be a long operation, and we don't have a place to document this such that developers will definitely notice it. There's no documentation on UriHandler.openUri about errors, or whether it's supposed to return immediately.

@igordmn what do you think is more important here? Failing correctly or not blocking?

Nohus:
Going by the "rule of least surprise", I would say developers wouldn't expect this call to be blocking, but the call silently failing due to the user having a misconfigured default browser setting wouldn't be unexpected.

In all internet browsers I tried clicking a link with an invalid scheme just does nothing with no indication of failure. (And it seems Google Chrome does not check the return status of `xdg-open` on Linux).

skiko/src/awtMain/kotlin/org/jetbrains/skiko/Actuals.awt.kt
```diff
@@ -41,7 +41,17 @@ internal actual fun makeDefaultRenderFactory(): RenderFactory =
     }
 
 internal actual fun URIHandler_openUri(uri: String) {
-    Desktop.getDesktop().browse(URI(uri))
+    val desktop = Desktop.getDesktop()
+    if (desktop.isSupported(Desktop.Action.BROWSE)) {
+        desktop.browse(URI(uri))
+    } else when (hostOs) {
+        OS.Linux -> {
+            Runtime.getRuntime().exec(arrayOf("xdg-open", URI(uri).toString()))
```
m-sasha, line 49: It should parse as a URL, not just a URI.

m-sasha (changes requested):

skiko/src/awtMain/kotlin/org/jetbrains/skiko/Actuals.awt.kt
Nohus: Done. To preserve the existing exception behavior, I first parse as URI and then as URL. This way the user code only gets URI exceptions as in all other cases, and never the different `MalformedURLException`.

m-sasha: It shouldn't throw `UnsupportedOperationException` in this case, because the operation is supported; it's the argument that's wrong. It should throw `IllegalArgumentException` (per `Desktop.browse` docs).

Nohus: > It should throw IllegalArgumentException (per Desktop.browse docs).

I didn't realize the URI has to be a valid URL, based on the method name `URIHandler_openUri` I would think it can handle URIs and not only specifically URLs. Why is not called `URLHandler_openUrl` if other valid URIs are considered illegal arguments? Confused me. Updated.

Nohus: Not sure if we can delete the `URI(uri)` validation if we are validating the URL anyway. I would think we can delete it, but maybe there is some edge case scenario where a valid URL is somehow not a valid URI? Not sure.

m-sasha: `URIHandler_openUri` is supposed to handle URIs too, but `xdg-open` can't, and I prefer to sanitize the input as much as possible before passing it to a 3rd-party utility.

> Not sure if we can delete the URI(uri) validation I added on line 51 if we are validating the URL anyway. I would think we can delete it, but maybe there is some edge case scenario where a valid URL is somehow not a valid URI? Not sure.

We probably can. I understand URLs are a subset of URIs.

Nohus: > URIHandler_openUri is supposed to handle URIs too, but xdg-open can't, and I prefer to sanitize the input as much as possible before passing it to a 3rd-party utility.

If `URIHandler_openUri` is supposed to handle URIs, and the user has passed in a valid URI, then is it really an `IllegalArgumentException` that `xdg-open` doesn't support that URI (because it's not a URL)? This was my thinking here.

m-sasha: I see what you mean. Maybe throw `IOException` then (or pass through the `MalformedUrlException`; it's a subclass of `IOException`, strangely).

Nohus: All right, `MalformedUrlException` now goes to the caller (in the rare case a valid URI is not a valid URL).

skiko/src/awtMain/kotlin/org/jetbrains/skiko/Actuals.awt.kt
```diff
@@ -41,7 +43,18 @@ internal actual fun makeDefaultRenderFactory(): RenderFactory =
     }
 
 internal actual fun URIHandler_openUri(uri: String) {
-    Desktop.getDesktop().browse(URI(uri))
+    val desktop = Desktop.getDesktop()
+    if (desktop.isSupported(Desktop.Action.BROWSE)) {
+        desktop.browse(URI(uri))
+    } else when (hostOs) {
+        OS.Linux -> {
+            URI(uri) // Validate URI
```
m-sasha, line 51: This seems redundant now.

Nohus: This is to keep the behavior consistent with other platforms using `Desktop.browse()`, where `URI()` throws `URISyntaxException`. If a developer relies on catching `URISyntaxException`, then that will break on Linux, but with this it won't.

m-sasha: No, it sounds reasonable, but please add a short comment explaining this reasoning.

Nohus: Added comment

m-sasha (approved):

m-sasha:
Thanks for the PR.

PR #811, opened 2023-10-04
Fix crash on Windows machines without OpenGL

igordmn:
Some Windows CI machines don't contain OpenGL, so Skiko/Compose will crash on them. It shouldn't crash, it should fallback to Software rendering

To properly fallback, we need to load opengl32.dll dynamically in runtime.

Fixes https://github.com/JetBrains/compose-multiplatform/issues/3243

# Test 1
- `gradlew publishToMavenLocal` is successful (if we use functions directly, it fails with "error LNK2019: unresolved external symbol")
- `SkiaLayerTest` is successful (it tests OpenGL as well)

# Test 2
1. Remove `C:\Windows\System32\opengl32.dll`
2. `SkiaLayerTest` still should be successful, but logs should contain that it fallbacks to Software rendering.

# Test 3
1. Compile local skiko
2. Run Compose run1 and tests

### DIFF
--- a/skiko/src/awtMain/cpp/windows/OpenGLLibrary.cc
+++ b/skiko/src/awtMain/cpp/windows/OpenGLLibrary.cc
@@ -2,18 +2,18 @@
 #include <gl/GL.h>
 #include <jawt_md.h>
 
-typedef void (*PROC_glFinish) (void);
-typedef void (*PROC_glGetIntegerv) (GLenum pname, GLint *data);
-typedef const GLubyte *(*PROC_glGetString) (GLenum name);
-typedef HGLRC (WINAPI * PROC_wglCreateContext) (HDC hDc);
-typedef BOOL (WINAPI * PROC_wglDeleteContext) (HGLRC oldContext);
-typedef PROC (WINAPI * PROC_wglGetProcAddress) (LPCSTR lpszProc);
-typedef BOOL (WINAPI * PROC_wglMakeCurrent) (HDC hDc, HGLRC newContext);
-typedef HGLRC (WINAPI * PROC_wglGetCurrentContext) (void);
+namespace OpenGL32Library {
+    typedef void (*PROC_glFinish) (void);
+    typedef void (*PROC_glGetIntegerv) (GLenum pname, GLint *data);
+    typedef const GLubyte *(*PROC_glGetString) (GLenum name);
+    typedef HGLRC (WINAPI * PROC_wglCreateContext) (HDC hDc);
+    typedef BOOL (WINAPI * PROC_wglDeleteContext) (HGLRC oldContext);
+    typedef PROC (WINAPI * PROC_wglGetProcAddress) (LPCSTR lpszProc);
+    typedef BOOL (WINAPI * PROC_wglMakeCurrent) (HDC hDc, HGLRC newContext);
+    typedef HGLRC (WINAPI * PROC_wglGetCurrentContext) (void);
 
-class OpenGL32Library {
-public:
-    HINSTANCE lib;
+    bool isLoaded = false;
+    HINSTANCE lib = 0;
 
     PROC_glFinish glFinish;
     PROC_glGetIntegerv glGetIntegerv;
@@ -24,62 +24,60 @@ public:
     PROC_wglMakeCurrent wglMakeCurrent;
     PROC_wglGetCurrentContext wglGetCurrentContext;
 
-    OpenGL32Library(HINSTANCE lib) {
-        this->lib = lib;
-        glFinish = (PROC_glFinish) GetProcAddress(lib, "glFinish");
-        glGetIntegerv = (PROC_glGetIntegerv) GetProcAddress(lib, "glGetIntegerv");
-        glGetString = (PROC_glGetString) GetProcAddress(lib, "glGetString");
-        wglCreateContext = (PROC_wglCreateContext) GetProcAddress(lib, "wglCreateContext");
-        wglDeleteContext = (PROC_wglDeleteContext) GetProcAddress(lib, "wglDeleteContext");
-        wglGetProcAddress = (PROC_wglGetProcAddress) GetProcAddress(lib, "wglGetProcAddress");
-        wglMakeCurrent = (PROC_wglMakeCurrent) GetProcAddress(lib, "wglMakeCurrent");
-        wglGetCurrentContext = (PROC_wglGetCurrentContext) GetProcAddress(lib, "wglGetCurrentContext");
+    void load() {
+        if (!isLoaded) {
+            isLoaded = true;
+            lib = LoadLibrary("opengl32.dll");
+            if (lib) {
+                glFinish = (PROC_glFinish) GetProcAddress(lib, "glFinish");
+                glGetIntegerv = (PROC_glGetIntegerv) GetProcAddress(lib, "glGetIntegerv");
+                glGetString = (PROC_glGetString) GetProcAddress(lib, "glGetString");
+                wglCreateContext = (PROC_wglCreateContext) GetProcAddress(lib, "wglCreateContext");
+                wglDeleteContext = (PROC_wglDeleteContext) GetProcAddress(lib, "wglDeleteContext");
+                wglGetProcAddress = (PROC_wglGetProcAddress) GetProcAddress(lib, "wglGetProcAddress");
+                wglMakeCurrent = (PROC_wglMakeCurrent) GetProcAddress(lib, "wglMakeCurrent");
+                wglGetCurrentContext = (PROC_wglGetCurrentContext) GetProcAddress(lib, "wglGetCurrentContext");
+            }
+        }
     }
 };
-
-OpenGL32Library* openGL32Library = nullptr;
 
 extern "C"
 {
     void glFinish(void) {
-        openGL32Library->glFinish();
+        OpenGL32Library::glFinish();
     }
 
     void glGetIntegerv(GLenum pname, GLint *data) {
-        openGL32Library->glGetIntegerv(pname, data);
+        OpenGL32Library::glGetIntegerv(pname, data);
     }
 
     const GLubyte * glGetString(GLenum name) {
-        return openGL32Library->glGetString(name);
+        return OpenGL32Library::glGetString(name);
     }
 
     HGLRC WINAPI wglCreateContext(HDC hDc) {
-        return openGL32Library->wglCreateContext(hDc);
+        return OpenGL32Library::wglCreateContext(hDc);
     }
 
     BOOL WINAPI wglDeleteContext(HGLRC oldContext) {
-        return openGL32Library->wglDeleteContext(oldContext);
+        return OpenGL32Library::wglDeleteContext(oldContext);
     }
 
     PROC WINAPI wglGetProcAddress(LPCSTR lpszProc) {
-        return openGL32Library->wglGetProcAddress(lpszProc);
+        return OpenGL32Library::wglGetProcAddress(lpszProc);
     }
 
     BOOL WINAPI wglMakeCurrent(HDC hDc, HGLRC newContext) {
-        return openGL32Library->wglMakeCurrent(hDc, newContext);
+        return OpenGL32Library::wglMakeCurrent(hDc, newContext);
     }
 
     HGLRC WINAPI wglGetCurrentContext() {
-        return openGL32Library->wglGetCurrentContext();
+        return OpenGL32Library::wglGetCurrentContext();
     }
 
     JNIEXPORT jboolean JNICALL Java_org_jetbrains_skiko_OpenGLLibraryKt_loadOpenGLLibraryWindows(JNIEnv *env, jobject obj) {
-        if (openGL32Library == 0) {
-            HINSTANCE lib = LoadLibrary("opengl32.dll");
-            if (lib) {
-                openGL32Library = new OpenGL32Library(lib);
-            }
-        }
-        return openGL32Library != 0;
+        OpenGL32Library::load();
+        return OpenGL32Library::lib != 0;
     }
 }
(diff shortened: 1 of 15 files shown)


skiko/src/awtMain/cpp/windows/OpenGLLibrary.cc
```diff
@@ -0,0 +1,85 @@
...
+    }
+
+    JNIEXPORT jboolean JNICALL Java_org_jetbrains_skiko_OpenGLLibraryKt_loadOpenGLLibraryWindows(JNIEnv *env, jobject obj) {
+        if (openGL32Library == 0) {
+            HINSTANCE lib = LoadLibrary("opengl32.dll");
+            if (lib) {
+                openGL32Library = new OpenGL32Library(lib);
+            }
+        }
+        return openGL32Library != 0;
```
MatkovIvan:
```suggestion
        return openGL32Library != nullptr;
```

skiko/src/awtMain/cpp/windows/OpenGLLibrary.cc
```diff
@@ -0,0 +1,85 @@
...
+    BOOL WINAPI wglMakeCurrent(HDC hDc, HGLRC newContext) {
+        return openGL32Library->wglMakeCurrent(hDc, newContext);
+    }
+
+    HGLRC WINAPI wglGetCurrentContext() {
+        return openGL32Library->wglGetCurrentContext();
+    }
+
+    JNIEXPORT jboolean JNICALL Java_org_jetbrains_skiko_OpenGLLibraryKt_loadOpenGLLibraryWindows(JNIEnv *env, jobject obj) {
+        if (openGL32Library == 0) {
```
MatkovIvan:
```suggestion
        if (openGL32Library == nullptr) {
```

BTW it's not thread safe. Is it OK here?

skiko/src/jvmMain/cpp/common/openglapi.cc
```diff
@@ -15,30 +15,10 @@
...
-
-JNIEXPORT void JNICALL Java_org_jetbrains_skiko_OpenGLApi_glClearColor(JNIEnv * env, jobject object, jfloat r, jfloat g, jfloat b, jfloat a) {
-	glClearColor(r, g, b, a);
-}
-
-JNIEXPORT void JNICALL Java_org_jetbrains_skiko_OpenGLApi_glClear(JNIEnv * env, jobject object, jint mask) {
-	glClear(mask);
-}
-
 JNIEXPORT void JNICALL Java_org_jetbrains_skiko_OpenGLApi_glFinish(JNIEnv * env, jobject object) {
```
MatkovIvan, line 18: Why do you keep part of the functions here?

skiko/src/awtMain/cpp/windows/OpenGLLibrary.cc
```diff
@@ -0,0 +1,85 @@
...
+            HINSTANCE lib = LoadLibrary("opengl32.dll");
```
MatkovIvan: [Docs](https://www.khronos.org/opengl/wiki/Load_OpenGL_Functions) suggest to try `wglGetProcAddress` before that

skiko/src/awtMain/cpp/windows/OpenGLLibrary.cc
```diff
@@ -0,0 +1,85 @@
...
+        wglGetProcAddress = (PROC_wglGetProcAddress) GetProcAddress(lib, "wglGetProcAddress");
+        wglMakeCurrent = (PROC_wglMakeCurrent) GetProcAddress(lib, "wglMakeCurrent");
+        wglGetCurrentContext = (PROC_wglGetCurrentContext) GetProcAddress(lib, "wglGetCurrentContext");
+    }
+};
+
+OpenGL32Library* openGL32Library = nullptr;
+
+extern "C"
+{
```
MatkovIvan, line 46:
```suggestion
extern "C" {
```

skiko/src/awtMain/cpp/windows/OpenGLLibrary.cc
```diff
@@ -0,0 +1,85 @@
...
+    void glFinish(void) {
```
MatkovIvan, line 47: Do we really need to examine linker to resolve functions with names that already exists? Can we just call our owns (with prefix or so)?

skiko/src/awtMain/cpp/windows/OpenGLLibrary.cc
```diff
@@ -24,62 +24,60 @@ class OpenGL32Library {
...
     HGLRC WINAPI wglGetCurrentContext() {
-        return openGL32Library->wglGetCurrentContext();
+        return OpenGL32Library::wglGetCurrentContext();
     }
 
     JNIEXPORT jboolean JNICALL Java_org_jetbrains_skiko_OpenGLLibraryKt_loadOpenGLLibraryWindows(JNIEnv *env, jobject obj) {
-        if (openGL32Library == 0) {
-            HINSTANCE lib = LoadLibrary("opengl32.dll");
-            if (lib) {
-                openGL32Library = new OpenGL32Library(lib);
```
MatkovIvan: No dynamic alloc 🎉  That's better, thanks

skiko/src/awtMain/cpp/windows/OpenGLLibrary.cc
igordmn: Done

skiko/src/awtMain/cpp/windows/OpenGLLibrary.cc
igordmn: Outdated

skiko/src/awtMain/cpp/windows/OpenGLLibrary.cc
igordmn: Not sure that it was matter, as it should only be called once and create a small object

skiko/src/awtMain/cpp/windows/OpenGLLibrary.cc
MatkovIvan: Actually no, let me add it to new place then

skiko/src/awtMain/cpp/windows/OpenGLLibrary.cc
```diff
@@ -0,0 +1,85 @@
...
+    typedef HGLRC (WINAPI * PROC_wglCreateContext) (HDC hDc);
+    typedef BOOL (WINAPI * PROC_wglDeleteContext) (HGLRC oldContext);
+    typedef PROC (WINAPI * PROC_wglGetProcAddress) (LPCSTR lpszProc);
+    typedef BOOL (WINAPI * PROC_wglMakeCurrent) (HDC hDc, HGLRC newContext);
+    typedef HGLRC (WINAPI * PROC_wglGetCurrentContext) (void);
+
+    bool isLoaded = false;
+    HINSTANCE lib = 0;
+
+    PROC_glFinish glFinish;
```
MatkovIvan, line 18: Same for these global vars. There are two options here:
1. mark it as `static` (no linker symbol for it, will be available just in this file)
2. Move it inside function (my vote for it). Like
```c
void glFinish(void) {
    typedef void (*PROC_glFinish) (void);
    static PROC_glFinish proc = (PROC_glFinish) GetProcAddress(lib, "glFinish"); // Will be called once and stored between function calls
    proc();
}
```

skiko/src/awtMain/cpp/windows/OpenGLLibrary.cc
```diff
@@ -0,0 +1,85 @@
...
+    typedef void (*PROC_glGetIntegerv) (GLenum pname, GLint *data);
+    typedef const GLubyte *(*PROC_glGetString) (GLenum name);
+    typedef HGLRC (WINAPI * PROC_wglCreateContext) (HDC hDc);
+    typedef BOOL (WINAPI * PROC_wglDeleteContext) (HGLRC oldContext);
+    typedef PROC (WINAPI * PROC_wglGetProcAddress) (LPCSTR lpszProc);
+    typedef BOOL (WINAPI * PROC_wglMakeCurrent) (HDC hDc, HGLRC newContext);
+    typedef HGLRC (WINAPI * PROC_wglGetCurrentContext) (void);
+
+    bool isLoaded = false;
+    HINSTANCE lib = 0;
```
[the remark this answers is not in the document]
igordmn: Done

skiko/src/awtMain/cpp/windows/OpenGLLibrary.cc
```diff
@@ -0,0 +1,85 @@
...
+        return OpenGL32Library::wglMakeCurrent(hDc, newContext);
+    }
+
+    HGLRC WINAPI wglGetCurrentContext() {
+        return OpenGL32Library::wglGetCurrentContext();
+    }
+
+    JNIEXPORT jboolean JNICALL Java_org_jetbrains_skiko_OpenGLLibraryKt_loadOpenGLLibraryWindows(JNIEnv *env, jobject obj) {
+        OpenGL32Library::load();
+        return OpenGL32Library::lib != 0;
```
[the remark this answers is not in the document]
igordmn: Done

skiko/src/awtMain/cpp/windows/OpenGLLibrary.cc
```diff
@@ -0,0 +1,85 @@
...
+    PROC_wglGetCurrentContext wglGetCurrentContext;
+
+    // Loads the opengl32.dll into memory
+    // If loading isn't successful, skip loading the next time
+    // Isn't thread safe, call it in synchronized block
+    void load() {
+        if (!isLoaded) {
+            isLoaded = true;
+            lib = LoadLibrary("opengl32.dll");
+            if (lib) {
```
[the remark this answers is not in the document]
igordmn: Done, now the error passed as RenderException (also, fixed a bug in it, and refactored)

skiko/src/awtMain/cpp/windows/OpenGLLibrary.cc
```diff
@@ -0,0 +1,85 @@
...
+    typedef void (*PROC_glFinish) (void);
+    typedef void (*PROC_glGetIntegerv) (GLenum pname, GLint *data);
+    typedef const GLubyte *(*PROC_glGetString) (GLenum name);
+    typedef HGLRC (WINAPI * PROC_wglCreateContext) (HDC hDc);
+    typedef BOOL (WINAPI * PROC_wglDeleteContext) (HGLRC oldContext);
+    typedef PROC (WINAPI * PROC_wglGetProcAddress) (LPCSTR lpszProc);
+    typedef BOOL (WINAPI * PROC_wglMakeCurrent) (HDC hDc, HGLRC newContext);
+    typedef HGLRC (WINAPI * PROC_wglGetCurrentContext) (void);
+
+    bool isLoaded = false;
```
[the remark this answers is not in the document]
igordmn: Done

skiko/src/awtMain/cpp/windows/OpenGLLibrary.cc
igordmn: The second looks good, done.

skiko/src/awtMain/cpp/windows/exceptions_handler.cc
```diff
@@ -47,19 +43,35 @@ const char *getDescription(DWORD code)
...
 }
 
-#endif
\ No newline at end of file
+
+void throwJavaRenderExceptionByErrorCode(JNIEnv *env, const char *function, DWORD code) {
+    char fullMsg[1024];
+    char *msg = 0;
+    FormatMessage(FORMAT_MESSAGE_ALLOCATE_BUFFER | FORMAT_MESSAGE_FROM_SYSTEM,
+        NULL, code, MAKELANGID(LANG_NEUTRAL, SUBLANG_DEFAULT), (LPTSTR) &msg, 0, NULL);
```
MatkovIvan, line 68:
```suggestion
        NULL, code, MAKELANGID(LANG_NEUTRAL, SUBLANG_DEFAULT), (LPTSTR) &msg, 1024, NULL);
```
You should pass the size to avoid stack overflow

MatkovIvan: Or even like that. `sizeof` should work for this case
```suggestion
        NULL, code, MAKELANGID(LANG_NEUTRAL, SUBLANG_DEFAULT), (LPTSTR) &msg, sizeof(msg), NULL);
```

igordmn: > sizeof(msg),

msg isn't allocated, we can't use it.

```
You should pass the size to avoid stack overflow
```
Why there will be a stack overflow?

[The doc](https://learn.microsoft.com/en-us/windows/win32/api/winbase/nf-winbase-formatmessage) says that it represents the minimal allocated size of a string:
```
If FORMAT_MESSAGE_ALLOCATE_BUFFER is set, this parameter specifies the minimum number of TCHARs to allocate for an output buffer.
```

MatkovIvan: > msg isn't allocated

it is.

> Why there will be a stack overflow?

*buffer overflow

MatkovIvan: Ah, `FORMAT_MESSAGE_ALLOCATE_BUFFER` flag. then you don't need to allocate the buffer in stack

MatkovIvan: nvm, it was `fullMsg`

MatkovIvan (approved):

K-o-d-y:
I'm using lwjgl which started crashing after this push

K-o-d-y:
https://github.com/JetBrains/compose-multiplatform/tree/master/experimental/lwjgl-integration

K-o-d-y:
@igordmn

PR #813, opened 2023-10-19
set OS name and architecture as archive classifier for JVM artifacts

kropp:
With OS name and architecture moved to classifier, it would be easier to reference these artifacts in some Gradle features, such as dependency verification and locking.

### DIFF
--- a/skiko/build.gradle.kts
+++ b/skiko/build.gradle.kts
@@ -1168,7 +1168,8 @@ fun skikoJvmRuntimeJarTask(
 ) = project.registerSkikoTask<Jar>("skikoJvmRuntimeJar", targetOs, targetArch) {
     dependsOn(awtJar)
     val target = targetId(targetOs, targetArch)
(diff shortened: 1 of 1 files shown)


skiko/build.gradle.kts
```diff
@@ -1168,7 +1168,8 @@ fun skikoJvmRuntimeJarTask(
 ) = project.registerSkikoTask<Jar>("skikoJvmRuntimeJar", targetOs, targetArch) {
     dependsOn(awtJar)
     val target = targetId(targetOs, targetArch)
-    archiveBaseName.set("skiko-$target")
+    archiveBaseName.set("skiko")
```
igordmn, line 1171: Looks good, but let's wait merging until we release Compose 1.5.10 (we need a hotfix for skiko)

kropp: Sure, no problem

igordmn (approved):

PR #816, opened 2023-10-23
Handle exceptions during SkiaSwingLayer rendering

Walingar:
Related to: https://github.com/JetBrains/jewel/issues/184 and https://github.com/JetBrains/jewel/issues/182

### DIFF
--- a/skiko/src/awtMain/kotlin/org/jetbrains/skiko/swing/SkiaSwingLayer.kt
+++ b/skiko/src/awtMain/kotlin/org/jetbrains/skiko/swing/SkiaSwingLayer.kt
@@ -115,6 +115,8 @@ open class SkiaSwingLayer(
                 redrawerManager.findNextWorkingRenderApi()
                 repaint()
             }
+        } catch (e: Exception) {
+            Logger.warn(e) { "Exception during SkiaSwingLayer rendering" }
         }
     }
 


m-sasha:
I'm usually against catching exceptions like that unless there's a very good reason (like maybe if you're controlling a nuclear reactor). It's better to even crash the process because then the exception is noticed and fixed faster. I only recommend catching exceptions if they're thrown by a 3rd-party black-box, which can't be fixed.

skiko/src/awtMain/kotlin/org/jetbrains/skiko/swing/SkiaSwingLayer.kt
```diff
@@ -115,6 +115,8 @@ open class SkiaSwingLayer(
                 redrawerManager.findNextWorkingRenderApi()
                 repaint()
             }
+        } catch (e: Exception) {
+            Logger.warn(e) { "Exception during SkiaSwingLayer rendering" }
```
Walingar, line 119: @m-sasha I agree, but I would say that this part of the framework is crucial, so not handling exceptions here would cause completely broken view (like in attached issue)

m-sasha: But why would there be exceptions here? Is it not better to discover and fix them instead?
In Compose we catch exceptions at the window level and typically show an error dialog followed by shutting down the application.

Walingar: They can appear only because of the broken internal contracts (out mistakes) and for sure they should be fixed. Like here https://github.com/JetBrains/compose-multiplatform-core/pull/878

Walingar: Then we should introduce this policy for swing bridge then too. Do you think it would be better?

m-sasha: I would recommend that, yes. I wonder what @igordmn thinks.

igordmn: We had similar code in the past (try-catch-log for rendering), and it was bad UX for end users (the window just stops to draw). We moved exception handling responsibility to `SkikoView`, as exceptions happen on its level, and it knows what to do with it.

In case of Swing components (which ComposePanel is), what is the policy for exceptions during rendering?

Regarding what policy we choose, we should implement it on the `ComposeBridge` level. Actually it is already implemented (see `ComposePanel.exceptionHandler`), we just need to define the default behavior.

m-sasha: > In case of Swing components (which ComposePanel is), what is the policy for exceptions during rendering?

Looks like it just calls the thread's uncaught exception handler (from `EventDispatchThread.pumpOneEventForFilters`).

igordmn: If that is the case, it seems everything already works this way?

m-sasha: Only by default (which I also think is a bad idea). In Swing I can set the thread's uncaught exception handler and change the behavior.

PR #821, opened 2023-11-01
don't crash an app if we got empty run during text shaping

SergeevPavel:


skiko/src/commonMain/cpp/common/include/TextLineRunHandler.hh
```diff
@@ -56,32 +56,34 @@ public:
 
     void commitRunBuffer(const RunInfo& info) override {
         TextLine::Run& run = fLine->fRuns.back();
-        int32_t glyph = 0;
-        int32_t graphemesInGlyph = 1;
-        SkScalar glyphLeft = run.fPos[glyph].fX;
-        run.fBreakOffsets.reserve(info.utf8Range.size() + 1);
-        run.fBreakPositions.reserve(info.utf8Range.size() + 1);
+        if (0 < info.glyphCount) {
```
MatkovIvan, line 59: Let's invert and early return to avoid extra indentation.

SergeevPavel: we also advance `fPosition` here

PR #822, opened 2023-11-07
In MetalRedrawer, stop delaying rendering as soon as the window becomes non-occluded.

m-sasha:
Occasionally, when the window is just getting show, `isOccluded` will return `true` while we're already trying to draw into it. This causes `MetalRedrawer.draw` to delay for 300ms in order to avoid wasting resources drawing while the window is occluded. But because this delay is not interrupted even when the window becomes non-occluded, we get a delay drawing the 2nd frame.

This PR adds a call from native code `NSWindowDidChangeOcclusionStateNotification` that interrupts the delay when the window becomes non-occluded, so that drawing can resume immediately.

### DIFF
--- a/skiko/src/awtMain/objectiveC/macos/MetalRedrawer.mm
+++ b/skiko/src/awtMain/objectiveC/macos/MetalRedrawer.mm
@@ -17,7 +17,7 @@
 
 #include <assert.h>
 
-#include "../../../jvmMain/cpp/common/interop.hh"
+#include "common/interop.hh"
 
 @implementation AWTMetalLayer
 
--- a/skiko/src/awtMain/objectiveC/macos/MetalRedrawer.mm
+++ b/skiko/src/awtMain/objectiveC/macos/MetalRedrawer.mm
@@ -15,6 +15,10 @@
 
 #import "MetalDevice.h"
 
+#include <assert.h>
+
+#include "../../../jvmMain/cpp/common/interop.hh"
+
 @implementation AWTMetalLayer
 
 - (id)init
@@ -50,6 +54,23 @@
 }
 
 @end
+
+/// Linked from skiko/src/jvmMain/cpp/common/impl/Library.cc
+/// clang treats extern symbol declarations as C in Objective-C++(.mm) and doesn't mangle them
+extern JavaVM *jvm;
+
+static JNIEnv *resolveJNIEnvForCurrentThread() {
+    JNIEnv *env;
+    int envStat = jvm->GetEnv((void **)&env, SKIKO_JNI_VERSION);
+
+    if (envStat == JNI_EDETACHED) {
+        jvm->AttachCurrentThread((void **) &env, NULL);
+    }
+
+    assert(env);
+
+    return env;
+}
 
 extern "C"
 {
@@ -92,11 +113,23 @@ JNIEXPORT jlong JNICALL Java_org_jetbrains_skiko_redrawer_MetalRedrawer_createMe
         /// max inflight command buffers count matches swapchain size to avoid overcommitment
         device.inflightSemaphore = dispatch_semaphore_create(device.layer.maximumDrawableCount);
 
-        if (transparency)
-        {
-            NSWindow* window = (__bridge NSWindow*) (void *) windowPtr;
+        NSWindow* window = (__bridge NSWindow*) (void *) windowPtr;
+
+        if (transparency) {
             window.hasShadow = NO;
         }
+
+        jclass redrawerClass = env->GetObjectClass(redrawer);
+        jmethodID onOcclusionStateChanged = env->GetMethodID(redrawerClass, "onOcclusionStateChanged", "(Z)V");
+
+        [[NSNotificationCenter defaultCenter] addObserverForName:NSWindowDidChangeOcclusionStateNotification
+                                                          object:window
+                                                           queue:[NSOperationQueue mainQueue]
+                                                      usingBlock:^(NSNotification * _Nonnull note) {
+            BOOL isOccluded = ([window occlusionState] & NSWindowOcclusionStateVisible) == 0;
+            JNIEnv *jniEnv = resolveJNIEnvForCurrentThread();
+            jniEnv->CallObjectMethod(layer.javaRef, onOcclusionStateChanged, isOccluded);
+        }];
 
         return (jlong) (__bridge_retained void *) device;
     }
@@ -173,14 +206,6 @@ JNIEXPORT void JNICALL Java_org_jetbrains_skiko_redrawer_MetalRedrawer_disposeDe
         env->DeleteGlobalRef(device.layer.javaRef);
         [device.layer removeFromSuperlayer];
         [CATransaction flush];
-    }
-}
-
-JNIEXPORT jboolean JNICALL Java_org_jetbrains_skiko_redrawer_MetalRedrawer_isOccluded(
-    JNIEnv *env, jobject redrawer, jlong windowPtr) {
-    @autoreleasepool {
-        NSWindow* window = (__bridge NSWindow*) (void *) windowPtr;
-        return ([window occlusionState] & NSWindowOcclusionStateVisible) == 0;
     }
 }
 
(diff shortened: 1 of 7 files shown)


skiko/src/awtMain/objectiveC/macos/MetalRedrawer.mm
```diff
@@ -92,12 +113,24 @@ JNIEXPORT jlong JNICALL Java_org_jetbrains_skiko_redrawer_MetalRedrawer_createMe
...
+        jmethodID onOcclusionStateChanged = env->GetMethodID(redrawerClass, "onOcclusionStateChanged", "(Z)V");
+
+        [[NSNotificationCenter defaultCenter] addObserverForName:NSWindowDidChangeOcclusionStateNotification
+                                                          object:window
+                                                           queue:[NSOperationQueue mainQueue]
+                                                      usingBlock:^(NSNotification * _Nonnull note) {
+            BOOL isOccluded = ([window occlusionState] & NSWindowOcclusionStateVisible) == 0;
+            JNIEnv *jniEnv = resolveJNIEnvForCurrentThread();
+            jniEnv->CallObjectMethod(layer.javaRef, onOcclusionStateChanged, isOccluded);
+        }];
```
elijah-semyonov, line 132: This method returns a handle, which must be used to deregister this observation once it's not needed.
https://developer.apple.com/documentation/foundation/nsnotificationcenter/1411723-addobserverforname

elijah-semyonov (changes requested):

skiko/src/awtMain/objectiveC/macos/MetalRedrawer.mm
m-sasha: Fixed (hopefully)

elijah-semyonov (approved):

skiko/src/awtMain/objectiveC/macos/MetalRedrawer.mm
```diff
@@ -92,12 +113,25 @@ JNIEXPORT jlong JNICALL Java_org_jetbrains_skiko_redrawer_MetalRedrawer_createMe
...
-        if (transparency)
-        {
-            NSWindow* window = (__bridge NSWindow*) (void *) windowPtr;
+        NSWindow* window = (__bridge NSWindow*) (void *) windowPtr;
+
+        if (transparency) {
             window.hasShadow = NO;
         }
 
+        jclass redrawerClass = env->GetObjectClass(redrawer);
```
igordmn, line 122:
```suggestion
        static jclass redrawerClass = env->GetObjectClass(redrawer);
```

skiko/src/awtMain/objectiveC/macos/MetalRedrawer.mm
```diff
@@ -92,12 +113,25 @@ JNIEXPORT jlong JNICALL Java_org_jetbrains_skiko_redrawer_MetalRedrawer_createMe
...
+        jmethodID onOcclusionStateChanged = env->GetMethodID(redrawerClass, "onOcclusionStateChanged", "(Z)V");
```
igordmn, line 123:
```suggestion
        static jmethodID onOcclusionStateChanged = env->GetMethodID(redrawerClass, "onOcclusionStateChanged", "(Z)V");
```

igordmn (approved):

skiko/src/awtMain/objectiveC/macos/MetalRedrawer.mm
m-sasha: Are you sure this is correct in Obj-C? Will `env->GetObjectClass(redrawer);` only be called the first time? No issues with multithreading?

Also, I'm reading that `jclass` instances (but not `jmethodID`) can get garbage collected... AI says we should convert it to a global reference with `env->NewGlobalRef`.

igordmn: Not sure, we need to check.

If there are issues, we need to measure performance, or extract this code to C, or to [interop.cc](https://github.com/JetBrains/skiko/blob/master/skiko/src/jvmMain/cpp/common/interop.cc). We always used static/global variables for this in the past, so we should have a reason for doing this differently.

igordmn: > env->NewGlobalRef

Sounds right

igordmn: > No issues with multithreading?

Usually this class is created in the Swing thread

m-sasha: This is called once per window, no? Doesn't sound like performance is critical.

I'll change it to a static initialized to null with an explicit check, so I don't have to store the jclass too.

skiko/src/awtMain/objectiveC/macos/MetalRedrawer.mm
```diff
@@ -15,6 +15,10 @@
 
 #import "MetalDevice.h"
 
+#include <assert.h>
+
+#include "../../../jvmMain/cpp/common/interop.hh"
```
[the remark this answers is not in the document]
m-sasha: If I add 
```
    includeHeadersNonRecursive(projectDir.resolve("src/jvmMain/cpp/common"))
```

then this import will be `#include "interop.hh"`, but we have several `interop.hh` files, so it'll be confusing I think.

I can add
```
includeHeadersNonRecursive(projectDir.resolve("src"))
```
and `#include "jvmMain/cpp/common/interop.hh"`. WDYT?

skiko/src/awtMain/objectiveC/macos/MetalRedrawer.mm
igordmn: > This is called once per window, no? Doesn't sound like performance is critical.

Yes, so it can affect window startup. And if we would do this everywhere, it may be noticible every time instead of just the first time. I don't have numbers, so we either stick to the established approach, or measure.

> initialized to null with an explicit check

Why? The static code looks fine to me in the C context, we only need to check Objective C.

igordmn: > Why?

Ah, to not store jclass. Missed this ). Ok

m-sasha: @igordmn updated to:
```
static jmethodID getOnOcclusionStateChangedMethodID(JNIEnv *env, jobject redrawer) {
    static jmethodID onOcclusionStateChanged = NULL;
    if (onOcclusionStateChanged == NULL) {
        jclass redrawerClass = env->GetObjectClass(redrawer);
        onOcclusionStateChanged = env->GetMethodID(redrawerClass, "onOcclusionStateChanged", "(Z)V");
    }
    return onOcclusionStateChanged;
}
```

m-sasha: > Yes, so it can affect window startup. And if we would do this everywhere, it may be noticible every time instead of just the first time. I don't have numbers, so we either stick to the established approach, or measure.

Tested out of curiosity:
```
        struct timespec tstart={0,0}, tend={0,0};
        clock_gettime(CLOCK_MONOTONIC, &tstart);
        jclass redrawerClass = env->GetObjectClass(redrawer);
        onOcclusionStateChanged = env->GetMethodID(redrawerClass, "onOcclusionStateChanged", "(Z)V");
        clock_gettime(CLOCK_MONOTONIC, &tend);
        printf("Getting onOcclusionStateChanged method ID took %d ns.\n", tend.tv_nsec - tstart.tv_nsec);
```

prints 
```
Getting onOcclusionStateChanged method ID took 4000 ns.
```

skiko/src/awtMain/objectiveC/macos/MetalRedrawer.mm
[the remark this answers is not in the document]
MatkovIvan: What about
```
    includeHeadersNonRecursive(projectDir.resolve("src/jvmMain/cpp"))
```
?

PR #827, opened 2023-11-10
Fix problem with conversion of Kotlin String to C char *

dima-avdeev-jb:
Previously it was a problem with converting  Kotlin `Array<String>` to C `char**`, because Kotlin String was conterted without Zero byte at the end.

### DIFF
--- a/skiko/src/nativeMain/kotlin/org/jetbrains/skia/impl/Native.native.kt
+++ b/skiko/src/nativeMain/kotlin/org/jetbrains/skia/impl/Native.native.kt
@@ -283,11 +283,12 @@ private external fun initCallbacks(
 )
 
 /**
- * C++ needs char* with zero byte at the end. So, we need to increase original string and write 0 at end.
+ * Converts String to zero-terminated utf-8 byte array.
  */
 private fun convertToZeroTerminatedString(string: String): ByteArray {
-    // encodeToByteArray encodes to utf8
-    val utf8 = string.encodeToByteArray()
+    //  C++ needs char* with zero byte at the end. So we need to copy array with an extra zero byte.
+
+    val utf8 = string.encodeToByteArray() // encodeToByteArray encodes to utf8
     // TODO Remove array copy, use `skString(data, length)` instead of `skString(data)`
     return utf8.copyOf(utf8.size + 1)
 }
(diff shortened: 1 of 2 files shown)


skiko/src/nativeMain/kotlin/org/jetbrains/skia/impl/Native.native.kt
```diff
@@ -284,3 +281,13 @@ private external fun initCallbacks(
     callVoid: COpaquePointer,
     dispose: COpaquePointer
 )
+
+/**
+ * C++ needs char* with zero byte at the end. So, we need to increase original string and write 0 at end.
```
pjBooms, line 286: I would remove this comment. The function name is self-evident now

igordmn (approved):

pjBooms (approved):

skiko/src/nativeMain/kotlin/org/jetbrains/skia/impl/Native.native.kt
igordmn: I would keep it, but inside the function. Without it, it is not obvious why we add +1

igordmn: zero-terminated string is a C term, not everyone knows it

igordmn: > I would keep it, but inside the function

Actually, it is fine outside as well

skiko/src/nativeMain/kotlin/org/jetbrains/skia/impl/Native.native.kt
```diff
@@ -284,3 +281,10 @@ private external fun initCallbacks(
     callVoid: COpaquePointer,
     dispose: COpaquePointer
 )
+
+private fun convertToZeroTerminatedString(string: String): ByteArray {
+    // encodeToByteArray encodes to utf8
+    val utf8 = string.encodeToByteArray()
+    // TODO Remove array copy, use `skString(data, length)` instead of `skString(data)`
```
MatkovIvan, line 288: Is there still no way to avoid double-copy in K/N?

dima-avdeev-jb: We can do it, with this TODO comment. But for now it is safe to stay double copy

skiko/src/nativeMain/kotlin/org/jetbrains/skia/impl/Native.native.kt
dima-avdeev-jb: Added comment back

pjBooms: Let's write outside comment as: 

/**
 * Converts String to zero-terminated utf-8 byte array.
*/

 and add the comment inside:

//  C++ needs char* with zero byte at the end. So we need to allocate array with an extra zero byte.

pjBooms (approved):

PR #828, opened 2023-11-14
Build Skiko with local Skia to mavenLocal

dima-avdeev-jb:
Added script
 - to clone and build Skia, with ability to change sources
 - publish Skiko with this local Skia to mavenLocal
 
By default iosSimulatorArm64 will be used


igordmn (approved):
Thanks!

Works on Windows without any issues.

The script itself also works on macOs for building iOS, but I can't completely sure about the end result, as I had a compilation error on "include" of std lib before this PR.

PR #833, opened 2023-11-24
Make tests pass with debug build

SergeevPavel:


SergeevPavel:
also, @igordmn what do you think about always running tests against debug build on CI?
Pros: it will catch more issues because of enabled asserts, and in general debug build may fail fast, where release may accidentally pass.
Cons: We will test different build, not that one that we actually using in prod

PR #835, opened 2023-11-28
Remove AwtFontManager and related code

rock3r:
Since https://github.com/JetBrains/compose-multiplatform-core/pull/918, we don't need this code anymore in Compose, which was its only user.

The still-relevant code was moved to Compose's desktop target.


rock3r:
As discussed with @igordmn :)

igordmn (approved):
LGTM, thanks!

PR #837, opened 2023-12-05
Transparency support for D3D swap chain

### LINKED ISSUE #327: Failed to create DirectX12 device when transparent = true
```kotlin
import androidx.compose.ui.window.Window
import androidx.compose.ui.window.application

fun main() = application {
    Window(::exitApplication, transparent = true, undecorated = true) {

    }
}
```
If transparent set to true, the console will prints out.
```
Failed to create DirectX12 device.
```

Info:
Windows 11
NVIDIA GeForce RTX 2070 with Max-Q Design
DirectX version: 12(Fl 12.1)

MatkovIvan:
Required for https://github.com/JetBrains/compose-multiplatform-core/pull/915

Fixes #327

### DIFF
--- a/skiko/src/awtMain/cpp/windows/directXRedrawer.cc
+++ b/skiko/src/awtMain/cpp/windows/directXRedrawer.cc
@@ -20,6 +20,9 @@
 #include <dxgi1_4.h>
 #include <dxgi1_6.h>
 
+#include <dcomp.h>
+#pragma comment(lib, "dcomp.lib")
+
 const int BuffersCount = 2;
 
 class DirectXDevice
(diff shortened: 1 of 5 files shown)


skiko/src/jvmMain/cpp/common/stubs.cc
```diff
@@ -59,6 +59,11 @@ JNIEXPORT void JNICALL Java_org_jetbrains_skiko_context_Direct3DContextHandler_f
     skikoUnimplemented("Java_org_jetbrains_skiko_context_Direct3DContextHandler_flush");
 }
 
+JNIEXPORT jlong JNICALL Java_org_jetbrains_skiko_redrawer_Direct3DRedrawer_chooseAdapter(JNIEnv *env, jobject redrawer, jint adapterPriority) {
```
eymar, line 62: Is it called anywhere? Why do we need this new function without an implementation?

MatkovIvan: It should not be called, but we should have all native JNI references declared - so it's like no-op implementation for Linux where we don't have DirectX. I just fixed the inconsistency here, but probably it makes sense to revisit this approach some day.

eymar: Got it

igordmn (changes requested):
It works on `SkiaAwtSample` if we set `window.background = Color(0, 0, 0, 0)`, but now we have an issue with input handling - click events aren't consumed by the window:

https://github.com/JetBrains/skiko/assets/5963351/7d5fe2e4-27a9-4e2c-a52a-ef5d81f2015d

We should fix event handling before merging it.

skiko/src/awtMain/cpp/windows/directXRedrawer.cc
```diff
@@ -53,24 +56,39 @@ class DirectXDevice
...
+        swapChainDesc.BufferCount = BuffersCount;
+        swapChainDesc.Scaling = DXGI_SCALING_STRETCH;
         swapChainDesc.SwapEffect = DXGI_SWAP_EFFECT_FLIP_DISCARD;
-        swapChainDesc.SampleDesc.Count = 1;
-        swapChainDesc.Scaling = DXGI_SCALING_NONE;
-        swapChainFactory4->CreateSwapChainForHwnd(queue.get(), window, &swapChainDesc, nullptr, nullptr, &swapChain1);
+        swapChainDesc.AlphaMode = DXGI_ALPHA_MODE_PREMULTIPLIED;
+        swapChainFactory4->CreateSwapChainForComposition(queue.get(), &swapChainDesc, nullptr, &swapChain1);
+
+        IDCompositionDevice* g_pdcDevice = nullptr;
```
igordmn, line 80:
```suggestion
        DCompositionCreateDevice(0, IID_PPV_ARGS(&g_pdcDevice));
        g_pdcDevice->CreateTargetForHwnd(window, true, &g_pdcTarget);
        g_pdcDevice->CreateVisual(&g_pdcVisual);
        g_pdcVisual->SetContent(swapChain1.get());
        g_pdcTarget->SetRoot(g_pdcVisual.get());
        g_pdcDevice->Commit();
```

and move objects to the fields:
```
    gr_cp<IDCompositionDevice> g_pdcDevice;
    gr_cp<IDCompositionTarget> g_pdcTarget;
    gr_cp<IDCompositionVisual> g_pdcVisual;
```

Otherwise, it seems we have a memory leak.

MatkovIvan:
> if we set `window.background = Color(0, 0, 0, 0)`, but now we have an issue with input handling - click events aren't consumed by the window

[As was always before, and for all other renderer API on Windows](https://github.com/JetBrains/compose-multiplatform-core/blame/976f6feab9f7539a02c264c8adc1a6b35eadecd1/compose/ui/ui/src/desktopMain/kotlin/androidx/compose/ui/awt/ComposeWindowDelegate.desktop.kt#L251-L253). I wrote comment with explanation here: [#915:  ComposeWindowDelegate.desktop.kt R112-R120](https://github.com/JetBrains/compose-multiplatform-core/pull/915/files#diff-8f2abdf764e8ac99c78379b60a6ecf78f3e57a2c08c2b2bcdfb82d0bf66d0513R112-R120)

```kt
/*
 * Windows makes clicks on transparent pixels fall through, but it doesn't work
 * with GPU accelerated rendering since this check requires having access to pixels from CPU.
```

My position here is to DROP `DwmEnableBlurBehindWindow` hack from other renderers like OpenGL to avoid setting window parameters by skiko view. But it's a separate store

skiko/src/awtMain/cpp/windows/directXRedrawer.cc
MatkovIvan: It's just auto-pointer-like objects that holding `nullptr` before setting. I don't see what moving the declaration changes.

igordmn (approved):

PR #843, opened 2023-12-12
Refactor interop with js in k/js and k/wasm source sets

eymar:
**Breaking changes:**
- `class CanvasRenderer` is now internal for k/js and k/wasm
- functions in `Convertors` are now internal for k/js and k/wasm
- top level `var onContentScaleChanged` was moved into SkiaLayer for k/js and k/wasm

**Description:**
We used to have as much common/shared code as possible, including js-interop code. But k/js and k/wasm stdlibs do not provide shared W3C types and APIs, despite the packages and some signatures match (it's intentional, and it won't change). 
Unfortunately, it looks like code duplication. But the underlying function and class symbols are not the same.

Therefore we should treat js-interop for k/js and k/wasm as indepedendent, not-commonizable code in most cases.

It was possible to extract some small parts of interop into W3CSubset. For other things, I moved the implementations into jsMain and wasmJsMain.


skiko/src/jsWasmMain/kotlin/org/jetbrains/skiko/w3c/W3CSubset.kt
```diff
@@ -0,0 +1,46 @@
...
+ * for use in both kotlin/js and kotlin/wasm code.
+ * Given that the Kotlin standard library does not provide common (shared between kotlin/js and kotlin/wasm) W3C API,
+ * this file serves to fill that void with custom partial declarations.
+ *
+ * These declarations do not cover all attributes and methods of the respective W3C types,
+ * but only those needed by our implementations.
+ */
+
+@Suppress("ClassName")
+internal external object window {
```
eymar, line 14: Intentionally `window`, not `Window`

skiko/build.gradle.kts
```diff
@@ -1663,14 +1663,6 @@ project.tasks.withType<org.jetbrains.kotlin.gradle.dsl.KotlinJsCompile>().config
     )
 }
 
-if (supportJs && supportWasm) {
-    project.tasks.whenTaskAdded {
-        if (name == "compileJsWasmMainKotlinMetadata") {
-            enabled = false
```
eymar, line 1669: This refactoring enables us to get rid of this hack.

samples/SkiaJsSample/src/jsWasmMain/kotlin/org/jetbrains/skiko/sample/js/App.jsWasm.kt
```diff
@@ -35,4 +33,9 @@ internal fun runApp() {
...
         skiaLayer.needRedraw()
     }
-}
\ No newline at end of file
+}
+
+
+// Note: despite the actual implementation looks identical, it can't be common - there is no W3C common API
+// return type is Any because k/js and k/wasm don't have a common type for HTML elements
+expect fun findHTMLCanvasElementById(id: String): Any
```
Schahen, line 41: two points
* Nowhere in this code we are relying on this returning actually HTMLCanvas - whatever check happen, if happen at all, happens no here
* Forcing non-nullability in one place is better that in many places

So my suggestion would be to just rename this to something more generic, like just findElementById, make that function return Any? 
```
actual fun findElementById(id: String): Any? = document.getElementById(id)
```
and then just `val canvas = findElementById("c$index")!!`

samples/SkiaMultiplatformSample/src/wasmJsMain/kotlin/org/jetbrains/skiko/sample/App.wasm.kt
```diff
@@ -1,5 +1,21 @@
...
+import org.jetbrains.skiko.SkiaLayer
+import org.w3c.dom.HTMLCanvasElement
+
 fun main() {
     runApp()
 }
+
+internal fun runApp() {
+    val skiaLayer = SkiaLayer()
+    skiaLayer.onContentScaleChanged = { scale -> println(scale) }
```
Schahen, line 14: is this intended?  I mean, are we leaving this for debug purposes?

skiko/src/jsMain/kotlin/org/jetbrains/skiko/Actuals.js.kt
```diff
@@ -1,8 +1,25 @@
...
 
-internal actual fun getEventTimestamp(e: UIEvent): Long {
+internal fun getEventTimestamp(e: UIEvent): Long {
     return e.timeStamp.toLong()
-}
\ No newline at end of file
+}
+
+internal actual fun CursorManager_setCursor(component: Any, cursor: Cursor) {
+    if (component is HTMLElement) {
```
Schahen, line 11: It's just an issue of taste and I don't have any strong view about this, I'm just say that the other option would be to just 

```
internal actual fun CursorManager_setCursor(component: Any, cursor: Cursor) {
    (component as? HTMLElement)?.style?.cursor = cursor
}

internal actual fun CursorManager_getCursor(component: Any): Cursor? {
    return (component as? HTMLElement)?.style?.cursor
}
```

again, I'm not even sure this has any benefits, just wanted to discuss

skiko/src/jsMain/kotlin/org/jetbrains/skiko/SkiaLayer.js.kt
```diff
@@ -0,0 +1,64 @@
...
+import org.w3c.dom.events.MouseEvent
+import org.w3c.dom.events.WheelEvent
+
+internal actual fun SkiaLayer.bindCanvasEventsToSkikoView(
+    canvas: HTMLCanvasElement
+) {
+    val htmlCanvas = canvas.unsafeCast<org.w3c.dom.HTMLCanvasElement>()
+
+    htmlCanvas.addEventListener("mousedown", { event ->
+        event as MouseEvent
```
Schahen, line 15: We can have helper extenstion function like Element.addMouseEventListener,  addKeyBoardEventListener
or event just generic version of it so that we won't have to case in every handler  (just scrolled down and realised that his was just moved so anyway any potential changes are not supposed to be in the framework of this particular PR, so nevermind)

samples/SkiaJsSample/build.gradle.kts
```diff
@@ -6,6 +6,8 @@ repositories {
     mavenLocal()
     mavenCentral()
     maven("https://maven.pkg.jetbrains.space/public/p/compose/dev")
+
+    // mavenLocal()
```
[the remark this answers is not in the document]
eymar: Right! I missed it.

samples/SkiaMultiplatformSample/src/wasmJsMain/kotlin/org/jetbrains/skiko/sample/App.wasm.kt
eymar: It's just for debug purpose. And it's not new code, it's copy-pasted from jsWasmMain common

samples/SkiaJsSample/build.gradle.kts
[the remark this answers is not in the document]
eymar: fixed

samples/SkiaJsSample/src/jsWasmMain/kotlin/org/jetbrains/skiko/sample/js/App.jsWasm.kt
eymar: fixed

skiko/src/jsMain/kotlin/org/jetbrains/skiko/Actuals.js.kt
eymar: fixed.

The benefit is that it's a shorter version :) but still readable.

igordmn (commented):
API review:

public -> internal changes of `toSkikoEvent` looks okay to me. 

By the way, we want to remove Skiko*Event's from Skiko in the future, and bind native events directly in Compose.

Schahen (approved):

PR #844, opened 2023-12-12
Add js exports generation

eymar:
This contribution was made by @ilgonmic ! Thank you Ilya!

**Description:**
To ensure that setup.mjs file contains correct exports, we better generate it than maintain manually.
It's done by a compiler plugin, which looks at WasmImport annotations and takes the name value from it.

P.S. our build.gradle.kts is quite big already (~1.8k lines). I tried to move new part to buildSrc, but I didn't manage to make work. I'll try to think about it separately. 

UPD: a separate PR to split the build.gradle.kts into smaller parts - https://github.com/JetBrains/skiko/pull/845


skiko/build.gradle.kts
```diff
@@ -376,7 +383,24 @@ kotlin {
...
+
+            main.compileTaskProvider.configure {
+                outputs.file(setupMjs)
+            }
+
+            test.compileTaskProvider.configure {
+                outputs.file(skikoTestMjs)
+            }
+
+            listOf(main, test).forEach {
```
Schahen, line 397: Do we need this piece at all?

skiko/build.gradle.kts
```diff
@@ -1679,4 +1710,45 @@ tasks.withType<org.jetbrains.kotlin.gradle.tasks.KotlinNativeCompile>().configur
     // https://youtrack.jetbrains.com/issue/KT-56583
     compilerOptions.freeCompilerArgs.add("-XXLanguage:+ImplicitSignedToUnsignedIntegerConversion")
     kotlinOptions.freeCompilerArgs += "-Xopt-in=kotlinx.cinterop.ExperimentalForeignApi"
-}
\ No newline at end of file
+}
+
+abstract class AbstractImportGeneratorCompilerPluginSupportPlugin(
```
Schahen, line 1715: it seems it might sense to keep this in buildSrc

Schahen (approved):

skiko/build.gradle.kts
eymar: I rebased on master, therefore this comment doesn't link anymore to the code.
This part is indeed needed:
`KotlinCompilerPluginSupportPlugin` has a method `fun getPluginArtifact()` which we override.  We return `SubpluginArtifact(SkikoArtifacts.groupId, IMPORT_GENERATOR)` with null version. Therefore by default it tries to get the same version as koltin version. 

Since our compiler plugin is not published, we need to substitute this implicit dependency.

skiko/build.gradle.kts
eymar: Moved to buildSrc.

skiko/build.gradle.kts
eymar: Also I moved it to buildSrc

PR #857, opened 2024-01-17
Populate currentSystemTheme for browsers

eygraber:


eymar:
We aim to keep skiko responsible only for drawing. We didn't follow this before, but now we're going to move non-drawing platform specific integrations to Compose (at least for web and ios). 

Was your intention to make it work in Compose (https://github.com/JetBrains/compose-multiplatform-core/pull/998)?
Or do you need it in skiko too for some reason?

eygraber:
I don't personally need it, just figured I'd add missing functionality while I was doing it in compose.

PR #858, opened 2024-01-19
Don't try to init D3D with zero size

MatkovIvan:
Currently it fails to init swap chain because the size of the element is zero. It's not valid parameter in `DXGI_SWAP_CHAIN_DESC1`

`[SKIKO] warn: Exception in draw scope org.jetbrains.skiko.RenderException: Native exception in [Java_org_jetbrains_skiko_redrawer_Direct3DRedrawer_initSwapChain], code 3221225477: EXCEPTION_ACCESS_VIOLATION`

Fixes https://github.com/JetBrains/compose-multiplatform/issues/4105

### DIFF
--- a/skiko/src/awtMain/kotlin/org/jetbrains/skiko/redrawer/Direct3DRedrawer.kt
+++ b/skiko/src/awtMain/kotlin/org/jetbrains/skiko/redrawer/Direct3DRedrawer.kt
@@ -107,7 +107,7 @@ internal class Direct3DRedrawer(
     fun resizeBuffers(width: Int, height: Int) = resizeBuffers(device, width, height)
 
     fun getBufferIndex() = getBufferIndex(device)
-    fun initSwapChain() = initSwapChain(device)
+    fun initSwapChain(width: Int, height: Int) = initSwapChain(device, width, height)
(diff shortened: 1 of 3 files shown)


skiko/src/awtMain/kotlin/org/jetbrains/skiko/redrawer/Direct3DRedrawer.kt
```diff
@@ -84,10 +84,14 @@ internal class Direct3DRedrawer(
     }
 
     private fun drawAndSwap(withVsync: Boolean) = synchronized(drawLock) {
-        if (!isDisposed) {
-            contextHandler.draw()
-            swap(device, withVsync)
+        if (isDisposed) {
+            return
         }
+        if (layer.width <= 0 || layer.height <= 0) {
```
igordmn, line 90: `layer.width` is AWT-manipulated property, and we can have a race condition when accessing it from another thread (`dispatcherToBlockOn`).

Better to move it outside of `withContext(dispatcherToBlockOn) {`. Or remember the local size inside `resizeBuffers`/`makeDirectXSurface` (I vote for it)

igordmn (approved):

PR #864, opened 2024-02-05
Remove js/wasm events handling from skiko (almost) completely

Schahen:
This is the part of COMPOSE-891 task - the idea is to have this in compose - skiko should be a place for skia wrappers and nothing more


eymar:
There is also Convertors.wasm.kt - Can be removed as well, I guess.

eymar (approved):

PR #866, opened 2024-02-07
Call `Library.load()` before calling native getCurrentSystemTheme()

m-sasha:
Fixes: https://github.com/JetBrains/compose-multiplatform/issues/4208

### DIFF
--- a/skiko/src/awtMain/kotlin/org/jetbrains/skiko/SystemTheme.awt.kt
+++ b/skiko/src/awtMain/kotlin/org/jetbrains/skiko/SystemTheme.awt.kt
@@ -1,10 +1,13 @@
 package org.jetbrains.skiko
 
 actual val currentSystemTheme: SystemTheme
-    get() = when (getCurrentSystemTheme()) {
-        0 -> SystemTheme.LIGHT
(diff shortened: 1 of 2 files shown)


eymar (approved):

skiko/src/awtMain/kotlin/org/jetbrains/skiko/SystemTheme.awt.kt
```diff
@@ -1,10 +1,13 @@
 package org.jetbrains.skiko
 
 actual val currentSystemTheme: SystemTheme
-    get() = when (getCurrentSystemTheme()) {
-        0 -> SystemTheme.LIGHT
-        1 -> SystemTheme.DARK
-        else -> SystemTheme.UNKNOWN
+    get() {
+        Library.load()
```
MatkovIvan, line 5: It shouldn't be called on each call

m-sasha: I wouldn't say "shouldn't" (`Library.load` protects against being called multiple times), but yes, it's slightly faster not to.

PR #869, opened 2024-02-09
Fix a native crash on `makeGL`

igordmn:
It is a regression after https://github.com/JetBrains/skiko/pull/811

After that PR, OpenGL functions are linked in runtime via function `loadOpenGLLibrary`. We didn't call it in `makeGL` functions. Now we call.

- load library before calling `makeGL`
- add `isLoaded` to avoid loading it multiple times
- throw RenderException on macOs/iOS, as OpenGL there is deprecated

Fixes https://github.com/JetBrains/compose-multiplatform/issues/3972


skiko/src/commonMain/kotlin/org/jetbrains/skia/BackendRenderTarget.kt
```diff
@@ -16,16 +18,10 @@ class BackendRenderTarget internal constructor(ptr: NativePointer) : Managed(ptr
...
-                _nMakeGL(
-                    width,
-                    height,
-                    sampleCnt,
-                    stencilBits,
-                    fbId,
-                    fbFormat
-                )
-            )
+            loadOpenGLLibrary()
```
elijah-semyonov, line 21: Do we ever create `BackendRenderTarget` before initialising context? From a first look this one seems redundant.

elijah-semyonov (approved):

skiko/src/commonMain/kotlin/org/jetbrains/skia/BackendRenderTarget.kt
igordmn: Seems it isn't needed indeed. But not because we init context. Because we can create it without context and even without OpenGL. Checked it just now.

eymar:
This change affected CfW (k/js) running in Safari :) https://kotlinlang.slack.com/archives/C01F2HV7868/p1708540690334899?thread_ts=1708527218.429189&cid=C01F2HV7868 
I'll prepare a fix.

PR #870, opened 2024-02-09
Nan vals in textstyle

SergeevPavel:


igordmn (approved):

skiko/buildSrc/src/main/kotlin/tasks/configuration/CommonTasksConfiguration.kt
```diff
@@ -134,4 +142,46 @@ fun Project.configureSignAndPublishDependencies() {
             }
         }
     }
+}
+
+fun KotlinTarget.generateVersion(
```
igordmn, line 147: Why this function is moved?

SergeevPavel: it's used in all targets but by some reason during the build.gradle.kts split it migrated to NativeTasks

skiko/src/jvmMain/cpp/common/impl/Managed.cc
```diff
@@ -1,4 +1,5 @@
 #include <iostream>
+#include <cstdint>
```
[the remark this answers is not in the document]
SergeevPavel: `uintptr_t` is defined in `<cstdint>` https://cplusplus.com/reference/cstdint/

igordmn (approved):

PR #874, opened 2024-02-20
Remove SkiaJsSample completely

Schahen:
SkiaMultiplatformSample covers any possible scenario one can be interested in, including actually the js-only solution as well


Schahen:
We've decided to preserve bouncing balls in the mpp demo so I'm closing this one so far

PR #879, opened 2024-02-26
Use coroutines 1.8.0 for all targets

eymar:
According to COMPOSE-939


skiko/build.gradle.kts
```diff
@@ -212,15 +211,15 @@ kotlin {
                 val jsWasmMain by creating {
                     dependsOn(nativeJsMain)
                     dependencies {
```
igordmn, line 213: We can just remove the `dependencies` block

igordmn (approved):

PR #889, opened 2024-03-11
Expose `BBHFactory` to record bounds in `Picture`

MatkovIvan:


m-sasha:
Is it doing this always now? I don't have any specific knowledge, but it could slow drawing...

See https://github.com/JetBrains/compose-multiplatform-core/pull/1090 and https://issues.skia.org/issues/324465764
I considered using BBHFactory but decided against it just in case it worsens performance.

skiko/src/commonMain/kotlin/org/jetbrains/skia/PictureRecorder.kt
```diff
@@ -44,9 +45,10 @@ class PictureRecorder internal constructor(ptr: NativePointer) : Managed(ptr, _F
      *
      * @param bounds the cull rect used when recording this picture. Any drawing the falls outside
      * of this rect is undefined, and may be drawn or it may not.
+     * @param bbh optional acceleration structure
```
igordmn, line 48: It doesn't sound descriptive. Where it from?

igordmn (approved):

skiko/src/commonMain/kotlin/org/jetbrains/skia/PictureRecorder.kt
MatkovIvan: https://github.com/JetBrains/skia/blob/6726cd3e1ac4f30d6874c278008cdf24dfc9b470/include/core/SkPictureRecorder.h#L41C16-L41C27

MatkovIvan:
> Is it doing this always now?

no, it just exposes a parameter. `null` (as before) by default

m-sasha:
I'd recommend checking whether it affects performance when there are many objects drawn, before using it for drawing arbitrary user-provided objects in Compose.

If it doesn't affect performance, we should use it instead of my hack with the large picture size.

m-sasha:
Had another thought - is it possible to avoid managing and passing the BBH factory in/out of the JVM? We only need it there to know when to delete it.

Maybe an API like a `drawPictureWithBBH(block: (SkCanvas) -> Unit)` with native code being responsible for both allocating and deleting the BBH factory?

PR #897, opened 2024-03-24
Follow OS-specific paths for installed binaries and logs instead of `~/.skiko`

### LINKED ISSUE #707: Follow XDG Base Directory
XDG Base Directory specification defines where these files should be looked for by defining one or more base directories relative to which files should be located.
https://specifications.freedesktop.org/basedir-spec/basedir-spec-latest.html

As for now, applications that use skiko as dependency create .skiko folder in user $HOME directory.

62832:
Currently, Skiko always creates the `$HOME/.skiko` directory when used as a dependency by other projects, which can only be overridden by the `skiko.data.path` property. This PR does away with this directory and makes Skiko more compliant with OS-specific standards for stored program data and logs.

- On Linux-based systems (including Android), Skiko will follow the [XDG Base Directory Specification](https://specifications.freedesktop.org/basedir-spec/basedir-spec-latest.html) and store installed binaries under `$XDG_DATA_HOME/skiko` and logs under `$XDG_STATE_HOME/skiko`, with the appropriate fallbacks if the former variables are unset. As a result, this PR closes #707.
- On Windows, installed binaries and logs will be stored under `%LOCALAPPDATA%/Skiko`.
- On macOS and iOS, installed binaries will be stored under `~/Library/Application Support/Skiko` and logs under `~/Library/Logs`.


62832:
It came to my attention at some point that the choice of following XDG paths for *Android* is actually incorrect. Android apparently has its own storage scheme for such data somewhere under the `/data` directory, but it's unclear to me what this should be for Skiko.

PR #915, opened 2024-04-16
Option to not throw RenderException when use OpenGL on macOS

igordmn:
To allow using OpenGL on macOs, call:
```
System.setProperty("skiko.macos.opengl.enabled", "true")
```

By request from user in DM who uses LWJGL+Skiko on macOS.


elijah-semyonov (approved):

igordmn:
The PR is changed.

Now we just add a flag to reenable OpenGL.

Enabling OpenGL by default isn't trivial, because we need to throw `RenderException` if OpenGL isn't available (for cases when user knows how to handle it, for example - switch to Software rendering)

For iOS there weren't requests from users, and it seems it didn't work, as we didn't link the OpenGL framework before.

elijah-semyonov (approved):

PR #916, opened 2024-04-21
Add isMaximized and setMaximized methods to PlatformOperations

MohamedRejeb:
Related to: https://github.com/JetBrains/compose-multiplatform-core/pull/1148

Hi,
I added `setMaximized` methods to `PlatformOperations` which is using a native implementation for macos.
I added `maximized` property in HardwareLayer and SkiaLayer.awt.
This will fix a problem that we have on Compose Desktop when we try to do something like this: floating -> fullscreen -> maximized, the window will stay fullscreen, more details in the PR above.

The swing implementation on Windows and Linux is not tested yet.

The current implementation is not final, but it's working, so let me know what you think.

### DIFF
--- a/skiko/src/awtMain/kotlin/org/jetbrains/skiko/HardwareLayer.kt
+++ b/skiko/src/awtMain/kotlin/org/jetbrains/skiko/HardwareLayer.kt
@@ -53,6 +53,10 @@ internal open class HardwareLayer(
         get() = platformOperations.isFullscreen(this)
         set(value) = platformOperations.setFullscreen(this, value)
 
+    var maximized: Boolean
+        get() = platformOperations.isMaximized(this)
+        set(value) = platformOperations.setMaximized(this, value)
+
     fun disableTitleBar(customHeaderHeight: Float) {
         platformOperations.disableTitleBar(this, customHeaderHeight)
     }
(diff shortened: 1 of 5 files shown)


igordmn:
Thanks for the PR! 

Sorry for the delay, I plan to look at it more closely later. There are potential race conditions with the current code, I still not sure if they are relevant or not, I need to look at it more carefully.

MohamedRejeb:
Any updates on this?

igordmn:
Sorry, no update

igordmn:
A preliminary review:

1. The most obvious race condition is in `performSelectorOnMainThread`. Between calling `setMaximized` and performing the scheduled task, other operations can appear (checking `isMaximized`, or changing `isFullscreen`)

2. There is existing race condition inside `osxSetFullscreenNative` (because of the same `performSelectorOnMainThread`). It worked not well so far. If we introduce an additional logic, we'll have more bugs.

3. I am not sure about `toggling` instead of `setting`. Multiple scheduled `toggle`'s will result in an undesired state.

4. We have 2 different operation doing the same, but slightly differently (the new `toggleFullScreenAndZoom` and the old `makeFullscreen`). It can lead to an inconsistent state.

5. We should have this contract:
```
layer.isFullscreen = true
println(layer.isFullscreen) // should be true immediately
println(layer.isMaximized ) // should be false immediately

layer.isMaximized = true
println(layer.isMaximized ) // should be true immediately
println(layer.isFullscreen) // should be false immediately
```
But the current logic may not guarantee that. Also I am not sure if the native functions also can guarantee it. If not, we should write workarounds.

kropp:
I'm closing this PR for now as it has unresolved reviewer comments and conflicts with the main branch.

PR #920, opened 2024-05-02
Support accessibility in SkiaSwingLayer

m-sasha:
To properly support accessibility, SkiaSwingLayer needs to subclass `JPanel`, not `JComponent`.
Also, it needs to implement `requestNativeFocusOnAccessible`.

Fixes https://github.com/JetBrains/compose-multiplatform/issues/4731

Note that this doesn't yet fully fix a11y in `ComposePanel`. It still doesn't work well when Swing and Compose are mixed. `ComposePanel` appears to be highlighted as a whole, rather than individual widgets in it, by VoiceOver. It is, however, possible to go "inside" by focus traversal.

### DIFF
--- a/skiko/src/awtMain/kotlin/org/jetbrains/skiko/Accessibility.kt
+++ b/skiko/src/awtMain/kotlin/org/jetbrains/skiko/Accessibility.kt
@@ -1,8 +1,69 @@
 package org.jetbrains.skiko
 
+import kotlinx.coroutines.*
+import java.awt.Component
+import java.awt.KeyboardFocusManager
+import java.awt.event.FocusEvent
+import java.beans.PropertyChangeEvent
 import javax.accessibility.Accessible
+import javax.accessibility.AccessibleContext
 
 /**
  * See [nativeInitializeAccessible] doc for details
  */
 internal external fun initializeCAccessible(accessible: Accessible)
+
+/**
+ * A helper class for implementing requesting accessibility focus on a given accessible.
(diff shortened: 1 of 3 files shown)


igordmn:
> ComposePanel appears to be highlighted as a whole

In the past I fixed this for the default `ComposePanel`. I.e. When we have focus on a swing component and press Tab - it doesn't focus the ComposePanel, it focuses the Compose component inside. See `ComposeFocusTest`.

Do we need to do the same for `System.setProperty("compose.layers.type", "COMPONENT")`?

igordmn:
> ComposePanel appears to be highlighted as a whole

Or it is accessibility-only issue, and focus is okay?

m-sasha:
> > ComposePanel appears to be highlighted as a whole
> 
> Or is it accessibility-only issue, and focus is okay?

It's an accessibility-only issue:

https://github.com/JetBrains/skiko/assets/5230206/0b59c81d-f84e-4778-a34f-b52057076849

igordmn (approved):

igordmn:
> Fixes https://github.com/JetBrains/compose-multiplatform/issues/4731
> It is, however, possible to go "inside" by focus traversal.

I tested it on Windows. If I used it correctly, it seems doesn't work. The panel just skipped during traversal, and it isn't possible to go inside it.

rock3r:
Is the issue really closed if the solution doesn't work on Windows?

PR #923, opened 2024-05-14
Redraw immediately on panel size change

MatkovIvan:
Fixes a couple of issues like: https://github.com/JetBrains/compose-multiplatform/issues/4744 (TODO: collect all references)

Before (slow-mo)

https://github.com/JetBrains/skiko/assets/1836384/5768b139-d08f-4142-8c14-2969107807f9

After (slow-mo)

https://github.com/JetBrains/skiko/assets/1836384/cc9a5103-3d30-412e-a5a4-baf590d657b8

### DIFF
--- a/skiko/src/awtMain/kotlin/org/jetbrains/skiko/SkiaLayer.awt.kt
+++ b/skiko/src/awtMain/kotlin/org/jetbrains/skiko/SkiaLayer.awt.kt
@@ -310,18 +310,7 @@ actual open class SkiaLayer internal constructor(
     override fun paint(g: java.awt.Graphics) {
         Logger.debug { "Paint called on: $this" }
         checkContentScale()
-
-        // `paint` can be called when we already inside `draw` method.
-        //
-        // For example if we call some AWT function inside renderer.onRender,
-        // such as `jframe.isEnabled = false` on Linux
-        //
-        // To avoid recursive call of `draw` (we don't support recursive calls) we just schedule redrawing.
-        if (isRendering) {
-            redrawer?.needRedraw()
-        } else {
-            redrawer?.redrawImmediately()
-        }
+        tryRedrawImmediately()
     }
 
     override fun setBounds(x: Int, y: Int, width: Int, height: Int) {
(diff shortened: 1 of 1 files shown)


skiko/src/awtMain/kotlin/org/jetbrains/skiko/SkiaLayer.awt.kt
```diff
@@ -326,6 +324,16 @@ actual open class SkiaLayer internal constructor(
...
     }
 
+    override fun setBounds(x: Int, y: Int, width: Int, height: Int) {
+        super.setBounds(x, y, width, height)
+
+        // To avoid visual artifacts (especially on Windows/DirectX),
+        // redrawing should be performed immediately, without scheduling to "later".
+        // Subscribing to events instead of overriding this method won't help too.
+        redrawer?.syncSize()
+        redrawer?.redrawImmediately()
```
igordmn, line 334:
```suggestion
        if (isRendering) {
            redrawer?.needRedraw()
        } else {
            redrawer?.redrawImmediately()
        }
```

in case someone will call `setBounds` in a composition. We had reports in the past, that is why we have this in `paint`.

igordmn, line 334: Do this reproduces because we changed hirerachy in Compose? We have `paint` called before for this case.

MatkovIvan: The reproduction/video is pure skiko, without Compose. Seems `paint` called later/not always

skiko/src/awtMain/kotlin/org/jetbrains/skiko/SkiaLayer.awt.kt
```diff
@@ -312,13 +310,27 @@ actual open class SkiaLayer internal constructor(
     override fun paint(g: java.awt.Graphics) {
         Logger.debug { "Paint called on: $this" }
         checkContentScale()
+        tryRedrawImmediately()
+    }
+
+    override fun setBounds(x: Int, y: Int, width: Int, height: Int) {
```
igordmn, line 316: org.jetbrains.skiko.SkiaLayerTest[awt] > resize window[awt] FAILED

fails on Windows

MatkovIvan: Yep, I saw this. Regardless of CI state, please check the fix locally to make sure that we agree that the change makes sense

igordmn: The fix is okay, though I see "jumps" on resize. But it is better than the current behaviour.

https://github.com/JetBrains/skiko/assets/5963351/196783f3-b1d0-484e-8c92-21c81bfd5f40

Regarding the failed test - it fails locally as well, and it is the only failed test.

MatkovIvan: The problem was with software renderer - it just stops to display updates after try to redraw during layout despite any render/flush/... calls after that. Spend some time on it, but the reason isn't clear.

Added a check to apply my fix only to Direct3D to not just fix this issue but also reduce the scope of this change

MatkovIvan: Help with what? The issue was only in Windows/DirectX

igordmn (approved):

elijah-semyonov:
This PR breaks compose-core tests:
Screenshot 2024-06-13 at 14 29 44

PR #924, opened 2024-05-14
[WIP] Add MingwX64 (Windows) support

xolider:
This PR adds support for Windows (MingwX64).
It uses LLVM (Clang) for compiling Skia rather than MSVC. To successfully build the mingwX64 target, you need to have LLVM installed and the "bin" folder in your PATH.

This is a work in progress. At this writing time, it successfully builds native bridge and link it to a static `.lib` library

If you have any comments or ideas to make it stable, please feel free to leave them here :)

I know there is an other opened PR for this goal, but it seems to be out-of-date and not maintained anymore

**UPDATE**: the `compileKotlinMingwX64 `task now completes successfully

### DIFF
--- a/DEVELOPMENT.md
+++ b/DEVELOPMENT.md
@@ -7,15 +7,10 @@
     sudo apt-get install ninja-build fontconfig libfontconfig1-dev libglu1-mesa-dev libxrandr-dev libdbus-1-dev zip libx11-dev
     ```
   * `Windows`
-    1. Download [Visual Studio Build Tools 2019](https://learn.microsoft.com/en-us/visualstudio/releases/2019/history) (search "BuildTools" on the page).
-    2. During the installation, select "Desktop development with C++"
-    3. Add an environment variable SKIKO_VSBT_PATH=C:/Program Files (x86)/Microsoft Visual Studio/2019/BuildTools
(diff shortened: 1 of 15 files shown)


xolider:
To successfully build the mingwX64 target, you need to have LLVM installed and the "bin" folder in your PATH.

xolider:
**UPDATE**: the `compileKotlinMingwX64` task now completes successfully

xolider:
**To do:**

- [ ] Implement the Skia layer
- [ ] Implement the clipboard logic
- [ ] Implement the cursor selector

Syrou:
Looking forward to see this added

cvicartadeo:
> Looking forward to see this added

This error may due to the Skia's compilation method. May be the LLVM solution is not the right one. I would try to compile with MSYS2 Mingw64, but I currently run out of time with my current job.

xolider:
> Looking forward to see this added

This error may due to the Skia's compilation method. Maybe the LLVM solution is not the right one. I would try to compile with MSYS2 Mingw64 to get rid of MSVC, but that means no access to Win32 API. I currently run out of time with my current job to check that

An another possible thing is that the Konan LLVM version is too old. I also tried to compile for linuxX64 and had issues with the provided libc library that was too old comparing my host's one, and lead to incompatibilities issues. As far is I heard, Jetbrains plan to bump LLVM/GCC versions with the next Kotlin 2 compiler

PR #927, opened 2024-05-17
Turn off vsync throttling when macOS hardware sleeps

elijah-semyonov:
Speculative fix for [a freeze issue](https://youtrack.jetbrains.com/issue/COMPOSE-1441/Desktop-skiko-freeze-after-sleep-on-Mac-with-multiple-external-monitors).

### DIFF
--- a/skiko/src/awtMain/objectiveC/macos/DisplayLinkThrottler.mm
+++ b/skiko/src/awtMain/objectiveC/macos/DisplayLinkThrottler.mm
@@ -7,9 +7,6 @@
 #import <stdatomic.h>
 
 @interface DisplayLinkThrottler : NSObject
-
-- (void)onVSync;
-
 @end
 
 static CVReturn displayLinkCallback(CVDisplayLinkRef displayLink, const CVTimeStamp *now, const CVTimeStamp *outputTime, CVOptionFlags flagsIn, CVOptionFlags *flagsOut, void *displayLinkContext) {
@@ -24,6 +21,7 @@ static CVReturn displayLinkCallback(CVDisplayLinkRef displayLink, const CVTimeSt
     NSScreen *_displayLinkScreen;
     CVDisplayLinkRef _displayLink;
     NSConditionLock *_vsyncConditionLock;
+    BOOL _isSleeping;
 }
 
 - (instancetype)init {
@@ -33,9 +31,32 @@ static CVReturn displayLinkCallback(CVDisplayLinkRef displayLink, const CVTimeSt
         _displayLinkScreen = nil;
         _displayLink = nil;
         _vsyncConditionLock = [[NSConditionLock alloc] initWithCondition: 1];
+        _isSleeping = NO;
+
+        NSNotificationCenter *notificationCenter = [[NSWorkspace sharedWorkspace] notificationCenter];
+
+        [notificationCenter addObserver:self
+                            selector:@selector(systemWillSleep:)
+                            name:NSWorkspaceWillSleepNotification
+                            object:nil];
+
+        [notificationCenter addObserver:self
+                            selector:@selector(systemDidWake:)
+                            name:NSWorkspaceDidWakeNotification
+                            object:nil];
     }
 
     return self;
+}
+
+- (void)systemWillSleep:(NSNotification *)notification {
+    _isSleeping = YES;
+    [self invalidateDisplayLink];
+    [self onVSync];
+}
+
+- (void)systemDidWake:(NSNotification *)notification {
+    _isSleeping = NO;
 }
 
 - (void)onVSync {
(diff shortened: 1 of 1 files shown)


SergeevPavel (approved):

skiko/src/awtMain/objectiveC/macos/DisplayLinkThrottler.mm
```diff
@@ -33,19 +34,42 @@ - (instancetype)init {
...
     return self;
 }
 
+- (void)systemWillSleep:(NSNotification *)notification {
+    _isSleeping = YES;
+    [self invalidateDisplayLink];
+    [self onVSync];
+}
+
+- (void)systemDidWake:(NSNotification *)notification {
```
igordmn, line 61: When the system wakes up, we need to notify the window to redraw the content. If system calls the java function `paint()`, it will redraw it. But can we guarantee that? If not, we should send a signal to AWT here.

igordmn (commented):
Thanks!

I left important comments need to be discussed.

skiko/src/awtMain/objectiveC/macos/DisplayLinkThrottler.mm
igordmn: It seems the fix isn't about disabling redrawing, but about disabling vsync. In this case, no need to notify. But will we have unlimited FPS, and should stop drawing in case of sleeping?

skiko/src/awtMain/objectiveC/macos/DisplayLinkThrottler.mm
```diff
@@ -65,6 +89,11 @@ - (void)invalidateDisplayLink {
 }
 
 - (void)setupDisplayLinkForWindow:(NSWindow *)window {
+    if (_isSleeping) {
+        /// System is sleeping, don't setup display link
```
igordmn, line 93: Should we nullify the display link?

skiko/src/awtMain/objectiveC/macos/DisplayLinkThrottler.mm
```diff
@@ -33,19 +34,42 @@ - (instancetype)init {
         _displayLinkScreen = nil;
         _displayLink = nil;
         _vsyncConditionLock = [[NSConditionLock alloc] initWithCondition: 1];
+        _isSleeping = NO;
+
+        NSNotificationCenter *notificationCenter = [[NSWorkspace sharedWorkspace] notificationCenter];
+
+        [notificationCenter addObserver:self
```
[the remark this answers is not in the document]
elijah-semyonov: No, this API is target-selector and doesn't require explicit observer removal (will be done automatically in dealloc)

skiko/src/awtMain/objectiveC/macos/DisplayLinkThrottler.mm
elijah-semyonov: No, it's `nil` at this point, because when _isSleeping is set to true, display link is also deinitialised.

skiko/src/awtMain/objectiveC/macos/DisplayLinkThrottler.mm
elijah-semyonov: The issue is about deadlock on the conditional variable encpsulating vsync state. I can't reproduce that so my speculation is to destroy this mechanism temporarily while the device is sleeping.
I have an idea  that have some really funny race-condition, where the logical screen for the window changes mid-flight due to some OS-level reconfigurations (flickering when connecting a display/waking up), so we wait on vsync from the screen which no longer exists/and AppKit didn't gracefully changed the screen to the existing one, but I have no idea currently how to fix _that_ so I'd better try this solution and see if it fixes the issue.

elijah-semyonov: Potentially, it can indeed cause discarded drawables (because of redundant redraws, the initial issue that vsync mechanism fixed) if multiple were presented during a single hardware vsync after the the sleep notification, but that's an edge case that I bet will not affect the users

SergeevPavel:
I've noticed one more strange thing, not sure that it's related with this changes, but anyway. Usually we call `performDraw()` and waiting for vsync on our special blocking dispatcher. But in the problematic stack traces I see that we came from `JComponent#paint` and through `redrawImmediately` we call `performDraw` and block EventDispatchThread until we get vsync. I'm not sure how often Swing calls the `paint` method, but blocking (not suspending) on EventDispatchThread may lead to poor performance because it's used not only for rendering but also for user event processing and performing callbacks that update UI related state.
So, the question, do we really need to wait for vsync when `JComponent#paint` is called? May we draw without waiting for vsync or maybe at least unblock paint and offload blocking to blockingDispatcher?

skiko/src/awtMain/objectiveC/macos/DisplayLinkThrottler.mm
igordmn: vsync also prevents unlimited FPS, which can load GPU/CPU. Do this code run for some time after sleeping?

PR #928, opened 2024-05-18
Fix PixelRef.rowBytes type

9il:
BTW, `pixels` pointer is missing in `PixelRef`.


elijah-semyonov:
It's indeed not a pointer:
https://api.skia.org/classSkPixelRef.html#ae3f171a1d99f2adfd0683654dac3efc7

@igordmn this fix is correct one, seems like we simply didn't use this API anywhere?

MatkovIvan (changes requested):
The type in JNI/C++ needs to be updated too.

https://github.com/JetBrains/skiko/blob/5ed970f08c3c6b6ff4fbbab74edd289fce16a54a/skiko/src/jvmMain/cpp/common/PixelRef.cc#L17

kropp:
Obsolete, has been fixed already

PR #935, opened 2024-06-07
Add API to configure the fallback fonts provider when all default methods return null 

eymar:
New API:

```kotlin
class TypefaceFontProviderWithFallback
class FontMgrWithFallback
```

Usage:

```kotlin
val fontCollection = FontCollection()

val notoEmojiTypeface = Typeface.makeFromData(Data.makeFromBytes(notoEmojisBytes))
val fallbackProvider = TypefaceFontProviderWithFallback().apply {
    registerTypeface(notoEmojiTypeface)
}

fontCollection.setDefaultFontManager(FontMgrWithFallback(fallbackProvider))
```

Screenshot 2024-06-07 at 14 48 28

### DIFF
--- a/skiko/src/commonMain/kotlin/org/jetbrains/skia/paragraph/TypefaceFontProvider.kt
+++ b/skiko/src/commonMain/kotlin/org/jetbrains/skia/paragraph/TypefaceFontProvider.kt
@@ -12,7 +12,7 @@ import org.jetbrains.skia.impl.NativePointer
 import org.jetbrains.skia.impl.getPtr
 import org.jetbrains.skia.impl.interopScope
 
-open class TypefaceFontProvider private constructor(
+open class TypefaceFontProvider internal constructor(
     ptr: NativePointer
 ) : FontMgr(ptr) {
 
--- a/skiko/src/commonMain/kotlin/org/jetbrains/skia/paragraph/TypefaceFontProviderWithFallback.kt
+++ b/skiko/src/commonMain/kotlin/org/jetbrains/skia/paragraph/TypefaceFontProviderWithFallback.kt
@@ -11,7 +11,7 @@ import org.jetbrains.skia.impl.reachabilityBarrier
 
 class TypefaceFontProviderWithFallback private constructor(
     ptr: NativePointer,
-) : FontMgr(ptr) {
+) : TypefaceFontProvider(ptr) {
 
     constructor() : this(_nMakeAsFallbackProvider())
 
@@ -21,7 +21,7 @@ class TypefaceFontProviderWithFallback private constructor(
         }
     }
 
-    fun registerTypeface(typeface: Typeface?, alias: String? = null): TypefaceFontProviderWithFallback {
+    override fun registerTypeface(typeface: Typeface?, alias: String?): TypefaceFontProviderWithFallback {
         return try {
             Stats.onNativeCall()
             interopScope {


eymar:
Without fallback or with only 1 font registered in fallback provider: 

Screenshot 2024-06-07 at 15 40 23
Screenshot 2024-06-07 at 15 40 39
Screenshot 2024-06-07 at 15 40 50

skiko/src/jvmMain/cpp/common/paragraph/TypefaceFontProvider.cc
```diff
@@ -12,7 +13,7 @@ extern "C" JNIEXPORT jlong JNICALL Java_org_jetbrains_skia_paragraph_TypefaceFon
     return reinterpret_cast<jlong>(instance);
 }
 
-extern "C" JNIEXPORT void JNICALL Java_org_jetbrains_skia_paragraph_TypefaceFontProviderKt__1nRegisterTypeface
+extern "C" JNIEXPORT jint JNICALL Java_org_jetbrains_skia_paragraph_TypefaceFontProviderKt__1nRegisterTypeface
```
MatkovIvan, line 16: There is return for this int

skiko/src/commonMain/kotlin/org/jetbrains/skia/paragraph/FontCollection.kt
```diff
@@ -106,7 +106,6 @@ class FontCollection internal constructor(ptr: NativePointer) : RefCnt(ptr) {
             reachabilityBarrier(this)
         }
     }
-
```
MatkovIvan, line 109: Please revert such formatting (no changes in this file)

skiko/src/jvmMain/cpp/common/paragraph/TypefaceFontProvider.cc
eymar: no compilation error despite this. 
thanks! fixed

samples/SkiaMultiplatformSample/src/wasmJsMain/kotlin/org/jetbrains/skiko/sample/Utils.kt
```diff
@@ -0,0 +1,56 @@
...
+import org.w3c.xhr.XMLHttpRequest
+import kotlin.coroutines.resume
+import kotlin.coroutines.resumeWithException
+import kotlin.coroutines.suspendCoroutine
+import kotlin.wasm.unsafe.UnsafeWasmMemoryApi
+import kotlin.wasm.unsafe.withScopedMemoryAllocator
+
+suspend fun loadRes(url: String): ArrayBuffer {
+    return suspendCoroutine { continuation ->
+        val req = XMLHttpRequest()
```
Schahen, line 14: nothing blocking, just out of pure interest - why not fetch?  Worser external declarations in js/wasm?

skiko/src/commonTest/kotlin/org/jetbrains/skia/FontFallbackTest.kt
```diff
@@ -0,0 +1,53 @@
...
+            val df = it.defaultFallback(0x6C34, FontStyle.NORMAL, null)
+            val glyphs = df!!.getStringGlyphs("水")
+
+            if (kotlinBackend.isWeb()) {
+                assertEquals("Noto Sans SC", df.familyName)
+                assertContentEquals(shortArrayOf(16391), glyphs)
+            } else {
+                // the actual font can be different on different systems,
+                // so just making sure it's not null
+                assertNotEquals(null, df?.familyName)
```
Schahen, line 35: looks like safe-call is redundant here

samples/SkiaMultiplatformSample/src/wasmJsMain/kotlin/org/jetbrains/skiko/sample/Utils.kt
eymar: copy pasted from another place without thinking. Will fix :)

skiko/src/commonTest/kotlin/org/jetbrains/skia/FontFallbackTest.kt
```diff
@@ -0,0 +1,53 @@
...
+            }
+        }
+    }
+
+    @Test
+    fun testChinese_noFallback() = runTest {
```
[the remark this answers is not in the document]
eymar: it's possible. we have a common test source for js and Wasm

skiko/src/commonTest/kotlin/org/jetbrains/skia/FontFallbackTest.kt
eymar: I agree with you. But the compiler doesn't like it:

> Type inference failed. The value of the type parameter T should be mentioned in input types (argument types, receiver type or expected type). Try to specify it explicitly.


If I add the type explicitly like this: `assertNotEquals<String?>(null, df.familyName)` 
it says that I can remove explicit type arguments. When I do, it shows the error as before.

Looks like a bug.

samples/SkiaMultiplatformSample/src/wasmJsMain/kotlin/org/jetbrains/skiko/sample/Utils.kt
eymar: fixed

skiko/src/commonTest/kotlin/org/jetbrains/skia/FontFallbackTest.kt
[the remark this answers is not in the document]
eymar: fixed

skiko/src/commonTest/kotlin/org/jetbrains/skia/FontFallbackTest.kt
eymar: changed to explicit type and removed a safe-call

Schahen (approved):

skiko/src/commonMain/kotlin/org/jetbrains/skia/paragraph/TypefaceFontProvider.kt
```diff
@@ -12,22 +12,40 @@ import org.jetbrains.skia.impl.NativePointer
...
+) : FontMgr(ptr) {
+
+    constructor(): this(TypefaceFontProvider_nMake(), false)
+
     companion object {
         init {
             staticLoad()
         }
+
+        fun createAsFallbackProvider(): TypefaceFontProvider {
```
igordmn, line 27: It looks like there is no reason for user to not always create `TypefaceFontProvider.createAsFallbackProvider`. The API and behaviour is the same, the only difference is that the fallback one can be passed to another function.

We can just pass `TypefaceFontProvider_nMakeAsFallbackProvider` in the constructor, and remove this function (and other unused functions/type checks after that).

skiko/src/commonMain/kotlin/org/jetbrains/skia/FontMgr.kt
```diff
@@ -10,6 +11,13 @@ open class FontMgr : RefCnt {
         }
 
         val default = FontMgr(_nDefault(), false)
+
+        fun defaultWithFallbackFontProvider(fallback: TypefaceFontProvider): FontMgr {
```
igordmn, line 15: The functionality of adding fallback seems doesn't relate to the default FontMgr, and the API looks unnecessary restricted from an external user perspective (allowing fallbacks only for the default manager).

I would make a separate extended class instead:
```
class FontMgrWithFallback: FontMgr() {
    constructor(fallback: TypefaceFontProvider): this(_nDefaultWithFallbackFontProvider(fallback._ptr))
}
```

This also mimics the C++ API, which is a good thing for keeping Skiko API as close as possible to the original.

User can use the API this way then:
```
val fallbackProvider = TypefaceFontProvider().apply {
    registerTypeface(notoEmojiTypeface)
}

fontCollection.setDefaultFontManager(FontMgrWithFallback(FontMgr.default, fallbackProvider))
```

igordmn (commented):
Thanks!

Looks good, but the API can be improved.

skiko/src/commonMain/kotlin/org/jetbrains/skia/FontMgr.kt
eymar: The usage is a bit different though:

the default `TypefaceFontProvider` won't provide the fallbacks.
But `TypefaceFontProvider.createAsFallbackProvider()` will provide them.

eymar: - add a new `TypefaceFontProviderWithFallback` it extends the existing `TypefaceFontProvider`
- `TypefaceFontProvider` is now an open class
- added a new`class FontMgrWithFallback(fallbackProvider: TypefaceFontProviderWithFallback)`, which relies on skia's FontMgr.default under the hood + our new fallback provider

skiko/src/commonMain/kotlin/org/jetbrains/skia/paragraph/TypefaceFontProviderWithFallback.kt
```diff
@@ -0,0 +1,56 @@
...
+import org.jetbrains.skia.ExternalSymbolName
+import org.jetbrains.skia.FontMgr
+import org.jetbrains.skia.ModuleImport
+import org.jetbrains.skia.Typeface
+import org.jetbrains.skia.impl.*
+import org.jetbrains.skia.impl.getPtr
+import org.jetbrains.skia.impl.interopScope
+import org.jetbrains.skia.impl.reachabilityBarrier
+
+class TypefaceFontProviderWithFallback private constructor(
```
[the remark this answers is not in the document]
eymar: Thanks! fixed now.

igordmn (approved):

PR #937, opened 2024-06-10
Fix Font linear metrics binding

MatkovIvan:
Add forgotten setter for linear metrics. Changed it to property.

Original functions have names [`isLinearMetrics`](https://github.com/JetBrains/skia/blob/6726cd3e1ac4f30d6874c278008cdf24dfc9b470/include/core/SkFont.h#L115-L119C12) and [`setLinearMetrics`](https://github.com/JetBrains/skia/blob/6726cd3e1ac4f30d6874c278008cdf24dfc9b470/include/core/SkFont.h#L156-L163)

### DIFF
--- a/skiko/src/commonMain/kotlin/org/jetbrains/skia/Font.kt
+++ b/skiko/src/commonMain/kotlin/org/jetbrains/skia/Font.kt
@@ -148,16 +148,25 @@ class Font : Managed {
         }
 
     /**
-     * @return  true if font and glyph metrics are requested to be linearly scalable
+     * Returns true if font and glyph metrics are requested to be linearly scalable.
+     *
+     * @return  true if font and glyph metrics are requested to be linearly scalable.
      */
-    fun areMetricsLinear(): Boolean {
-        return try {
+    var isMetricsLinear: Boolean
+        get() = try {
             Stats.onNativeCall()
             _nAreMetricsLinear(_ptr)
         } finally {
             reachabilityBarrier(this)
         }
-    }
+        set(value) {
+            try {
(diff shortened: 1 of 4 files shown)


skiko/src/commonMain/kotlin/org/jetbrains/skia/Font.kt
```diff
@@ -148,16 +148,25 @@ class Font : Managed {
...
 
     /**
-     * @return  true if font and glyph metrics are requested to be linearly scalable
+     * Returns true if font and glyph metrics are requested to be linearly scalable.
+     *
+     * @return  true if font and glyph metrics are requested to be linearly scalable.
      */
-    fun areMetricsLinear(): Boolean {
-        return try {
+    var isMetricsLinear: Boolean
```
igordmn, line 155:
```suggestion
    var isLinearMetrics: Boolean
```

I would name it this way, according the old [Skija policy](https://github.com/JetBrains/skija/blob/master/docs/Conventions.md).

skiko/src/commonMain/kotlin/org/jetbrains/skia/Font.kt
```diff
@@ -148,16 +148,25 @@ class Font : Managed {
...
+        get() = try {
             Stats.onNativeCall()
             _nAreMetricsLinear(_ptr)
         } finally {
             reachabilityBarrier(this)
         }
-    }
+        set(value) {
```
igordmn, line 162: Worth to add a smoke test in `FontTests` (just set and get)

igordmn (approved):

PR #940, opened 2024-06-12
Make getAccessibleContext return nullable AccessibleContext

elijah-semyonov:


igordmn (approved):

igordmn:
Non-nullable was made in https://github.com/JetBrains/skiko/pull/920/files

m-sasha:
I probably copied it from HardwareLayer; it's not nullable there too.

PR #958, opened 2024-07-09
Take ownership over NSWindow raw pointer when passing it from and to JNI

elijah-semyonov:
Speculative fix for https://youtrack.jetbrains.com/issue/FL-27529/Native-crash-on-macOS-while-being-idle

Speculative scenario:
```
// before

void bar(NSObject *object) {
    NSLog(@"%@", object);
}

void foo(void* ptr) {
    NSObject *object = (__bridge NSObject *)ptr;
    bar(object);
}

int main(int argc, const char * argv[]) {
    @autoreleasepool {
        NSObject *strongRef = [NSObject new];
        __weak NSObject *weakRef = strongRef;
        void *ptr = (__bridge void *)weakRef;
        strongRef = nil;
        foo(ptr); //crash
        
        NSLog(@"Hello");
    }
    return 0;
}
```

```
// after
void bar(NSObject *object) {
    NSLog(@"%@", object);
}

void foo(void* ptr) {
    NSObject *object = (__bridge_transfer NSObject *)ptr;
    bar(object);
}

int main(int argc, const char * argv[]) {
    @autoreleasepool {
        NSObject *strongRef = [NSObject new];
        __weak NSObject *weakRef = strongRef;
        void *ptr = (__bridge_retained void *)weakRef;
        strongRef = nil;
        foo(ptr); // no crash
        
        NSLog(@"Hello");
    }
    return 0;
}

```

### DIFF
--- a/skiko/src/awtMain/kotlin/org/jetbrains/skiko/SkiaLayer.awt.kt
+++ b/skiko/src/awtMain/kotlin/org/jetbrains/skiko/SkiaLayer.awt.kt
@@ -212,6 +212,8 @@ actual open class SkiaLayer internal constructor(
     /**
      * Returns the pointer to an OS specific window handle (native resource)
      * which the current [SkiaLayer] is attached.
+     * On macOS accessing this property returns an owning raw pointer to the NSWindow.
+     * Failure to pass this pointer to consuming APIs will result in a memory leak.
      */
     val windowHandle: Long
         get() = backedLayer.windowHandle
--- a/skiko/src/awtMain/objectiveC/macos/DisplayLinkThrottler.mm
+++ b/skiko/src/awtMain/objectiveC/macos/DisplayLinkThrottler.mm
@@ -162,13 +162,20 @@ JNIEXPORT void JNICALL Java_org_jetbrains_skiko_redrawer_DisplayLinkThrottler_di
 }
 
 JNIEXPORT void JNICALL Java_org_jetbrains_skiko_redrawer_DisplayLinkThrottler_waitVSync(JNIEnv *env, jobject obj, jlong throttlerPtr, jlong windowPtr) {
-    DisplayLinkThrottler *throttler = (__bridge DisplayLinkThrottler *) (void *) throttlerPtr;
-    NSWindow *window = (__bridge NSWindow *) (void *) windowPtr;
+    @autoreleasepool {
+        DisplayLinkThrottler *throttler = (__bridge DisplayLinkThrottler *) (void *) throttlerPtr;
 
-    // CVDisplayLink is conditionally set up on each draw request to track window.screen change to match throttling
-    // with the refresh rate of the actual screen the window is shown on.
-    [throttler setupDisplayLinkForWindow:window];
-    [throttler waitVSync];
+        NSWindow *window = nil;
+
+        if (windowPtr != 0) {
+            window = (__bridge_transfer NSWindow *) (void *) windowPtr;
+        }
+
+        // CVDisplayLink is conditionally set up on each draw request to track window.screen change to match throttling
+        // with the refresh rate of the actual screen the window is shown on.
+        [throttler setupDisplayLinkForWindow:window];
+        [throttler waitVSync];
+    }
 }
 
 }
--- a/skiko/src/awtMain/objectiveC/macos/Drawlayer.mm
+++ b/skiko/src/awtMain/objectiveC/macos/Drawlayer.mm
@@ -362,7 +362,14 @@ JNIEXPORT jlong JNICALL Java_org_jetbrains_skiko_HardwareLayer_getWindowHandle(J
 {
     @autoreleasepool {
         LayerHandler *layer = findByObject(env, component);
-        return (jlong) (__bridge void*) layer.window;
+
+        NSWindow *window = layer.window;
+
+        if (window) {
+            return (jlong) (__bridge_retained void*) window;
+        } else {
+            return 0;
+        }
     }
 }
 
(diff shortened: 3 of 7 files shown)


skiko/src/awtMain/kotlin/org/jetbrains/skiko/SkiaLayer.awt.kt
```diff
@@ -212,6 +212,8 @@ actual open class SkiaLayer internal constructor(
     /**
      * Returns the pointer to an OS specific window handle (native resource)
      * which the current [SkiaLayer] is attached.
+     * On macOS accessing this property returns an owning raw pointer to the NSWindow.
+     * Failure to pass this pointer to consuming APIs will result in a memory leak.
```
igordmn, line 216: `windowHandle` is a public API in Compose.

I have concerns with the current approach:
- calling it requires users to think about how to transfer ownership
- not sure, but it probably prohibits calling it multiple times
- when we pass it to `waitVSync`, can it still be accessible by user?
- on other systems, calling this doesn't transfer ownership - the window is only valid until it isn't closed

Can we fix the internal issue without changing ownership in this call? For example, we can take ownership on a redrawer creation and release it on `dispose`.

skiko/src/awtMain/objectiveC/macos/MetalRedrawer.mm
```diff
@@ -121,6 +121,12 @@ JNIEXPORT jlong JNICALL Java_org_jetbrains_skiko_redrawer_MetalRedrawer_createMe
     JNIEnv *env, jobject redrawer, jlong windowPtr, jboolean transparency, jlong adapterPtr, jlong platformInfoPtr)
 {
     @autoreleasepool {
+        NSWindow *window = nil;
+
+        if (windowPtr != 0) {
```
igordmn, line 126: `createMetalDevice` requires `window` to always exist. We should write the code with consideration that we always get a window here, without checking it with `if`.

skiko/src/awtMain/objectiveC/macos/Drawlayer.mm
```diff
@@ -362,7 +362,14 @@ JNIEXPORT jlong JNICALL Java_org_jetbrains_skiko_HardwareLayer_getWindowHandle(J
...
     @autoreleasepool {
         LayerHandler *layer = findByObject(env, component);
-        return (jlong) (__bridge void*) layer.window;
+
+        NSWindow *window = layer.window;
+
+        if (window) {
+            return (jlong) (__bridge_retained void*) window;
+        } else {
+            return 0;
```
igordmn, line 371: In which cases it might be 0?

skiko/src/awtMain/objectiveC/macos/DisplayLinkThrottler.mm
```diff
@@ -162,13 +162,20 @@ JNIEXPORT void JNICALL Java_org_jetbrains_skiko_redrawer_DisplayLinkThrottler_di
...
-    [throttler waitVSync];
+        NSWindow *window = nil;
+
+        if (windowPtr != 0) {
+            window = (__bridge_transfer NSWindow *) (void *) windowPtr;
+        }
+
+        // CVDisplayLink is conditionally set up on each draw request to track window.screen change to match throttling
+        // with the refresh rate of the actual screen the window is shown on.
+        [throttler setupDisplayLinkForWindow:window];
```
igordmn, line 176: If it is zero, then we don't wait for vsync? In which cases this can happen?

igordmn (changes requested):

skiko/src/awtMain/objectiveC/macos/DisplayLinkThrottler.mm
igordmn: I looked at the code, and it is written in consideration that `window` always exists in this moment:

```
    override fun dispose() = synchronized(drawLock) {
        frameDispatcher.cancel()
        contextHandler.dispose()
        disposeDevice(device.ptr)
        adapter.dispose()
        displayLinkThrottler.dispose()
        _device = null
        super.dispose()
    }

    private fun performDraw() = synchronized(drawLock) {
        if (!isDisposed) {
            ...
            displayLinkThrottler.waitVSync(windowPtr = layer.windowHandle)
            ...
        }
    }
```

`AWT` should retain link to the window until we call `super.dispose()` (or after it). Not sure if it is true in reality.

skiko/src/awtMain/objectiveC/macos/MetalRedrawer.mm
elijah-semyonov: I was thinking about putting an assertion in there, do they execute at all?

skiko/src/awtMain/objectiveC/macos/Drawlayer.mm
elijah-semyonov: Assigning a weak property to a strong ref will unwrap it and writing `nil` inside, if the object was deallocated.

skiko/src/awtMain/objectiveC/macos/MetalRedrawer.mm
igordmn: We usually add `check(windowPtr)` on the java side if we are not sure this is always successful.

skiko/src/awtMain/objectiveC/macos/DisplayLinkThrottler.mm
elijah-semyonov: Yeah, I don't know the reason for marking this specific spot as `weak` and there is nobody to ask.

elijah-semyonov:
The initial assumption was incorrect. The crash doesn't happen in:
`NSWindow *window = (__bridge NSWindow *) (void *) windowPtr;`
We reach `setupDisplayLinkForWindow`.

PR #959, opened 2024-07-10
Speculative fix for a crash

elijah-semyonov:
Usage of main-thread AppKit APIs in other thread is assumed to lead to the crash.
Second attempt for https://github.com/JetBrains/skiko/pull/958.

Structure display link control reactively around Wake/Sleep and Screen change notifications.

### DIFF
--- a/skiko/src/awtMain/objectiveC/macos/DisplayLinkThrottler.mm
+++ b/skiko/src/awtMain/objectiveC/macos/DisplayLinkThrottler.mm
@@ -68,6 +68,7 @@ static CVReturn displayLinkCallback(CVDisplayLinkRef displayLink, const CVTimeSt
         _window = window;
         _displayLink = nil;
         _vsyncConditionLock = [[NSConditionLock alloc] initWithCondition: 1];
+        _lock = [NSLock new];
         _isSleeping = NO;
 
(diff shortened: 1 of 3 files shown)


skiko/src/awtMain/objectiveC/macos/DisplayLinkThrottler.mm
```diff
@@ -21,125 +48,186 @@ static CVReturn displayLinkCallback(CVDisplayLinkRef displayLink, const CVTimeSt
...
 
-    result = CVDisplayLinkCreateWithCGDisplay([screenID unsignedIntValue], &displayLink);
+    result = CVDisplayLinkCreateWithCGDisplay(
+        _currentScreenID /* truncated to uint32_t aka CGDirectDisplayID */,
+        &_displayLink
+    );
 
     if (result != kCVReturnSuccess) {
-        return nil;
+        _displayLink = nil;
```
igordmn, line 208:
```suggestion
        _displayLink = nil;
        _displayLinkScreenID = nil;
```

Shouldn't we do this as well in all `return` points?

elijah-semyonov: No, and _displayLinkScreenID is uint32_t, we can't assign anything meaningful here.

igordmn (approved):

PR #960, opened 2024-07-19
Split skia download/unzipping task creation and invocation

Schahen:
This commit is first step in set of actions we need to take to build skia-package via gradle in the skiko repository

However it makes sense on it's own as well.

### DIFF
--- a/skiko/build.gradle.kts
+++ b/skiko/build.gradle.kts
@@ -1,8 +1,10 @@
+import de.undercouch.gradle.tasks.download.Download
 import org.gradle.api.tasks.testing.logging.TestExceptionFormat
 import org.gradle.crypto.checksum.Checksum
 import org.jetbrains.compose.internal.publishing.MavenCentralProperties
 import org.jetbrains.kotlin.gradle.targets.js.dsl.ExperimentalWasmDsl
 import tasks.configuration.*
+import kotlin.collections.HashMap
 
 plugins {
     kotlin("multiplatform")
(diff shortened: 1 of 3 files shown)


skiko/build.gradle.kts
```diff
@@ -12,6 +14,47 @@ plugins {
     id("org.gradle.crypto.checksum") version "1.4.0"
 }
 
+fun createDownloadTasks() {
```
eymar, line 17: Is it possible to keep this function in buildSrc?  The motivation is to limit the growth of build.gradle.kts

Schahen: Totally makes sense, renamed it to `declareSkiaTasks` and moved to the SkikoProjectContext (which belongs to buildSrc)

skiko/buildSrc/src/main/kotlin/properties.kt
```diff
@@ -162,7 +156,7 @@ class SkikoProperties(private val myProject: Project) {
                 System.getenv()["SKIA_DIR"]
                 ?: System.getProperty("skia.dir")
                 ?: myProject.findProperty("skia.dir")?.toString()
-            )?.let { File(it) }?.takeIf { it.isDirectory }
+            )?.let { File(it) }
```
eymar, line 159: Was this check for `isDirectory` removed intentionnaly?

eymar (approved):

skiko/buildSrc/src/main/kotlin/properties.kt
Schahen: Without validation  there was always a possibility to just have a seemingly valid local build which in actuality was remote build. I  totally agree it's better to keep it in getter. Pushed

PR #964, opened 2024-07-20
Adding an implementation for skiaLayer.detach()

kubapet:
Once SkiaLayer was created, it wasn't possible to reinstantiate it gracefully (e.g. when changing the geometry of the underlying canvas). It wasn't releasing the GL context so it wasn't possible to reuse the same HTML canvas when the window was resized etc.

I'm not sure if this is enough though, but it fixed my issues at least.


MatkovIvan (commented):
> but it fixed my issues at least.

Could you please provide test cases?

kubapet:
> > but it fixed my issues at least.
> 
> Could you please provide test cases?

Sorry for late reply, I was having vacation.
Unfortunately, I'm not able to reproduce it any more after upgrading my system. I was originally getting 
`WebGL: INVALID_OPERATION: drawElements: no valid shader program in use`
when trying to reuse the same SkiaLayer instance and reattaching it to resized HTML canvas.

So I guess it was related to a specific browser/GL stack on my machine.
I'm cancelling this PR for time being and will reopen it if I manage to reproduce it.
Thanks

kubapet:
Unable to reproduce the originally addressed issue

PR #968, opened 2024-07-25
Add a flag to control number of buffers used by Metal. Add a flag to control Vsync on window resize

SergeevPavel:
#### Number of buffers
MacOS [doc](https://developer.apple.com/documentation/quartzcore/cametallayer/2938720-maximumdrawablecount?language=objc). As far as I know AWT uses double buffering by default. So we would like to experiment with this setting in Fleet. Also I expect that double buffering should reduce user interaction latency.

#### Control VSync on window resize

I've noticed that on Linux and Windows we don't wait for VSync when repainting synchronously, but only on macOS we do.
I suspect that locking EDT when we are waiting for VSync might lead to problems with app responsivenes. So it would be nice to experiment with this setting.

### DIFF
--- a/skiko/src/androidMain/kotlin/org/jetbrains/skiko/Actuals.android.kt
+++ b/skiko/src/androidMain/kotlin/org/jetbrains/skiko/Actuals.android.kt
@@ -16,7 +16,7 @@ internal class AndroidOpenGLRedrawer(
 ) : Redrawer {
     override fun dispose() = TODO()
     override fun needRedraw() = TODO()
-    override fun redrawImmediately(waitForVsync: Boolean) = TODO()
+    override fun redrawImmediately() = TODO()
 
     override val renderInfo: String get() = "Android renderer"
 }
--- a/skiko/src/androidMain/kotlin/org/jetbrains/skiko/Actuals.android.kt
+++ b/skiko/src/androidMain/kotlin/org/jetbrains/skiko/Actuals.android.kt
@@ -16,7 +16,7 @@ internal class AndroidOpenGLRedrawer(
 ) : Redrawer {
     override fun dispose() = TODO()
     override fun needRedraw() = TODO()
-    override fun redrawImmediately() = TODO()
+    override fun redrawImmediately(waitForVsync: Boolean) = TODO()
 
     override val renderInfo: String get() = "Android renderer"
 }
--- a/skiko/src/awtMain/kotlin/org/jetbrains/skiko/redrawer/MetalRedrawer.kt
+++ b/skiko/src/awtMain/kotlin/org/jetbrains/skiko/redrawer/MetalRedrawer.kt
@@ -114,7 +114,7 @@ internal class MetalRedrawer(
         inDrawScope {
             update(System.nanoTime())
             if (!isDisposed) { // Redrawer may be disposed in user code, during `update`
-                performDraw()
+                performDraw(waitVsync = SkikoProperties.macOSWaitForPreviousFrameVsyncOnRedrawImmediately)
             }
         }
     }
--- a/skiko/src/awtMain/kotlin/org/jetbrains/skiko/redrawer/MetalRedrawer.kt
+++ b/skiko/src/awtMain/kotlin/org/jetbrains/skiko/redrawer/MetalRedrawer.kt
@@ -67,8 +67,9 @@ internal class MetalRedrawer(
 
     init {
         onDeviceChosen(adapter.name)
+        val numberOfBuffers = properties.frameBuffering.numberOfBuffers() ?: 0 // zero means default for system
         val initDevice = layer.backedLayer.useDrawingSurfacePlatformInfo {
-            MetalDevice(createMetalDevice(layer.windowHandle, layer.transparency, adapter.ptr, it))
+            MetalDevice(createMetalDevice(layer.windowHandle, layer.transparency, numberOfBuffers, adapter.ptr, it))
         }
         _device = initDevice
         contextHandler = MetalContextHandler(layer, initDevice, adapter)
@@ -154,13 +155,14 @@ internal class MetalRedrawer(
         windowOcclusionStateChannel.trySend(isOccluded)
     }
 
-    private fun performDraw() = synchronized(drawLock) {
+    private fun performDraw(waitVsync: Boolean = true) = synchronized(drawLock) {
         if (!isDisposed) {
-            // Wait for vsync because:
-            // - macOS drops the second/next drawables if they are sent in the same vsync
-            // - it makes frames consistent and limits FPS
-            displayLinkThrottler.waitVSync()
-
+            if (waitVsync) {
+                // Wait for vsync because:
+                // - macOS drops the second/next drawables if they are sent in the same vsync
+                // - it makes frames consistent and limits FPS
+                displayLinkThrottler.waitVSync()
+            }
             autoreleasepool {
                 contextHandler.draw()
             }
@@ -185,7 +187,7 @@ internal class MetalRedrawer(
         setLayerVisible(device.ptr, isVisible)
     }
 
-    private external fun createMetalDevice(window: Long, transparency: Boolean, adapter: Long, platformInfo: Long): Long
+    private external fun createMetalDevice(window: Long, transparency: Boolean, frameBuffering: Int, adapter: Long, platformInfo: Long): Long
     private external fun disposeDevice(device: Long)
     private external fun resizeLayers(device: Long, x: Int, y: Int, width: Int, height: Int)
     private external fun setLayerVisible(device: Long, isVisible: Boolean)
--- a/skiko/src/awtMain/kotlin/org/jetbrains/skiko/redrawer/MetalRedrawer.kt
+++ b/skiko/src/awtMain/kotlin/org/jetbrains/skiko/redrawer/MetalRedrawer.kt
@@ -109,12 +109,12 @@ internal class MetalRedrawer(
         frameDispatcher.scheduleFrame()
     }
 
-    override fun redrawImmediately() {
+    override fun redrawImmediately(waitForVsync: Boolean) {
         check(!isDisposed) { "MetalRedrawer is disposed" }
         inDrawScope {
             update(System.nanoTime())
             if (!isDisposed) { // Redrawer may be disposed in user code, during `update`
-                performDraw()
+                performDraw(waitForVsync)
             }
         }
     }
--- a/skiko/src/awtMain/kotlin/org/jetbrains/skiko/redrawer/MetalRedrawer.kt
+++ b/skiko/src/awtMain/kotlin/org/jetbrains/skiko/redrawer/MetalRedrawer.kt
@@ -109,12 +109,12 @@ internal class MetalRedrawer(
         frameDispatcher.scheduleFrame()
     }
 
-    override fun redrawImmediately(waitForVsync: Boolean) {
+    override fun redrawImmediately() {
         check(!isDisposed) { "MetalRedrawer is disposed" }
         inDrawScope {
             update(System.nanoTime())
             if (!isDisposed) { // Redrawer may be disposed in user code, during `update`
-                performDraw(waitForVsync)
+                performDraw()
             }
         }
     }
--- a/skiko/src/awtMain/objectiveC/macos/MetalRedrawer.mm
+++ b/skiko/src/awtMain/objectiveC/macos/MetalRedrawer.mm
@@ -118,7 +118,7 @@ extern "C"
 {
 
 JNIEXPORT jlong JNICALL Java_org_jetbrains_skiko_redrawer_MetalRedrawer_createMetalDevice(
-    JNIEnv *env, jobject redrawer, jlong windowPtr, jboolean transparency, jlong adapterPtr, jlong platformInfoPtr)
+    JNIEnv *env, jobject redrawer, jlong windowPtr, jboolean transparency, jint frameBuffering, jlong adapterPtr, jlong platformInfoPtr)
 {
     @autoreleasepool {
         id<MTLDevice> adapter = (__bridge id<MTLDevice>) (void *) adapterPtr;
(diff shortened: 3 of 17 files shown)


SergeevPavel:
@elijah-semyonov @igordmn could you take a look?

elijah-semyonov (approved):

skiko/src/androidMain/kotlin/org/jetbrains/skiko/Actuals.android.kt
```diff
@@ -16,7 +16,7 @@ internal class AndroidOpenGLRedrawer(
 ) : Redrawer {
     override fun dispose() = TODO()
     override fun needRedraw() = TODO()
-    override fun redrawImmediately() = TODO()
+    override fun redrawImmediately(waitForVsync: Boolean) = TODO()
```
igordmn, line 19: Without knowing your context, I am leaning towards not adding this parameter for configuration, and just change the default behaviour on macOs. I am not sure in which cases it is actually needed to wait for vsync in a case of `redrawImmediately`. 

Initially it was designed that this method shouldn't wait for vsync, and redraw the rendering buffer to the end of this method. Waiting for vsync on macOs was added because we could lose some frames if we don't wait.

skiko/src/awtMain/objectiveC/macos/MetalRedrawer.mm
```diff
@@ -133,6 +133,10 @@ JNIEXPORT jlong JNICALL Java_org_jetbrains_skiko_redrawer_MetalRedrawer_createMe
         [container setNeedsDisplayOnBoundsChange: YES];
 
         AWTMetalLayer *layer = [AWTMetalLayer new];
+        if (frameBuffering == 2 || frameBuffering == 3) {
```
igordmn, line 136: We either need to write it unconditionally, or add `else` branch. Otherwise setting frameBuffering = 2, then 0 won't change anything

skiko/src/awtMain/kotlin/org/jetbrains/skiko/redrawer/MetalRedrawer.kt
```diff
@@ -154,13 +155,14 @@ internal class MetalRedrawer(
...
-            // Wait for vsync because:
-            // - macOS drops the second/next drawables if they are sent in the same vsync
-            // - it makes frames consistent and limits FPS
-            displayLinkThrottler.waitVSync(windowPtr = layer.windowHandle)
-
+            if (waitVsync) {
+                // Wait for vsync because:
+                // - macOS drops the second/next drawables if they are sent in the same vsync
+                // - it makes frames consistent and limits FPS
+                displayLinkThrottler.waitVSync(windowPtr = layer.windowHandle)
```
igordmn, line 164: @elijah-semyonov, as far as I remember not calling this caused frame loosing, no matter what `maximumDrawableCount` was (a default is 3?)

skiko/src/commonMain/kotlin/org/jetbrains/skiko/GraphicsApi.kt
```diff
@@ -29,3 +29,17 @@ enum class GpuPriority(val value: String) {
         fun parseOrNull(value: String): GpuPriority? = GpuPriority.values().find { it.value == value }
     }
 }
+
+enum class FrameBuffering {
+    DEFAULT,
+    DOUBLE,
+    TRIPLE
```
igordmn, line 36: > refactor with [nextDrawable](https://developer.apple.com/documentation/quartzcore/cametallayer/1478172-nextdrawable?language=objc) being called on IO dispatcher and non-blocking introspection*

@elijah-semyonov, do we really need a separate thread?

As far as I understand Triple buffering, there a 3 buffers and we can swap them without blocking and moving operations to another thread:
```
updating buffer <-> back buffer <-> front buffer
```

igordmn (changes requested):

skiko/src/androidMain/kotlin/org/jetbrains/skiko/Actuals.android.kt
SergeevPavel: Yes, I started from that. The problem is that when we don't wait for VSync on window resize user experience slightly degrade. I mean that the gap between layer and window border become more noticable. So I would like to play with this settings a liitle bit

igordmn: >  redraw the rendering buffer to the end of this method

I.e., if vsync happens after this method, it is guarantied that we'll see the new content on the screen.

skiko/src/awtMain/kotlin/org/jetbrains/skiko/redrawer/MetalRedrawer.kt
```diff
@@ -154,13 +155,14 @@ internal class MetalRedrawer(
         windowOcclusionStateChannel.trySend(isOccluded)
     }
 
-    private fun performDraw() = synchronized(drawLock) {
+    private fun performDraw(waitVsync: Boolean = true) = synchronized(drawLock) {
```
igordmn, line 158: We don't call `performDraw = false` anywhere. I suppose we should change `redrawImmediately`

skiko/src/androidMain/kotlin/org/jetbrains/skiko/Actuals.android.kt
igordmn: >  when we don't wait for VSync on window resize

If we don't change the behaviour on resize (waiting for vsync), what is the case for `redrawImmediately(vsync = false)`?

In Compose we have 2 cases for `redrawImmediately`:
- window resize
- window showing (draw content before window is shown to avoid white/black blinking)

Also one more hypothetical case - manual calling to reduce input latency:
```
mouse move
change state
redrawImmediately

mouse move
change state
redrawImmediately

mouse move
change state
redrawImmediately

vsync
```

>  I mean that the gap between layer and window border become more noticable

This is strange. In my head waiting for something (vsync) should increase the gap, not decrease it.

skiko/src/awtMain/kotlin/org/jetbrains/skiko/redrawer/MetalRedrawer.kt
elijah-semyonov: That's right.

skiko/src/commonMain/kotlin/org/jetbrains/skiko/GraphicsApi.kt
elijah-semyonov: If we don't want to sleep on `nextDrawable` - yes, we need extra threads, it's blocking.

elijah-semyonov:
Not to jump between threads, the original rationale behind adding this functionality was this:
Excessive redraws lead to presented drawables being discarded silently. Compose internal state was out of sync with what was on the screen.

And the problem with macOS is that it's completely opaque on how it manages the drawable pool. It may, or may not block on `nextDrawable`.

```swift
import Cocoa

class View: NSView {
    let metalLayer: CAMetalLayer = CAMetalLayer()
    override init(frame frameRect: NSRect) {
        super.init(frame: frameRect)
        
        let metalDevice = MTLCreateSystemDefaultDevice()!
        metalLayer.device = metalDevice
        metalLayer.pixelFormat = .bgra8Unorm
        metalLayer.framebufferOnly = true
        metalLayer.maximumDrawableCount = 3
        metalLayer.allowsNextDrawableTimeout = false
        wantsLayer = true
    }
    
    override func layoutSubtreeIfNeeded() {
        super.layoutSubtreeIfNeeded()
        
        metalLayer.frame = bounds
        metalLayer.drawableSize = frame.size
    }
    
    required init?(coder: NSCoder) {
        nil
    }
    
    override func makeBackingLayer() -> CALayer {
        return metalLayer
    }
}

class ViewController: NSViewController {
    let metalView = View(frame: NSRect(x: 0, y: 0, width: 200, height: 200))

    override func loadView() {
        view = metalView
    }

    override func viewDidAppear() {
        super.viewDidAppear()
        
        var drawables: [CAMetalDrawable] = []
        for i in 0..<100 {
            let drawable = metalView.metalLayer.nextDrawable()!
            let texture = drawable.texture
            drawables.append(drawable)
            print(i, drawable, unsafeBitCast(texture, to: UInt.self))
        }
    }
}
```
On my Mac this snippet just spits 100 drawables at once (all with different addresses).

skiko/src/androidMain/kotlin/org/jetbrains/skiko/Actuals.android.kt
SergeevPavel: > This is strange. In my head waiting for something (vsync) should increase the gap

At my understanding blocking the paint method prevent window border from movement. Which means it may lag behind mouse cursor but move along with skia content (not perfectly).

> what is the case for redrawImmediately(vsync = false)?

I'm considering to change it on window resize in Fleet subclass of `SkiaLayer`. But I would like to have flexibility here, e.g. to be able to roll it back in case it would look too bad on 120hz displays. So it seems it may look different with different number of buffers in swap chain.

skiko/src/awtMain/kotlin/org/jetbrains/skiko/redrawer/MetalRedrawer.kt
SergeevPavel: Good catch! I missed it during refactoring

skiko/src/commonMain/kotlin/org/jetbrains/skiko/GraphicsApi.kt
igordmn: Discussed online - there can be a way to not wait for `nextDrawable` in a case of triple buffering, if we manually control acquired/completed `drawables`.

I.e. we can postpone commiting the third buffer, if 2 previous one are still not completed. And reuse it, when there will a urgent request for redrawing (`redrawImmediately` on resize)

skiko/src/awtMain/kotlin/org/jetbrains/skiko/redrawer/MetalRedrawer.kt
igordmn: @SergeevPavel, you should take this into account. If you disable waiting for vsync on macOs, some of the last drawn frame can be discarded. The alternative solution to vsync is might be [this](https://github.com/JetBrains/skiko/pull/968/files#r1695524472).

skiko/src/androidMain/kotlin/org/jetbrains/skiko/Actuals.android.kt
igordmn: It is okay to merge if it is more convenient to experiment in Fleet first.

But if it will be no big burden, and our goal is to fix glitches on resize, better to experiment with local Skiko (or some build from `a branch`). Because in the end we might not need this to be configurable. Also, the end solution should be applied to Compose as well, it means that resizing itself should be fixed in `SkiaLayer`, not in its subclass.

Configurable vsync here adds a burden of supporting it (people have to read/understand it) and implementing it properly (should `SkikoProperties.isVsyncEnabled` affect it, why is it different default values for different OS'es, what are use cases, etc)

igordmn: If we come to a conclusion that resize glitches are caused by vsync waiting, then we need to not only remove it, but also implement a proper triple buffering (see the other comment) that might help avoiding missed frames.

SergeevPavel: > It is okay to merge if it is more convenient to experiment in Fleet first.

Let's merge it then. What you do in `SkiaLayer` will be the default behaviour. TBH that it will be a problem for other library users. 

> If we come to a conclusion that resize glitches are caused by vsync waiting

I think the main cause of resize issue is concurrency model of AWT (with two threads EDT and AppKitThread) with Java2D backed by OpenGL layer with Skia Metal layer on top of it.
In JWM straightforward approach works quite well https://github.com/HumbleUI/JWM/blob/5bdc87bca4ebbb25514794dcb866e801620ea71a/macos/cc/LayerMetal.mm#L76
https://github.com/user-attachments/assets/afbdfcd2-449e-4b13-8de3-abc6cf2d858f

I think that the right way to fix it would be to get rid of AWT. But for now I'm looking for a way to improve UX with what we have now.

Also I suspect that sometimes EDT in Fleet spend too much time on this locking, which lead to unresponsiveness. And maybe we could sacrifice some window resize ux if it turns to be true.

skiko/src/commonMain/kotlin/org/jetbrains/skiko/GraphicsApi.kt
SergeevPavel: I'm not sure that I fully understand the idea. I would be happy to discuss the ideas on how to tackle this problem offline, or on call. So don't hesitate to approach me, I'm in office in every weekday

skiko/src/androidMain/kotlin/org/jetbrains/skiko/Actuals.android.kt
igordmn: > and implementing it properly 

Also, I noticed, that Windows/Linux has a different semantics - `wait for vsync after drawing`. On macOs we have `wait for vsync before drawing if the previous drawing was in this vsync`.

Do you need it for Windows/Linux? Can we just add a system property `skiko.rendering.macos.waitForPreviousFrameVsyncOnRedrawImmediately`?

> Let's merge it then. 

Before the merge we need to add a comment describing situation, to avoid future misunderstanding of this code:
```
/*
- wait for vsync before drawing if the previous drawing was in this vsync. Affects only `redrawImmediately` (called on window resize)
- affects only macOs
- this is for experiments for window resize
- setting it false may cause missing frames https://github.com/JetBrains/skiko/pull/708
*/
```
If we decide to implement Windows/Linux, I would implement it as another property:
```
Wait for vsync after drawing. Affects only `redrawImmediately` (called on window resize)

Windows/Linux also depends on SkikoProperties.isVsyncEnabled (greater priority)
```

skiko/src/commonMain/kotlin/org/jetbrains/skiko/GraphicsApi.kt
igordmn: > I would be happy to discuss the ideas

If you are working on improving resize, this should be considered. As it doesn't block, and might not have an issue of missing frame (a theory).

igordmn (approved):

PR #971, opened 2024-08-05
Fix skia download destination directory

eymar:
The download directory path now depends on the buildType: Debug and Release

___
It affected our buildServer CI pipeline where it builds Release first and then Debug. And downloadTask was considered UP-TO-DATE when Debug build was attempted. Therefore the build used "Release" version of skia instead of "Debug. "

Added a Debug build to the githiub workflow too.


eymar:
Merging it after the CI passed. The change is trivial.

Issue #967, opened 2024-07-24
 `RuntimeException "Can't wrap nullptr"` while working with ManagedString


Schahen:
Following code: 
```
import org.jetbrains.skia.ManagedString
import org.jetbrains.skia.shaper.IcuBidiRunIterator


val text = ""
println(ManagedString(text))
println(IcuBidiRunIterator(text, 0))
```

Fails on wasm with `RuntimeException "Can't wrap nullptr"`

Schahen:
Here's [first commit](https://github.com/JetBrains/skiko/commit/f21b1970769b84b0c19e6306ab277b06e8703ffe) failing for test

```
 @Test
    fun bugReport() {
        ManagedString("").use { str ->
            IcuBidiRunIterator(str, false, 0)
        }
    }
```

serjsysoev:
Confirming that this was fixed by #972, thank you!

serjsysoev:
Can confirm, this issue is fixed. `IcuBidiRunIterator` still throws sometimes on empty input. I will file a new issue once I find a stable reproducer.

okushnikov:
Please check the <a href="https://youtrack.jetbrains.com/issue/SKIKO-967">following ticket</a> on YouTrack for follow-ups to this issue. GitHub issues will be closed in the coming weeks.

PR #972, opened 2024-08-06
Use convertToZeroTerminatedString on both wasm and js targets

Schahen:
This is supposed to be fix for https://github.com/JetBrains/skiko/issues/967

The idea is pass strings as (utf8) byte arrays directly through the border
just like we do it on native

### DIFF
--- a/skiko/src/commonTest/kotlin/org/jetbrains/skia/ManagedStringTest.kt
+++ b/skiko/src/commonTest/kotlin/org/jetbrains/skia/ManagedStringTest.kt
@@ -24,6 +24,12 @@ class ManagedStringTest {
     }
 
     @Test
+    fun emptyStringTest() {
+        val s = ManagedString("")
+        assertEquals("", s.toString())
+    }
+
+    @Test
     fun canCreateAndReadManagedString() {
         val ms1 = ManagedString("Hello")
         assertEquals("Hello", ms1.toString())
--- a/skiko/src/commonTest/kotlin/org/jetbrains/skia/ManagedStringTest.kt
+++ b/skiko/src/commonTest/kotlin/org/jetbrains/skia/ManagedStringTest.kt
@@ -24,12 +24,6 @@ class ManagedStringTest {
     }
 
     @Test
-    fun emptyStringTest() {
-        val s = ManagedString("")
-        assertEquals("", s.toString())
-    }
-
-    @Test
     fun canCreateAndReadManagedString() {
         val ms1 = ManagedString("Hello")
         assertEquals("Hello", ms1.toString())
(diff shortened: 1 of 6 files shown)


skiko/src/nativeJsMain/kotlin/org/jetbrains/skia/impl/Native.nativejs.kt
```diff
@@ -0,0 +1,12 @@
...
+
+/**
+ * Converts String to zero-terminated utf-8 byte array.
+ */
+internal fun convertToZeroTerminatedString(string: String): ByteArray {
+    //  C++ needs char* with zero byte at the end. So we need to copy array with an extra zero byte.
+
+    val utf8 = string.encodeToByteArray() // encodeToByteArray encodes to utf8
+    // TODO Remove array copy, use `skString(data, length)` instead of `skString(data)`
+    return utf8.copyOf(utf8.size + 1)
```
eymar, line 11: Is it guaranteed that the 1 extra byte would be `\0`? Or maybe we better set its value explicitly?

Schahen: it's guaranteed by  spec for  ByteArray.copyOf from Kotlin stdlib

skiko/src/commonTest/kotlin/org/jetbrains/skia/ManagedStringTest.kt
```diff
@@ -23,6 +23,12 @@ class ManagedStringTest {
         assertEquals("你好", s3)
     }
 
+    @Test
+    fun emptyStringTest() {
```
eymar, line 27: What do you think about adding a few more simple tests with intiially empty strings:

`s.insert(0, "")`, `s.append("")`.

Also I'm wondering what is going to happen if we call insert with an index out of bounds. Also `remove` with an index out of bounds.

eymar (approved):

skiko/src/nativeJsMain/kotlin/org/jetbrains/skia/impl/Native.nativejs.kt
```diff
@@ -0,0 +1,12 @@
+package org.jetbrains.skia.impl
+
+/**
+ * Converts String to zero-terminated utf-8 byte array.
+ */
+internal fun convertToZeroTerminatedString(string: String): ByteArray {
```
MatkovIvan, line 6: The name is confusing - I thought that `toInterop` is recursive. It converts to `ByteArray`, not string

Also better to include "utf8" in some form to the name

PR #985, opened 2024-09-03
Preserve profiling information in skiko.wasm with an extra build flag

eymar:
Adding a new flag `-Pskiko.wasm.withProfiling=true` which adds `--profiling` flag to emcc, so it preserves the function names - makes it easier to use the browser profiler.

By default, it's disabled.


Schahen (approved):
LGTM

PR #988, opened 2024-09-16
Fix bounds of ComposePanel in IntelliJ on macOs

igordmn:
Fixes https://youtrack.jetbrains.com/issue/CMP-5856/Desktop-ComposePanel-size-breaks-with-.fillMax-modifiers#focus=Comments-27-10632441.0-0

Fixes https://youtrack.jetbrains.com/issue/CMP-5968/Compose-content-is-rendered-in-the-wrong-place-in-IJ-when-using-AWT-compositing

Regression after https://github.com/JetBrains/skiko/pull/661/files#diff-910a6e28fda20a00bc98c6a8a04f74ab701d79e841b9baddf146b810e610572fR363 (`setBounds` is called more often, but not enough as `ancestorMoved`)

When a panel changes its position without changing its size, `doLayout` isn't called because the content itself wasn't changed. But we still need to update the bounds of the underlying layer.

## Testing
- https://youtrack.jetbrains.com/issue/CMP-5856/Desktop-ComposePanel-size-breaks-with-.fillMax-modifiers#focus=Comments-27-10632441.0-0 isn't reproducible after the fix
- there are no resize glitches

### Fixes
- Fix bounds of ComposePanel in IntelliJ on macOs


skiko/src/awtMain/kotlin/org/jetbrains/skiko/SkiaLayer.awt.kt
```diff
@@ -143,7 +155,7 @@ actual open class SkiaLayer internal constructor(
         addPropertyChangeListener("graphicsContextScaleTransform") {
             Logger.debug { "graphicsContextScaleTransform changed for $this" }
             latestReceivedGraphicsContextScaleTransform = it.newValue as AffineTransform
-            redrawer?.syncSize()
+            revalidate()
```
igordmn, line 158: Changed, because layout depends on the scale inside `adjustSizeToContentScale`

igordmn: Probably it can fix https://youtrack.jetbrains.com/issue/CMP-5811/Moving-Desktop-window-across-monitors-causes-change-in-layout-since-1.7.0-alpha01-when-running-from-the-IDE

m-sasha: Sounds like it could bring back https://youtrack.jetbrains.com/issue/CMP-6550/Resizing-the-window-using-WindowState-breaks-the-UI

Is there a pure Compose reproducer I can try out?

igordmn: A pure Compose is:
```
import androidx.compose.foundation.background
import androidx.compose.foundation.layout.Box
import androidx.compose.foundation.layout.fillMaxSize
import androidx.compose.ui.Modifier
import androidx.compose.ui.awt.ComposePanel
import androidx.compose.ui.graphics.Color
import kotlinx.coroutines.delay
import kotlinx.coroutines.runBlocking
import org.jetbrains.skiko.MainUIDispatcher
import java.awt.Dimension
import javax.swing.JFrame
import javax.swing.JLayeredPane
import javax.swing.WindowConstants

fun main() {
    runBlocking(MainUIDispatcher) {
        val window = JFrame()
        val panel = ComposePanel()
        panel.setContent {
            Box(Modifier.fillMaxSize().background(Color.Red))
        }

        panel.size = Dimension(100, 100)
        val box = javax.swing.Box.createVerticalBox().apply {
            add(panel)
        }
        box.setBounds(0, 0, 100, 100)

        try {
            val container = JLayeredPane()
            container.add(box)
            window.contentPane.add(container)
            window.setLocation(200, 200)
            window.size = Dimension(200, 200)
            window.defaultCloseOperation = WindowConstants.DISPOSE_ON_CLOSE
            window.isUndecorated = true
            window.isVisible = true

            delay(1000)

            box.setBounds(100, 0, 100, 100)
            delay(10000)
        } finally {
            window.dispose()
        }
    }
}
```

After

igordmn: I checked https://youtrack.jetbrains.com/issue/CMP-6550/Resizing-the-window-using-WindowState-breaks-the-UI with this Skiko, it doesn't reproduce (it reproduces in 1.7.0-beta02 on my machine)

skiko/src/awtMain/kotlin/org/jetbrains/skiko/SkiaLayer.awt.kt
```diff
@@ -133,6 +135,16 @@ actual open class SkiaLayer internal constructor(
         @Suppress("LeakingThis")
         add(backedLayer)
 
+        addAncestorListener(object : AncestorListener {
+            override fun ancestorAdded(event: AncestorEvent?) = Unit
+
+            override fun ancestorRemoved(event: AncestorEvent?) = Unit
+
+            override fun ancestorMoved(event: AncestorEvent?) {
+                revalidate()
```
m-sasha, line 144: Why not `redrawer?.syncBounds()` here?

m-sasha (approved):

skiko/src/awtMain/kotlin/org/jetbrains/skiko/SkiaLayer.awt.kt
m-sasha: `revalidate` sounds more expensive...

skiko/src/awtMain/kotlin/org/jetbrains/skiko/SkiaLayer.awt.kt
```diff
@@ -579,16 +590,15 @@ actual open class SkiaLayer internal constructor(
...
         // We check values close to 0.5 and edit the size to avoid white lines glitch
-        if (diff > 0.4f && diff < 0.6f) {
-            rounded = value + 1f
+        return if (diff > 0.4f && diff < 0.6f) {
+            (value + 1f).toInt()
         } else {
-            rounded = value.toFloat()
+            value.toFloat().toInt()
         }
-        return rounded.toInt()
```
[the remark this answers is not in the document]
igordmn: Done

skiko/src/awtMain/kotlin/org/jetbrains/skiko/SkiaLayer.awt.kt
m-sasha: I guess I'm just not sure I understand why revalidation is needed here... Only the (absolute) position of the layer changes, not the size, so layout should not change.

Maybe `SwingUtilities.invokeLater { redrawer?.syncBounds() }` ?

skiko/src/awtMain/kotlin/org/jetbrains/skiko/SkiaLayer.awt.kt
```diff
@@ -647,3 +643,19 @@ internal fun Canvas.clipRectBy(rectangle: ClipRectangle, scale: Float) {
         true
     )
 }
+
+// TODO Recheck this method validity in 2 cases - full Window content, and a Panel content
+//  https://youtrack.jetbrains.com/issue/CMP-5447/Window-white-line-on-the-bottom-before-resizing
+/**
+ * Change [value] size to avoid rounding errors, when converting AWT Int coordinates to pixels.
```
m-sasha, line 650: What does it mean "avoid rounding errors"? What rounding errors? The link to the ticket does help though.

But why 0.4 and 0.6 specifically?

Should we maybe instead use the same function that determines the real content size and compare it to the window/container size? If it's smaller, then increase the unscaled size.

e.g.
```
return if ((value * contentScale).roundToInt() < windowSize)
    value + 1
else
    value
```
(assuming the scaled/real content size is `(value * contentScale).roundToInt()`).

Anyway, this is not a blocker for this PR in any way.

skiko/src/awtMain/kotlin/org/jetbrains/skiko/SkiaLayer.awt.kt
igordmn: We can't say that `invokeLater` is in sync with layout.

Actually I figured why there were resize glitches. `addAncestorListener` receives events asynchronously. I replaced it by `addHierarchyBoundsListener` which is called synchronously on the parent layout. We don't need `revalidate` then.

Also, I looked at the `revalidate` doc, and you are right, it is in't performant, as it actually revalidates the whole tree back to the root.

So, I removed it from the other place, I am not confident now to fix another very small issue (recalling `adjustSizeToContentScale` on content scale change).

igordmn: Sorry, wait a minute. I tested it with a wrong version. Glitches are back. Investigating

skiko/src/awtMain/kotlin/org/jetbrains/skiko/SkiaLayer.awt.kt
igordmn: Rounding errors when converting Int points (10) to pixels (12) with float scale (1.25)

I will change comment, but I will not change the logic in this PR, as it isn't relevant to it, and I also don't know the whole context of the fix itself, just knowledge of the issue.

igordmn: The link to the issue is in the comment above, this is directly related.

As far as I understand, with this code there was an attempt to increase the size of the component, to avoid white line (but instead overlapped one meaningful content line)

skiko/src/awtMain/kotlin/org/jetbrains/skiko/SkiaLayer.awt.kt
igordmn: I returned to `revalidate`.

I looked at the documentation:
https://docs.oracle.com/javase/7/docs/api/javax/swing/JComponent.html#revalidate()

- it notify the `validateRoot`, but it doesn't validate all children, just what was `invalidated`
- it seems that only `validate` is expensive, but it is deferred, and actually is already called on every parent size change or when a component is added/removed

I don't want to user `invokeLater` because:
- we need to handle a proper cancelation during `dispose`
- we need to handle double `invokeLater`
- we are not sure that it will be called the same time as the layout. I investigated resize glitches, and it appeared that we called `syncBounds` too often - we called it even if it haven't performed `doLayout` yet, and so `syncBounds` was based on incorrect size.

m-sasha (approved):

PR #990, opened 2024-09-18
Use actual openURL method on iOS

ASalavei:
Fixes the following error when calling `URIHandler_openUri` on iOS 18:

```
The caller of UIApplication.openURL(_:) needs to migrate to the non-deprecated UIApplication.open(_:options:completionHandler:). Force returning false (NO).
```

Fixes: https://youtrack.jetbrains.com/issue/CMP-6699

### DIFF
--- a/skiko/src/uikitMain/kotlin/org/jetbrains/skiko/Actuals.uikit.kt
+++ b/skiko/src/uikitMain/kotlin/org/jetbrains/skiko/Actuals.uikit.kt
@@ -4,7 +4,11 @@ import platform.Foundation.NSURL.Companion.URLWithString
 import platform.UIKit.UIApplication
 
 internal actual fun URIHandler_openUri(uri: String) {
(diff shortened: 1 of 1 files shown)


elijah-semyonov:
Need to build skiko and upgrade version then, it's critical due to apps likely not able to pass submission checks.

elijah-semyonov:
You are absolutely right, it's a forgotten leftover. We fix it first, remove later.

ASalavei:
Good that we spotted it. I believe we will discuss removal soon.
Anyway, the API seems to be going through a deprecation process, and there are some [more](https://github.com/JetBrains/skiko/blob/5766e0b3f011bc3c0847a7e51ba93bd696da13cd/skiko/src/commonMain/kotlin/org/jetbrains/skiko/Platform.kt) of them.

MatkovIvan (approved):

PR #994, opened 2024-10-08
Allow setting skiko properties using skiko.properties file inside jar

jakub-senohrabek-jb:
In order to use two different versions of skiko in one app (using multiple classloaders), it is required to use the unpacking strategy, as the skiko.library.path is a system property, which is same for all classloaders. 

This change will allow to use presigned natives shipped together with the app, while still being able to use multiple versions as the property skiko.library.path will be loaded from current classloader resources, with fallback to system property.


skiko/src/jvmMain/kotlin/org/jetbrains/skiko/SkikoProperties.kt
```diff
@@ -89,6 +89,17 @@ object SkikoProperties {
 
     val macOsOpenGLEnabled: Boolean get() = getProperty("skiko.macos.opengl.enabled")?.toBoolean() ?: false
 
+    private fun getProperty(key: String): String? {
+        return classLoaderProperties?.getProperty(key) ?: System.getProperty(key)
+    }
+
+    private val classLoaderProperties by lazy {
+        val res = SkikoProperties::class.java.classLoader.getResourceAsStream("skiko.properties") ?: return@lazy null
```
Walingar, line 97: Is this `skiko.properties` something new or it was used somewhere before?

jakub-senohrabek-jb: It is new.

jakub-senohrabek-jb:
I have added flag to toggle this feature (skiko.resource.properties.enabled)

Walingar (approved):

skiko/src/jvmMain/kotlin/org/jetbrains/skiko/SkikoProperties.kt
```diff
@@ -89,6 +89,23 @@ object SkikoProperties {
...
+        }
+    }
+
+    private val classLoaderProperties by lazy {
+        val res = SkikoProperties::class.java.classLoader.getResourceAsStream("skiko.properties") ?: return@lazy null
+        val props = Properties()
+        props.load(res)
+        props
+    }
+
```
m-sasha, line 108: Maybe like this instead?
```
    private val properties = run {
        val resourcePropertiesEnabled = System.getProperty("skiko.resource.properties.enabled")?.toBoolean() ?: false
        val resources = if (resourcePropertiesEnabled) {
            SkikoProperties::class.java.classLoader.getResourceAsStream("skiko.properties")
        } else {
            null
        }
        val systemProps = System.getProperties()
        if (resources == null) systemProps else Properties(systemProps).apply { load(resources) }
    }

    private fun getProperty(key: String): String? = properties.getProperty(key)
```

skiko/src/jvmMain/kotlin/org/jetbrains/skiko/SkikoProperties.kt
```diff
@@ -89,6 +89,23 @@ object SkikoProperties {
 
     val macOsOpenGLEnabled: Boolean get() = getProperty("skiko.macos.opengl.enabled")?.toBoolean() ?: false
 
+    val resourcePropertiesEnabled: Boolean get() = System.getProperty("skiko.resource.properties.enabled")?.toBoolean() ?: false
```
m-sasha, line 92: Can be private.

skiko/src/jvmMain/kotlin/org/jetbrains/skiko/SkikoProperties.kt
```diff
@@ -89,6 +89,23 @@ object SkikoProperties {
...
+        return if (resourcePropertiesEnabled) {
+            classLoaderProperties?.getProperty(key) ?: System.getProperty(key)
+        } else {
+            System.getProperty(key)
+        }
+    }
+
+    private val classLoaderProperties by lazy {
+        val res = SkikoProperties::class.java.classLoader.getResourceAsStream("skiko.properties") ?: return@lazy null
+        val props = Properties()
```
m-sasha, line 104: Slightly more idiomatic, maybe? `Properties().apply { load(res) }`

igordmn (approved):

PR #995, opened 2024-10-08
Add a new saveLayer method to Canvas (already exists in Skia)

aibobrov:


SergeevPavel (approved):

skiko/src/commonMain/kotlin/org/jetbrains/skia/Canvas.kt
```diff
@@ -1349,6 +1349,36 @@ open class Canvas internal constructor(ptr: NativePointer, managed: Boolean, int
         }
     }
 
+    fun saveLayer(layerRec: SaveLayerRec): Int {
```
eymar, line 1352: Could you please add a simple common test calling this API (in https://github.com/JetBrains/skiko/blob/1bd3058d694ee22d2b8ddaa6219a88dcb153b768/skiko/src/commonTest/kotlin/org/jetbrains/skia/CanvasTest.kt)?

The test would simply call a new method (at least with the defaults, or better to check both 'if' branches) to check that all targets (k/jvm, k/native, k/js and k/wasm) do not crash when the method is called. 

And if you want to be sure the function would work as you expect, you might consider adding a screenshot test in awtTest  (other targets don't support screenshot tests for now). You would use `val screenshots = ScreenshotTestRule()`.

SergeevPavel: I'm not sure that screenshot test will work in this case because this limitation https://github.com/JetBrains/skiko/blob/1bd3058d694ee22d2b8ddaa6219a88dcb153b768/skiko/src/awtTest/kotlin/org/jetbrains/skiko/util/ScreenshotTestRule.kt#L13
though we may add an examples into sample apps

eymar (approved):

PR #996, opened 2024-10-17
Reuse the BufferedImage in SwingOffscreenRenderer

samofcorinth:
Fix for https://youtrack.jetbrains.com/issue/CMP-6722/Excessive-garbage-generation-from-redrawing

Previously, the `draw` method would allocate a `BufferedImage` on every redraw, which is wasteful. This fix changed `SwingOffscreenRenderer` to reuse the same instance of `BufferedImage` and only clear it on redraws. Only changing the draw size would allocate a new `BufferedImage`.

### DIFF
--- a/skiko/src/awtMain/kotlin/org/jetbrains/skiko/swing/SwingOffscreenDrawer.kt
+++ b/skiko/src/awtMain/kotlin/org/jetbrains/skiko/swing/SwingOffscreenDrawer.kt
@@ -59,11 +59,13 @@ internal class SwingOffscreenDrawer(
     ): BufferedImage {
         val src = ByteBuffer.wrap(bytes)
         if (bufferedImage == null || bufferedImage?.width != width || bufferedImage?.height != height) {
+            bufferedImage?.flush()
             bufferedImage = BufferedImage(width, height, BufferedImage.TYPE_INT_ARGB_PRE)
             bufferedImageGraphics = bufferedImage?.createGraphics()
+        } else {
+            bufferedImageGraphics?.clearRect(0,0, width, height)
         }
         val image = bufferedImage!!
-        bufferedImageGraphics?.clearRect(0,0, width, height)
 
         val dstData = (image.raster.dataBuffer as DataBufferInt).data
         val srcData: IntBuffer = src.order(ByteOrder.LITTLE_ENDIAN).asIntBuffer()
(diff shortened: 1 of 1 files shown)


skiko/src/awtMain/kotlin/org/jetbrains/skiko/swing/SwingOffscreenDrawer.kt
```diff
@@ -56,7 +58,13 @@ internal class SwingOffscreenDrawer(
         dirtyRectangles: List<Rectangle>
     ): BufferedImage {
         val src = ByteBuffer.wrap(bytes)
-        val image = BufferedImage(width, height, BufferedImage.TYPE_INT_ARGB_PRE)
+        if (bufferedImage == null || bufferedImage?.width != width || bufferedImage?.height != height) {
+            bufferedImage = BufferedImage(width, height, BufferedImage.TYPE_INT_ARGB_PRE)
+            bufferedImageGraphics = bufferedImage?.createGraphics()
+        }
+        val image = bufferedImage!!
+        bufferedImageGraphics?.clearRect(0,0, width, height)
```
MatkovIvan, line 66:
```suggestion
        } else {
            bufferedImageGraphics?.clearRect(0, 0, width, height)
        }
        val image = bufferedImage!!
```

It seems clearing is only required in case of re-use

skiko/src/awtMain/kotlin/org/jetbrains/skiko/swing/SwingOffscreenDrawer.kt
```diff
@@ -56,7 +58,13 @@ internal class SwingOffscreenDrawer(
         dirtyRectangles: List<Rectangle>
     ): BufferedImage {
         val src = ByteBuffer.wrap(bytes)
-        val image = BufferedImage(width, height, BufferedImage.TYPE_INT_ARGB_PRE)
+        if (bufferedImage == null || bufferedImage?.width != width || bufferedImage?.height != height) {
+            bufferedImage = BufferedImage(width, height, BufferedImage.TYPE_INT_ARGB_PRE)
```
m-sasha, line 62: Possible additional optimizations:
1. Call `flush()` on the old `bufferedImage` to clear the GPU memory without waiting for GC.
2. (Questionable) If the new size is smaller on both dimensions, you could reuse the `BufferedImage` by using `getSubimage()`.

m-sasha: > I'm not sure if optimization 2 is worth the extra code complexity, since a one-off allocation is (relatively) cheap.

You could have many allocations in a short period when the window is being resized. But using `getSubImage()` by itself won't solve it if the window is getting bigger. What you could do, though, is to allocate a slightly bigger image upfront (say 20% larger on each axis) and use `getSubImage()` on that.

rock3r: Sounds like a nice follow-up, but we can probably do with the fix as-is for now? Should cover most cases already causing issues.

m-sasha (approved):

PR #997, opened 2024-10-28
turn off flutter hack by default, implement methods to control it

SergeevPavel:


SergeevPavel:
@eymar the hack is basically truncation of some float values in paragraph e.g. `maxIntrinsicWidth`, as far as I understood it was required to make some Flutter tests pass. The reason why I decided to turn it off is demonstrated in this test https://github.com/JetBrains/skiko/blob/159c4fb8be076216fa4cd3f443a61a3d12d8d314/skiko/src/commonTest/kotlin/org/jetbrains/skia/ParagraphTest.kt#L184

The issue is observable with Compose `TextMeasurer`, [here](https://github.com/JetBrains/compose-multiplatform-core/blob/c45c867d66f4eae133a7adedf60060aa710f2bfc/compose/ui/ui-text/src/commonMain/kotlin/androidx/compose/ui/text/TextMeasurer.kt#L320) we layout paragraph with `maxIntrinicWidth` which may lead to unexpected line wraps with some combination of font size and line length

SergeevPavel:
Also could you take a look what's wrong with `WebTarget` tests , is the fault result of my changes?

skiko/src/commonTest/kotlin/org/jetbrains/skia/ParagraphTest.kt
```diff
@@ -172,4 +179,35 @@ class ParagraphTest {
             }
         }
     }
+
+    @Test
+    fun `layout paragraph with its maxIntrinsicWidth shouldn't lead to wraps`() = runTest {
```
eymar, line 184: Web CI pipeline failed because k/js and k/wasm don't like the test names in `...`.

So changing the name should help: 
```suggestion
    fun layout_paragraph_with_its_maxIntrinsicWidth_shouldnt_lead_to_wraps() = runTest {
```

skiko/src/commonTest/kotlin/org/jetbrains/skiko/paragraph/ParagraphStyleTests.kt
```diff
@@ -55,4 +55,13 @@ class ParagraphStyleTests {
             assertEquals(gloriousRasterSettings, paragraphStyle.fontRastrSettings)
         }
     }
+
+    @Test
+    fun paragraphStyleRoundingHackTests() {
+        ParagraphStyle().use { paragraphStyle ->
```
eymar, line 61: Changing this one to this would make k/js happy :D

```kotlin
    @Test
    fun paragraphStyleRoundingHackTests() {
        ParagraphStyle().use { paragraphStyle ->
            assertFalse(paragraphStyle.isApplyRoundingHackEnabled)
            paragraphStyle.isApplyRoundingHackEnabled = true
            assertTrue(paragraphStyle.isApplyRoundingHackEnabled)
        }
    }
```
`_nGetApplyRoundingHack` returns 0 (there is no false/true in wasm). And JS equality comparision fails:
`0 === false --> false` 
___

Not using "add a suggestion" because new imports are needed anyway.

eymar: jfyi: to run the k/js tests locally:

```
./gradlew :cleanJsBrowserTest :jsBrowserTest -Pskiko.wasm.enabled=true -Pskiko.js.enabled=true
```

SergeevPavel: Thanks!

eymar (approved):

PR #1002, opened 2024-12-03
Build skiko.wasm with -s SUPPORT_LONGJMP=wasm

eymar:


Schahen (approved):
thanks for taking care!

PR #1003, opened 2024-12-09
Fix high CPU usage when window is occluded on macOS Native

Thomas-Vos:
My app is used a lot in the background, and I noticed high CPU usage in the native macOS version. With this PR, CPU usage is reduced when window is invisible by disabling drawing. This was already implemented for the JVM, and that behaviour is now copied to native macOS. So the code is similar to JVM parts in:
- https://github.com/JetBrains/skiko/blob/master/skiko/src/awtMain/kotlin/org/jetbrains/skiko/redrawer/MetalRedrawer.kt
- https://github.com/JetBrains/skiko/blob/master/skiko/src/awtMain/objectiveC/macos/MetalRedrawer.mm

The draw function is now suspend, which required moving some other functions around. I tried to keep the implementation similar to the JVM version.

# Testing
Tested using `./gradlew runNative` in `samples/SkiaMultiplatformSample`. Now the CPU usage is reduced a lot when moving another (non transparent) app over the window.

# Old behaviour

https://github.com/user-attachments/assets/60ad3398-3f44-4e2f-aca7-44a4607d0840

# New behaviour with 300ms timeout
See reduction in CPU usage when window is invisible

https://github.com/user-attachments/assets/106efdeb-0044-4411-9b26-5a6b8886ff04

# New behaviour without 300ms timeout

https://github.com/user-attachments/assets/b986548b-d7ab-497b-9ddb-5b46ca016f83

### DIFF
--- a/skiko/src/macosMain/kotlin/org/jetbrains/skiko/redrawer/MetalRedrawer.macos.kt
+++ b/skiko/src/macosMain/kotlin/org/jetbrains/skiko/redrawer/MetalRedrawer.macos.kt
@@ -5,6 +5,8 @@ import kotlinx.cinterop.autoreleasepool
 import kotlinx.cinterop.objcPtr
 import kotlinx.cinterop.usePinned
 import kotlinx.cinterop.useContents
+import kotlinx.coroutines.channels.Channel
+import kotlinx.coroutines.withTimeoutOrNull
 import org.jetbrains.skia.BackendRenderTarget
 import org.jetbrains.skia.DirectContext
 import org.jetbrains.skiko.FrameDispatcher
@@ -12,6 +14,8 @@ import org.jetbrains.skiko.SkikoDispatchers
 import org.jetbrains.skiko.SkiaLayer
 import org.jetbrains.skiko.context.ContextHandler
 import org.jetbrains.skiko.context.MacOsMetalContextHandler
+import platform.AppKit.NSWindowDidChangeOcclusionStateNotification
+import platform.AppKit.NSWindowOcclusionStateVisible
 import platform.CoreGraphics.CGColorCreate
 import platform.CoreGraphics.CGColorSpaceCreateDeviceRGB
 import platform.CoreGraphics.CGContextRef
@@ -26,6 +30,11 @@ import platform.QuartzCore.kCALayerHeightSizable
 import platform.QuartzCore.kCALayerWidthSizable
 import kotlin.system.getTimeNanos
 import platform.CoreGraphics.CGSizeMake
+import platform.Foundation.NSNotification
+import platform.Foundation.NSNotificationCenter
+import platform.Foundation.NSOperationQueue
+import platform.darwin.NSObjectProtocol
+import kotlin.concurrent.Volatile
 
 /**
  * Metal [Redrawer] implementation for MacOs.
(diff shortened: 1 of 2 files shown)


skiko/src/macosMain/kotlin/org/jetbrains/skiko/redrawer/MetalRedrawer.macos.kt
```diff
@@ -115,17 +140,30 @@ internal class MacOsMetalRedrawer(
...
         autoreleasepool {
             if (!isDisposed) {
                 skiaLayer.update(getTimeNanos())
                 contextHandler.draw()
             }
         }
+
+        // When window is not visible - it doesn't make sense to redraw fast to avoid battery drain.
+        if (isWindowOccluded) {
+            withTimeoutOrNull(300) {
```
Thomas-Vos, line 161: I copied this 300 millisecond value from the JVM code to keep it consistent. However, I don't understand the reason for this delay, why not just suspend indefinitely (on native & jvm)? I can remove this timeout if needed.

Thomas-Vos: After some testing it seems all works fine without the 300ms timeout. CPU usage is reduced even more, on both native and JVM. Added a screenrecording with results for native app in the PR. I could not find any specific reason this timeout was added in git blame, so I removed it. If there's something I am missing please let me know.

What still stands out to me is that CPU and memory usage are both much lower for native vs JVM on my M1 Max macbook, with the same FPS result (100fps, my refresh rate). This is the also the case for my own much heavier app.

Thomas-Vos:
@igordmn it would be great if this could be reviewed. This PR improves performance for native macOS but also JVM macOS.

skiko/src/macosMain/kotlin/org/jetbrains/skiko/redrawer/MetalRedrawer.macos.kt
igordmn: Rendering each 300 ms was added before `windowOcclusionStateChannel.receive()` is introduced to avoid full CPU loading, because a window in occluded state isn't synchronised with the display refresh rate (we had 1000 FPS).

After `windowOcclusionStateChannel.receive()` was [added](https://github.com/JetBrains/skiko/pull/822/files#diff-93181abd2f66fb6545699899707e7808847d25a2221698ace0d9bf289caf7687L143), the 300 ms just wasn't removed.

Everything will work as before if an app rendering logic doesn't contain any side effects besides just changing visuals. A hypothetical side effect in Compose could be some action in response to a state change inside a window's composition:
- send a system notification
- show a modal dialog
- send analytics

It is not correct to write such actions inside a window composition, because they don't affect window's visuals. It is fine  if they are called outside of window composition (in `LaunchedEffect`, or in the `application` composition).

Because it is incorrect, we can do this change.

But we should add Release notes in the PR description:
```
### Release Notes
## Breaking changes - Desktop
- rendering on macOs is no longer executed if a window is occluded. Some applications can stop working correctly if they contain non window-related side effects inside rendering logic.
```
It will be reworded when we use it in Compose to:
```
### Release Notes
## Breaking changes - Desktop
- rendering on macOs is no longer executed if a window is occluded. Some applications can stop working correctly if they contain non window-related side effects inside composition/layout/draw
```

## Release
@m-sasha, could you review the JVM change as well? It looks fine by me, but if you have doubts, we can reduce the scope of the PR, by keeping only the macos native changes.

igordmn (approved):

skiko/src/macosMain/kotlin/org/jetbrains/skiko/redrawer/MetalRedrawer.macos.kt
Thomas-Vos: @igordmn thanks for the explanation! I added the release notes to the description.

m-sasha:
Seems that the change @igordmn mentioned is the only one here for JVM.

I think it's a bit risky, but I'm not opposed to it. One thing we should do though, is to run all Compose tests with this change, before merging. @Thomas-Vos Could you do that?

m-sasha:
Just FYI, I noticed a strange (existing) behavior in my app. Sometimes if I run it and immediately switch to a different window before the app's window is shown, it will never get shown. I haven't investigated it, though, so it may or may not be related.

Thomas-Vos:
@m-sasha rebased and I ran the tests on latest compose-multiplatform-core in :compose only for macosArm64, some tests were already failing (19) even without this PR. After including Skiko local build I also got 19 failed tests. Not sure what to do about this.

I could not find any information on how to set up the compose-multiplatform repo with a local build of compose-multiplatform-core and skiko. So unfortunately I was not able to test that repo.

If it's too risky I can take out the change and put it in a separate PR. Then at least the macOS Native changes could get merged.

Screenshot 2025-02-08 at 03 47 28

m-sasha:
> @m-sasha rebased and I ran the tests on latest compose-multiplatform-core in :compose only for macosArm64, some tests were already failing (19) even without this PR. After including Skiko local build I also got 19 failed tests. Not sure what to do about this.

@Thomas-Vos Let me run the tests myself first then.

m-sasha:
> If we are not sure and it requires tests, let's just revert the jvm change and add withTimeout to macOs too.

I think that's best. I'm not 100% convinced that occluded windows should not recompose anyway.

Thomas-Vos:
@igordmn reverted the change, thanks for the review.

igordmn:
Thanks!

Also removed Release Notes.

PR #1006, opened 2025-01-02
Set -mios-simulator-version-min=12.0 for skiko build

eymar:
This is an alternative (better) fix for https://youtrack.jetbrains.com/issue/CMP-6721 

Similar change is skia is required too: https://github.com/JetBrains/skia-pack/pull/60

### DIFF
--- a/skiko/buildSrc/src/main/kotlin/tasks/configuration/NativeTasksConfiguration.kt
+++ b/skiko/buildSrc/src/main/kotlin/tasks/configuration/NativeTasksConfiguration.kt
@@ -74,7 +74,7 @@ fun SkikoProjectContext.compileNativeBridgesTask(
                     Arch.Arm64 -> arrayOf(
                         "-target", if (isUikitSim) "arm64-apple-ios-simulator" else "arm64-apple-ios",
                         "-isysroot", if (isUikitSim) iphoneSimSdk else iphoneOsSdk,
-                        if (isUikitSim) "-mios-simulator-version-min=12.0" else "-miphoneos-version-min=12.0"
+                        if (isUikitSim) "-mios-simulator-version-min=12.0" else "-mios-version-min=12.0"
                     )
                     Arch.X64 -> arrayOf(
                         "-target", "x86_64-apple-ios-simulator",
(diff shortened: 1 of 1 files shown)


skiko/buildSrc/src/main/kotlin/tasks/configuration/NativeTasksConfiguration.kt
```diff
@@ -74,7 +74,7 @@ fun SkikoProjectContext.compileNativeBridgesTask(
                     Arch.Arm64 -> arrayOf(
                         "-target", if (isUikitSim) "arm64-apple-ios-simulator" else "arm64-apple-ios",
                         "-isysroot", if (isUikitSim) iphoneSimSdk else iphoneOsSdk,
-                        "-miphoneos-version-min=12.0"
```
MatkovIvan, line 77: It seems `iphoneos-version-min` is outdated alias to `ios-version-min`. Should we align that?

igordmn (approved):

skiko/buildSrc/src/main/kotlin/tasks/configuration/NativeTasksConfiguration.kt
eymar: Could you please point me to the docs about ios-version-min? Where did you see that it's outdated?

MatkovIvan: In fact, it's just alias. See https://clang.llvm.org/docs/ClangCommandLineReference.html:
> -mios-simulator-version-min=<arg>, -miphonesimulator-version-min=<arg>
-mios-version-min=<arg>, -miphoneos-version-min=<arg>
Set iOS deployment target

"iPhoneOS" is here for historical compatibility reasons.
See https://en.wikipedia.org/wiki/IOS_version_history
> iPhone OS was renamed iOS following the release of the iPad, starting with iOS 4

eymar: thanks. fixed

PR #1008, opened 2025-01-08
Fix macOS transparent windows on OpenJDK

MatkovIvan:
Fixes
https://youtrack.jetbrains.com/issue/CMP-7352
https://youtrack.jetbrains.com/issue/CMP-7345

### DIFF
--- a/skiko/src/awtMain/objectiveC/macos/DisplayLinkThrottler.mm
+++ b/skiko/src/awtMain/objectiveC/macos/DisplayLinkThrottler.mm
@@ -88,7 +88,14 @@ static CVReturn displayLinkCallback(CVDisplayLinkRef displayLink, const CVTimeSt
             [weakSelf onScreenDidChange];
         }];
 
-        [self onScreenDidChange];
+        if (NSThread.currentThread.isMainThread) {
+            [self onScreenDidChange];
+        } else {
+            // In case of OpenJDK, EDT thread != NSThread main thread
+            dispatch_async(dispatch_get_main_queue(), ^{
+                [self onScreenDidChange];
+            });
+        }
     }
 
     return self;
--- a/skiko/src/awtMain/objectiveC/macos/DisplayLinkThrottler.mm
+++ b/skiko/src/awtMain/objectiveC/macos/DisplayLinkThrottler.mm
@@ -92,8 +92,19 @@ static CVReturn displayLinkCallback(CVDisplayLinkRef displayLink, const CVTimeSt
             [self onScreenDidChange];
         } else {
             // In case of OpenJDK, EDT thread != NSThread main thread
+            // There is one thing we should keep in mind. Postponing creation of vsync throttler will cause that for sometime there won't be any waiting:
+            //
+            // ```
+            // create window
+            // (10x) schedule render -> render
+            // onScreenDidChange
+            // schedule render -> wait vsync -> render
+            // ```
+            //
+            // It is probably okay, but possible alternative would be to wait for throttler creation. It is difficult, so we can just keep the current code.
+            __weak DisplayLinkThrottler *weakSelf = self;
             dispatch_async(dispatch_get_main_queue(), ^{
-                [self onScreenDidChange];
+                [weakSelf onScreenDidChange];
             });
         }
     }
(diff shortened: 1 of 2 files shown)


elijah-semyonov (approved):

skiko/src/awtMain/objectiveC/macos/DisplayLinkThrottler.mm
```diff
@@ -88,7 +88,14 @@ - (instancetype)initWithWindow:(NSWindow *)window {
             [weakSelf onScreenDidChange];
         }];
 
-        [self onScreenDidChange];
+        if (NSThread.currentThread.isMainThread) {
+            [self onScreenDidChange];
+        } else {
+            // In case of OpenJDK, EDT thread != NSThread main thread
+            dispatch_async(dispatch_get_main_queue(), ^{
+                [self onScreenDidChange];
```
igordmn, line 96:
```suggestion
                [weakSelf onScreenDidChange];
```
In case the layer is created and right away disposed (there are cases when this can happen).

Note: I don't have knowledge of Objective C weak references, so I can't be sure if this is correct. The point is to not call the callback after the object is deallocated.

skiko/src/awtMain/objectiveC/macos/MetalRedrawer.mm
```diff
@@ -115,6 +115,24 @@ static jmethodID getOnOcclusionStateChangedMethodID(JNIEnv *env, jobject redrawe
...
+}
+
+static void setWindowProperties(NSWindow* window, jboolean transparency) {
+    if (NSThread.currentThread.isMainThread) {
+        setWindowPropertiesUnsafe(window, transparency);
+    } else {
+        // In case of OpenJDK, EDT thread != NSThread main thread
+        dispatch_sync(dispatch_get_main_queue(), ^{
+        setWindowPropertiesUnsafe(window, transparency);
+        });
```
igordmn, line 132:
```suggestion
        __weak NSWindow *weakWindow = self;
        dispatch_sync(dispatch_get_main_queue(), ^{
            NSWindow* callbackWindow = weakWindow;
            if (callbackWindow) {
                setWindowPropertiesUnsafe(callbackWindow, transparency);
            }
        });
```

skiko/src/awtMain/objectiveC/macos/DisplayLinkThrottler.mm
```diff
@@ -88,7 +88,14 @@ - (instancetype)initWithWindow:(NSWindow *)window {
             [weakSelf onScreenDidChange];
         }];
 
-        [self onScreenDidChange];
+        if (NSThread.currentThread.isMainThread) {
+            [self onScreenDidChange];
+        } else {
+            // In case of OpenJDK, EDT thread != NSThread main thread
```
igordmn, line 94:
```suggestion
            // In case of OpenJDK, EDT thread != NSThread main thread
            // There is one thing we should keep in mind. Postponing creation of vsync throttler will cause that for sometime there won't be any waiting:
            //
            // ```
            // create window
            // (10x) schedule render -> render
            // onScreenDidChange
            // schedule render -> wait vsync -> render
            // ```
            //
            // It is probably okay, but possible alternative would be to wait for throttler creation. It is difficult, so we can just keep the current code.
```

skiko/src/awtMain/objectiveC/macos/MetalRedrawer.mm
```diff
@@ -115,6 +115,24 @@ static jmethodID getOnOcclusionStateChangedMethodID(JNIEnv *env, jobject redrawe
     return onOcclusionStateChanged;
 }
 
+
+static void setWindowPropertiesUnsafe(NSWindow* window, jboolean transparency) {
+    if (transparency) {
+        window.hasShadow = NO;
```
igordmn, line 121: Can it cause a blink because of scheduling? I.e. shadow is visible a moment when users start the application.

I guess it shouldn't cause a blink if `window.isVisible = true` schedules showing to the main thread as well.

MatkovIvan: Didn't notice that. Anyway, there is no way to do it from the background thread

igordmn (approved):

skiko/src/awtMain/objectiveC/macos/MetalRedrawer.mm
elijah-semyonov: This block is dispatched synchronously, there is no way the window could be nil

igordmn: Hmm, indeed. I didn't noticed that it is `dispatch_sync` 🤔.

I think, we should change it to `dispatch_async`.

In the past we encountered a dead lock when we tried to wait for mac's main thread inside the EDT thread, because this can happen:

1. mac os fires an event inside a main thread
2. it schdules an AWT task, and waits for it

I can find fixes/issues to understand this deadlock, if needed.

igordmn (approved):

PR #1010, opened 2025-01-13
Avoid calling SkiaLayer.revalidate() when the window itself is moved.

m-sasha:
Calling `SkiaLayer.revalidate()` when the window moves causes unnecessary layout and repainting.
This PR makes sure to only call it when the actual position of the layer in the window changes.

### DIFF
--- a/skiko/src/awtMain/kotlin/org/jetbrains/skiko/SkiaLayer.awt.kt
+++ b/skiko/src/awtMain/kotlin/org/jetbrains/skiko/SkiaLayer.awt.kt
@@ -8,6 +8,7 @@ import org.jetbrains.skiko.redrawer.Redrawer
 import org.jetbrains.skiko.redrawer.RedrawerManager
 import java.awt.Color
 import java.awt.Component
+import java.awt.Point
 import java.awt.event.*
 import java.awt.geom.AffineTransform
 import java.awt.im.InputMethodRequests
@@ -20,7 +21,6 @@ import javax.swing.SwingUtilities.isEventDispatchThread
 import javax.swing.UIManager
 import javax.swing.event.AncestorEvent
 import javax.swing.event.AncestorListener
-import kotlin.math.ceil
 import kotlin.math.floor
 
 actual open class SkiaLayer internal constructor(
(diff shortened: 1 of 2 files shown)


m-sasha:
I still need to verify that it doesn't break what https://github.com/JetBrains/skiko/pull/988 was fixing.

m-sasha:
The following may be a suitable reproducer of the issue originally fixed in #988 
```
@OptIn(ExperimentalFoundationApi::class)
fun main() = SwingUtilities.invokeAndWait {
    JFrame("Compose Test").also {
        it.size = Dimension(800, 700)
        val panel = JPanel()
        it.contentPane.add(panel)
        it.contentPane.layout = null
        panel.setSize(400, 300)
        panel.setBounds(100, 100, 400, 500)
        val composePanel = ComposePanel()
        composePanel.setSize(400, 300)
        composePanel.apply {
            setContent {
                Column(Modifier.size(400.dp, 500.dp)) {
                    Box(
                        Modifier
                            .size(400.dp, 200.dp)
                            .background(Color.Red)
                            .onClick {
                                panel.location = panel.location.let { l -> Point(l.x + 20, l.y + 20) }
                                println("Moving")
                            }
                    )
                    Box(
                        Modifier
                            .size(400.dp, 200.dp)
                            .background(Color.Blue)
                            .onClick {
                                composePanel.size = composePanel.size.let { s -> Dimension(s.width + 1, s.height + 1) }
                                println("Resizing")
                            }
                    )
                }
            }
        }
        panel.add(composePanel)
        it.isVisible = true
    }
}
```
Clicking the red rectangle should move the whole thing by 20 pixels down and to the right. If we remove the `AncestorListener` from `SkiaLayer.awt.kt` altogether, then it doesn't move (clicking the blue rectangle causes it to redraw at the correct location).
With the version in this PR, it appears to work correctly when clicking the red part.

igordmn (approved):
Rechecked https://youtrack.jetbrains.com/issue/CMP-5856/Desktop-ComposePanel-size-breaks-with-.fillMax-modifiers#focus=Comments-27-10632441.0-0 - still doesn't reproduce

(after commenting `validate()` - it reproduces)

PR #1012, opened 2025-01-15
SkiaLayerAnalytics. Rendering logging

igordmn:
Fixes https://youtrack.jetbrains.com/issue/SKIKO-994/Add-logging-for-rendering

### DIFF
--- a/skiko/src/commonMain/kotlin/org/jetbrains/skiko/SkiaLayerAnalytics.kt
+++ b/skiko/src/commonMain/kotlin/org/jetbrains/skiko/SkiaLayerAnalytics.kt
@@ -42,6 +42,8 @@ interface SkiaLayerAnalytics {
         fun contextInit() = Unit
         fun beforeFirstFrameRender() = Unit
         fun afterFirstFrameRender() = Unit
+        fun beforeFrameRender() = Unit
+        fun afterFrameRender() = Unit
 
         companion object {
             val Empty = object : DeviceAnalytics {}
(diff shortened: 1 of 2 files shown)


skiko/src/commonMain/kotlin/org/jetbrains/skiko/SkiaLayerAnalytics.kt
```diff
@@ -42,6 +42,8 @@ interface SkiaLayerAnalytics {
         fun contextInit() = Unit
         fun beforeFirstFrameRender() = Unit
         fun afterFirstFrameRender() = Unit
+        fun beforeFrameRender() = Unit
+        fun afterFrameRender() = Unit
```
MatkovIvan, line 46: Common interface is triggered only on AWT

MatkovIvan (changes requested):

skiko/src/commonMain/kotlin/org/jetbrains/skiko/SkiaLayerAnalytics.kt
igordmn: There was no intention to provide it for non-jvm targets.

This would require a bigger refactoring.

This PR is only extending what was before.

MatkovIvan (approved):

skiko/src/commonMain/kotlin/org/jetbrains/skiko/SkiaLayerAnalytics.kt
MatkovIvan: Let's track it somewhere

igordmn: I suggest to don't track until someone need it.

PR #1014, opened 2025-01-20
Adoption of `RenderNode` approach

MatkovIvan:
It is a simplified adoption of Android's `RenderNode`.

```kt
/**
 * <p>RenderNode is used to build hardware accelerated rendering hierarchies. Each RenderNode
 * contains both a display list as well as a set of properties that affect the rendering of the
 * display list. RenderNodes are used internally for all Views by default and are not typically
 * used directly.</p>
 *
 * <p>RenderNodes are used to divide up the rendering content of a complex scene into smaller
 * pieces that can then be updated individually more cheaply. Updating part of the scene only needs
 * to update the display list or properties of a small number of RenderNode instead of redrawing
 * everything from scratch. A RenderNode only needs its display list re-recorded when its content
 * alone should be changed. RenderNodes can also be transformed without re-recording the display
 * list through the transform properties.</p>
 ```
 
 This is a more correct approach to make `GraphicsLayer` in Compose invalidation independently.
 - Moved drawing callback to C++ side to avoid extra interop costs
 - Switched from `SkPicture` placeholder to custom `SkDrawable` implementation

### CODE REVIEW

File: skiko/src/commonMain/cpp/common/node/RenderNode.cpp [hunk 1]
```diff
@@ -0,0 +1,337 @@
...
+#include <SkShadowUtils.h>
+#include "node/RenderNode.h"
+#include "node/RenderNodeContext.h"
+
+namespace skiko {
+namespace node {
+
+// The goal with selecting the size of the rectangle here is to avoid limiting the
+// drawable area as much as possible.
+// Due to https://partnerissuetracker.corp.google.com/issues/324465764 we have to
```
igordmn, line 12:
```suggestion
// Due to https://issuetracker.google.com/issues/324465764 we have to
```

skiko/src/commonMain/cpp/common/node/RenderNode.cpp
```diff
@@ -0,0 +1,337 @@
...
+            canvas->clipRRect(*this->clipRRect);
+        } else if (this->clipPath) {
+            canvas->clipPath(*this->clipPath);
+        } else {
+            canvas->clipRect(SkRect::MakeWH(this->bounds.width(), this->bounds.height()));
+        }
+    }
+
+    if (this->layerPaint) {
+        canvas->saveLayer(SkRect::MakeWH(this->bounds.width(), this->bounds.height()), &*this->layerPaint);
```
igordmn, line 239: Where are we  doing `canvas->restore()`?

skiko/src/commonMain/kotlin/org/jetbrains/skiko/node/RenderNodeContext.kt
```diff
@@ -0,0 +1,51 @@
...
+    }
+
+    constructor(
+        measureDrawBounds: Boolean = false,
+    ) : this(RenderNodeContext_nMake(measureDrawBounds)) {
+        Stats.onNativeCall()
+    }
+
+    fun setLightingInfo(
+        centerX: Float = Float.MIN_VALUE,
```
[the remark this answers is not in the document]
MatkovIvan: Same as in C++ Android sources. Doesn't matter much actually

skiko/src/commonMain/cpp/common/node/RenderNode.cpp
MatkovIvan: Skia maintains this. `onDraw` is wrapped to auto restoring primitives.

Decided to remove duplication on our side after looking at skia's sources

igordmn (approved):

PR #1016, opened 2025-01-27
DirectXOffscreenContext to allow drawing into offscreen texture

igordmn:
Needed to run benchmarks without vsync interference.

The code extracted from Direct3DSwingRedrawer with minimal changes.

- implementation for other API will be added when it is needed (Metal is considered later)
- there are no plans to make commonMain abstraction in `skiko` as it is not needed at the moment

## Testing
- manually in benchmarks with this code:
```
@OptIn(ExperimentalSkikoApi::class)
class DirectXGraphicsContext() : DesktopGraphicsContext {
    private val context = DirectXOffscreenContext()
    private var texture: DirectXOffscreenContext.Texture? = null

    override fun surface(width: Int, height: Int): Surface {
        texture?.close()
        texture = context.Texture(width, height)
        return Surface.makeFromBackendRenderTarget(
            context.directContext,
            texture!!.backendRenderTarget,
            SurfaceOrigin.TOP_LEFT,
            SurfaceColorFormat.BGRA_8888,
            ColorSpace.sRGB,
            SurfaceProps(pixelGeometry = PixelGeometry.UNKNOWN)
        ) ?: throw IllegalStateException("Can't create Surface")
    }

    override suspend fun awaitGPUCompletion() {
        texture?.waitForCompletion()
    }

    override fun close() {
        texture?.close()
        context.close()
    }
}

interface DesktopGraphicsContext : GraphicsContext, AutoCloseable
```
- new test for SkiaSwingLayer
- new test for DirectXOffscreenContext


skiko/src/awtMain/kotlin/org/jetbrains/skiko/swing/SkiaSwingLayer.kt
```diff
@@ -28,15 +28,14 @@ open class SkiaSwingLayer(
     renderDelegate: SkikoRenderDelegate,
     analytics: SkiaLayerAnalytics = SkiaLayerAnalytics.Empty,
     externalAccessibleFactory: ((Component) -> Accessible)? = null,
+    private val properties: SkiaLayerProperties = SkiaLayerProperties()
```
MatkovIvan, line 31: It seems unrelated + breaking compatibility for no reason. Can we avoid it in this PR?

igordmn: This is needed in the test - to pass the graphic API:
https://github.com/JetBrains/skiko/pull/1016/files#diff-f09f8e25b6e613b4acc42736872562fec343ba85cd033a84464e2ee426a1cbd7R188

As for compatibility, we don't support it in Skiko

MatkovIvan (approved):

PR #1017, opened 2025-01-30
ANGLE support

Alovchin91:
TBW

### DIFF
--- a/skiko/build-with-local-skia.sh
+++ b/skiko/build-with-local-skia.sh
@@ -43,7 +43,7 @@ case $SKIA_TARGET in
     ;;
   "windows")
     SKIKO_TARGET_FLAGS="-Pskiko.awt.enabled=true"
-    if [[ $(uname -m) == 'arm64' ]]; then
+    if [[ $(powershell.exe -command '(Get-CimInstance Win32_OperatingSystem).OSArchitecture') == *'ARM 64'* ]]; then
       skikoMachines=("arm64")
     else
       skikoMachines=("x64")
@@ -87,4 +87,6 @@ cd "$SCRIPT_DIR"
 
 rm -rf build/classes/kotlin/* # We need to drop old cache. We can do it with ./gradlew clean as well, but it tooks longer time to redownload dependencies dir.
 
-./gradlew publishToMavenLocal $SKIKO_TARGET_FLAGS -Pskia.dir="$(pwd)/skia-pack/skia" -Pskiko.debug=$SKIA_DEBUG_MODE
+for skikoMachine in ${skikoMachines[@]}; do
+  ./gradlew publishToMavenLocal $SKIKO_TARGET_FLAGS -Pskia.dir="${SCRIPT_DIR}/../../skia-pack/skia" -Pskiko.debug=$SKIA_DEBUG_MODE -Pskiko.arch="$skikoMachine"
+done
--- a/skiko/buildSrc/src/main/kotlin/properties.kt
+++ b/skiko/buildSrc/src/main/kotlin/properties.kt
@@ -43,13 +43,13 @@ fun compilerForTarget(os: OS, arch: Arch): String =
             Arch.Wasm -> "Unexpected combination: $os & $arch"
         }
         OS.Android -> "clang++"
-        OS.Windows -> "cl.exe"
+        OS.Windows -> "clang-cl.exe"
         OS.MacOS, OS.IOS, OS.TVOS -> "clang++"
         OS.Wasm -> "emcc"
     }
 
 fun linkerForTarget(os: OS, arch: Arch): String =
-    if (os.isWindows) "link.exe" else compilerForTarget(os, arch)
+    if (os.isWindows) "lld-link.exe" else compilerForTarget(os, arch)
 
 val OS.dynamicLibExt: String
     get() = when (this) {
--- a/skiko/buildSrc/src/main/kotlin/tasks/configuration/JvmTasksConfiguration.kt
+++ b/skiko/buildSrc/src/main/kotlin/tasks/configuration/JvmTasksConfiguration.kt
@@ -59,7 +59,8 @@ fun SkikoProjectContext.createCompileJvmBindingsTask(
     )
     sourceRoots.set(srcDirs)
     if (targetOs != OS.Android) includeHeadersNonRecursive(jdkHome.resolve("include"))
-    includeHeadersNonRecursive(skiaHeadersDirs(skiaJvmBindingsDir.get()))
+    val skiaDir = skiaJvmBindingsDir.get()
+    includeHeadersNonRecursive(skiaHeadersDirs(skiaDir))
     val projectDir = project.projectDir
     includeHeadersNonRecursive(projectDir.resolve("src/awtMain/cpp/include"))
     includeHeadersNonRecursive(projectDir.resolve("src/jvmMain/cpp/include"))
@@ -97,16 +98,16 @@ fun SkikoProjectContext.createCompileJvmBindingsTask(
         OS.Windows -> {
             includeHeadersNonRecursive(windowsSdkPaths.includeDirs)
             includeHeadersNonRecursive(jdkHome.resolve("include/win32"))
+            includeHeadersNonRecursive(skiaDir.resolve("third_party/externals/angle2/include"))
+            includeHeadersNonRecursive(skiaDir.resolve("include/gpu"))
+            includeHeadersNonRecursive(skiaDir.resolve("src/gpu"))
             osFlags = arrayOf(
                 "/nologo",
                 *buildType.msvcCompilerFlags,
                 "/utf-8",
                 "/GR-", // no-RTTI.
                 "/FS", // Due to an error when building in Teamcity. https://docs.microsoft.com/en-us/cpp/build/reference/fs-force-synchronous-pdb-writes
-                // LATER. Angle rendering arguments:
-                // "-I$skiaDir/third_party/externals/angle2/include",
-                // "-I$skiaDir/src/gpu",
-                // "-DSK_ANGLE",
+                "-DSK_ANGLE",
             )
         }
         OS.Android -> {
--- a/skiko/buildSrc/src/main/kotlin/tasks/configuration/JvmTasksConfiguration.kt
+++ b/skiko/buildSrc/src/main/kotlin/tasks/configuration/JvmTasksConfiguration.kt
@@ -107,7 +107,7 @@ fun SkikoProjectContext.createCompileJvmBindingsTask(
                 "/utf-8",
                 "/GR-", // no-RTTI.
                 "/FS", // Due to an error when building in Teamcity. https://docs.microsoft.com/en-us/cpp/build/reference/fs-force-synchronous-pdb-writes
-                "-DSK_ANGLE",
+                "/DSK_ANGLE",
             )
         }
         OS.Android -> {
@@ -317,6 +317,7 @@ fun SkikoProjectContext.createLinkJvmBindings(
                         "ole32.lib",
                         "Propsys.lib",
                         "shcore.lib",
+                        "Shlwapi.lib",
                         "user32.lib",
                     )
                 )
--- a/skiko/buildSrc/src/main/kotlin/tasks/configuration/JvmTasksConfiguration.kt
+++ b/skiko/buildSrc/src/main/kotlin/tasks/configuration/JvmTasksConfiguration.kt
@@ -95,7 +95,6 @@ fun SkikoProjectContext.createCompileJvmBindingsTask(
             )
         }
         OS.Windows -> {
-            compiler.set(windowsSdkPaths.compiler.absolutePath)
             includeHeadersNonRecursive(windowsSdkPaths.includeDirs)
             includeHeadersNonRecursive(jdkHome.resolve("include/win32"))
             osFlags = arrayOf(
@@ -104,7 +103,7 @@ fun SkikoProjectContext.createCompileJvmBindingsTask(
                 "/utf-8",
                 "/GR-", // no-RTTI.
                 "/FS", // Due to an error when building in Teamcity. https://docs.microsoft.com/en-us/cpp/build/reference/fs-force-synchronous-pdb-writes
-                // LATER. Ange rendering arguments:
+                // LATER. Angle rendering arguments:
                 // "-I$skiaDir/third_party/externals/angle2/include",
                 // "-I$skiaDir/src/gpu",
                 // "-DSK_ANGLE",
@@ -287,7 +286,6 @@ fun SkikoProjectContext.createLinkJvmBindings(
             )
         }
         OS.Windows -> {
-            linker.set(windowsSdkPaths.linker.absolutePath)
             libDirs.set(windowsSdkPaths.libDirs)
             osFlags = mutableListOf<String>().apply {
                 addAll(buildType.msvcLinkerFlags)
--- a/skiko/src/awtMain/cpp/windows/AngleRedrawer.cc
+++ b/skiko/src/awtMain/cpp/windows/AngleRedrawer.cc
@@ -1,6 +1,7 @@
 #ifdef SK_ANGLE
 
 #include <Windows.h>
+#include <dwmapi.h>
 #include <jawt_md.h>
 #include "jni_helpers.h"
 #include "exceptions_handler.h"
@@ -199,13 +200,14 @@ extern "C"
     }
 
     JNIEXPORT void JNICALL Java_org_jetbrains_skiko_redrawer_AngleRedrawerKt_finishFrame(
-        JNIEnv *env, jobject redrawer, jlong devicePtr, jboolean isVsyncEnabled)
+        JNIEnv *env, jobject redrawer, jlong devicePtr)
     {
         __try
         {
             AngleDevice *angleDevice = fromJavaPointer<AngleDevice *>(devicePtr);
             eglMakeCurrent(angleDevice->display, angleDevice->surface, angleDevice->surface, angleDevice->context);
-            eglSwapInterval(angleDevice->display, (int)isVsyncEnabled); // wait for vsync
+            // For vsync we will use dwmFlush instead of swapInterval, see WindowsOpenGLRedrawer.kt
+            eglSwapInterval(angleDevice->display, 0);
             if (!eglSwapBuffers(angleDevice->display, angleDevice->surface))
             {
                 throwAngleException(env, __FUNCTION__, "Could not complete eglSwapBuffers.");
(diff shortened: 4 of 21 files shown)


skiko/buildSrc/src/main/kotlin/tasks/configuration/JvmTasksConfiguration.kt
```diff
@@ -95,19 +96,18 @@ fun SkikoProjectContext.createCompileJvmBindingsTask(
             )
         }
         OS.Windows -> {
-            compiler.set(windowsSdkPaths.compiler.absolutePath)
             includeHeadersNonRecursive(windowsSdkPaths.includeDirs)
             includeHeadersNonRecursive(jdkHome.resolve("include/win32"))
+            includeHeadersNonRecursive(skiaDir.resolve("third_party/externals/angle2/include"))
```
Alovchin91, line 101: Looks like `skia-pack` should also include `angle2` headers in the archive:

```
> D:/a/skiko/skiko/skiko/src/awtMain/cpp/windows/AngleRedrawer.cc
  > (16,10): 
  > fatal error: 
  > 'ganesh/gl/GrGLDefines.h' file not found
```

Alovchin91:
Looks like `skia-pack` should also include `skia\src\gpu\ganesh\gl\GrGLDefines.h` header in the archive:

```
> D:/a/skiko/skiko/skiko/src/awtMain/cpp/windows/AngleRedrawer.cc
> (16,10): 
> fatal error: 
> 'ganesh/gl/GrGLDefines.h' file not found
```

skiko/buildSrc/src/main/kotlin/properties.kt
```diff
@@ -43,13 +43,13 @@ fun compilerForTarget(os: OS, arch: Arch): String =
             Arch.Wasm -> "Unexpected combination: $os & $arch"
         }
         OS.Android -> "clang++"
-        OS.Windows -> "cl.exe"
+        OS.Windows -> "clang-cl.exe"
```
igordmn, line 46: It seems we still use Visual C++ compiler on Windows?

igordmn, line 46: The clang change in Skiko is better to extract to a separate PR, so we can easily bisect/revert in case of issues

skiko/build-with-local-skia.sh
```diff
@@ -43,7 +43,7 @@ case $SKIA_TARGET in
     ;;
```
igordmn, line 43: It will be good if we can include Angle as a jar via `skiko-awt-runtime-angle-$platform`

igordmn, line 43: As we mentioned in Slack, the additional Angle size is 14 mb. The Windows's Skiko x64 dll is 16.6 mb, so having Angle almost the same is suspicious.

Maybe we should find the right arguments for the compiler/linker to reduce the size? For example, for stripping unused code.

igordmn: > ANGLE depends on d3dcompiler_47.dll which is 8,4 MB alone

It looks like we should link with it, not include it?

igordmn: Does the binary includes only Angle or Skiko+Angle?

igordmn: > I've managed to teach angle-pack to build an "official build" (it's quite a challenge), but it actually became a bit bigger, most likely due to optimisations favouring speed over size.

I assume it is a static library `angle.lib`? If we link it to a dynamic library `skiko.dll`, or `angle.dll`, maybe there is an unused code, and we can strip it?

Alovchin91: I'll see if I can figure this out

igordmn: > It looks like we should link with it, not include it?

Searched my PC, each app ships it separately :(.

P.S. x64 size is 4.7mb, arm64 size is 8.4

Alovchin91: > It looks like we should link with it, not include it?

No, it's a redistributable DLL from Microsoft, pretty much like vcruntime140.dll and stuff.

> Does the binary includes only Angle or Skiko+Angle?
> I assume it is a static library angle.lib? If we link it to a dynamic library skiko.dll, or angle.dll, maybe there is an unused code, and we can strip it?

It's ANGLE only, and those are 2 dynamic libraries: `libEGL.dll` and `libGLESv2.dll`. libEGL is small but libGLESv2 contains all the backend code.

Alovchin91: > P.S. x64 size is 4.7mb, arm64 size is 8.4

Indeed, I work on an Arm64 machine, so the sizes I get are for Arm64. `d3dcompiler_47.dll` is much bigger on Arm64 because it's an [ARM64X binary](https://learn.microsoft.com/en-us/windows/arm/arm64x-pe) (containing both native Arm64 code and x64-emulation-compatible Arm64EC code).

skiko/buildSrc/src/main/kotlin/properties.kt
igordmn: I see, missed that

skiko/build-with-local-skia.sh
Alovchin91: @igordmn You can find the pre-built artifacts here: https://github.com/Alovchin91/angle-pack/releases/tag/66bc9cfa00

Alovchin91: @igordmn By the way, I've noticed that at least `skiko-windows-arm64.dll` built with Clang-CL is _almost 7 MB smaller_ than the MSVC one, 10,5 MB vs. 17,2 MB for Skiko 0.8.22.1.

skiko/src/awtMain/kotlin/org/jetbrains/skiko/redrawer/AngleRedrawer.kt
```diff
@@ -0,0 +1,117 @@
...
+
+    private val adapterName get() = AngleApi.glGetString(AngleApi.GL_RENDERER)
+
+    init {
+        device = layer.backedLayer.useDrawingSurfacePlatformInfo { platformInfo ->
+            createAngleDevice(platformInfo, layer.transparency).takeIf { it != 0L }
+                ?: throw RenderException("Failed to create ANGLE device.")
+        }
+        adapterName.let { adapterName ->
+            if (adapterName == null || !isVideoCardSupported(GraphicsApi.ANGLE, hostOs, adapterName)) {
```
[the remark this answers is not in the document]
Alovchin91: Isn't it wrong to report a null adapter name in `onDeviceChosen` though? That call is required.

skiko/src/awtMain/cpp/windows/AngleSupport.cc
```diff
@@ -0,0 +1,103 @@
...
+        const char *content = reinterpret_cast<const char *>(glGetString(value));
+        return env->NewStringUTF(content);
+    }
+
+    JNIEXPORT void JNICALL Java_org_jetbrains_skiko_AngleSupport_1jvmKt_loadAngleLibraryWindows(JNIEnv *env, jobject obj) {
+        TCHAR basePath[MAX_PATH] = TEXT("");
+        TCHAR libEGL[MAX_PATH] = TEXT("");
+        TCHAR libGLESv2[MAX_PATH] = TEXT("");
+        HMODULE hmodule = nullptr;
+        THROW_IF_NULL(GetModuleHandleEx(GET_MODULE_HANDLE_EX_FLAG_FROM_ADDRESS | GET_MODULE_HANDLE_EX_FLAG_UNCHANGED_REFCOUNT,
```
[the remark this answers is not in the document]
Alovchin91: I'm not sure how to answer this. This code looks for the ANGLE libraries in the exact same directory as `skiko-windows-<ARCH>.dll` for security (because we don't want to load it from a random place in the system). Does Skiko DLL get extracted somewhere in the Gradle case?

skiko/src/jvmMain/kotlin/org/jetbrains/skiko/SkikoProperties.kt
```diff
@@ -139,8 +144,8 @@ object SkikoProperties {
             OS.MacOS -> listOf(GraphicsApi.METAL, GraphicsApi.SOFTWARE_COMPAT)
             OS.Windows -> when (hostArch) {
                 // Skia isn't properly tested on OpenGL and Windows ARM (https://groups.google.com/g/skia-discuss/c/McoclAhLpvg?pli=1)
-                Arch.Arm64 -> listOf(GraphicsApi.DIRECT3D, GraphicsApi.SOFTWARE_FAST, GraphicsApi.SOFTWARE_COMPAT)
-                else -> listOf(GraphicsApi.DIRECT3D, GraphicsApi.OPENGL, GraphicsApi.SOFTWARE_FAST, GraphicsApi.SOFTWARE_COMPAT)
+                Arch.Arm64 -> listOf(GraphicsApi.ANGLE, GraphicsApi.DIRECT3D, GraphicsApi.SOFTWARE_FAST, GraphicsApi.SOFTWARE_COMPAT)
```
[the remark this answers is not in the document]
Alovchin91: Not at the moment, right now we're trying to solve instability issues with the DirectX backend.

skiko/src/awtMain/kotlin/org/jetbrains/skiko/redrawer/AngleRedrawer.kt
```diff
@@ -0,0 +1,117 @@
+package org.jetbrains.skiko.redrawer
+
+import kotlinx.coroutines.withContext
+import org.jetbrains.skia.BackendRenderTarget
+import org.jetbrains.skia.DirectContext
+import org.jetbrains.skiko.*
+import org.jetbrains.skiko.context.AngleContextHandler
+
+internal class AngleRedrawer(
```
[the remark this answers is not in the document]
Alovchin91: I've mentioned above that I have a `D3DKMTWaitForVerticalBlankEvent` based implementation committed; if you're interested, we could try that one. Or we could go directly with `eglSwapInterval`, let me know what you'd prefer.

skiko/src/awtMain/cpp/windows/AngleSupport.cc
```diff
@@ -0,0 +1,103 @@
...
+                                        &hmodule));
+        THROW_IF_NULL(GetModuleFileName(hmodule, basePath, sizeof(basePath)));
+        THROW_IF_NULL(PathRemoveFileSpec(basePath));
+        THROW_IF_NULL(PathCombine(libEGL, basePath, TEXT("libEGL.dll")));
+        THROW_IF_NULL(PathCombine(libGLESv2, basePath, TEXT("libGLESv2.dll")));
+        THROW_IF_NULL(AngleEGLLibrary = LoadLibrary(libEGL));
+        THROW_IF_NULL(LoadLibrary(libGLESv2));
+    }
+}
+
```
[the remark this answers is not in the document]
Alovchin91: `d3dcompiler_47.dll` is a redistributable component. It might have been installed by some other application, but I'm not sure if it's available on Windows out of the box. Seems like it [might be](https://support.microsoft.com/en-us/topic/update-for-the-d3dcompiler-47-dll-component-on-windows-server-2012-windows-7-and-windows-server-2008-r2-769c6690-ed30-4dee-8bf8-dfa30e2f8088) but I wouldn't bet on it.

skiko/src/awtMain/kotlin/org/jetbrains/skiko/redrawer/AngleRedrawer.kt
```diff
@@ -0,0 +1,117 @@
...
+private external fun createAngleDevice(platformInfo: Long, transparency: Boolean): Long
+private external fun makeCurrent(device: Long)
+private external fun makeAngleContext(device: Long): Long
+private external fun makeAngleRenderTarget(device: Long, width: Int, height: Int): Long
+private external fun finishFrame(device: Long)
+private external fun dwmFlush()
+private external fun disposeDevice(device: Long)
+
+@Suppress("unused")
+class AngleRedrawerException(message: String) : RuntimeException(message)
```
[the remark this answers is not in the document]
Alovchin91: Good question. I think I can reuse it.

skiko/src/jvmMain/kotlin/org/jetbrains/skiko/SkikoProperties.kt
```diff
@@ -139,8 +144,8 @@ object SkikoProperties {
             OS.MacOS -> listOf(GraphicsApi.METAL, GraphicsApi.SOFTWARE_COMPAT)
             OS.Windows -> when (hostArch) {
                 // Skia isn't properly tested on OpenGL and Windows ARM (https://groups.google.com/g/skia-discuss/c/McoclAhLpvg?pli=1)
-                Arch.Arm64 -> listOf(GraphicsApi.DIRECT3D, GraphicsApi.SOFTWARE_FAST, GraphicsApi.SOFTWARE_COMPAT)
-                else -> listOf(GraphicsApi.DIRECT3D, GraphicsApi.OPENGL, GraphicsApi.SOFTWARE_FAST, GraphicsApi.SOFTWARE_COMPAT)
+                Arch.Arm64 -> listOf(GraphicsApi.DIRECT3D, GraphicsApi.ANGLE, GraphicsApi.SOFTWARE_FAST, GraphicsApi.SOFTWARE_COMPAT)
+                else -> listOf(GraphicsApi.DIRECT3D, GraphicsApi.ANGLE, GraphicsApi.OPENGL, GraphicsApi.SOFTWARE_FAST, GraphicsApi.SOFTWARE_COMPAT)
```
Alovchin91, line 148: @igordmn As an option, we can remove ANGLE from here completely for now.

skiko/src/awtMain/kotlin/org/jetbrains/skiko/redrawer/AngleRedrawer.kt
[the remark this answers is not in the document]
igordmn: `adapterName` is used only for reporting. It isn't used for any logic.

So it is better to not crash when something used only for logging fails.

Also it is better to have the same logic as in the other renders.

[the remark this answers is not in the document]
igordmn: I reverted this check, please look.

skiko/src/jvmMain/kotlin/org/jetbrains/skiko/SkikoProperties.kt
```diff
@@ -127,7 +132,7 @@ object SkikoProperties {
         when(hostOs) {
             OS.MacOS -> return GraphicsApi.METAL
             OS.Linux -> return GraphicsApi.OPENGL
-            OS.Windows -> return GraphicsApi.DIRECT3D
+            OS.Windows -> return GraphicsApi.ANGLE
```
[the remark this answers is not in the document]
igordmn: I made this renderer under `skiko.rendering.angle.enabled`. We still need it in the beginning of the queue, otherwise it won't be enabled.

skiko/src/awtMain/cpp/windows/AngleSupport.cc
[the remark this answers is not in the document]
igordmn: A new laptop on Windows 11 with all updates has it.

I would suggest to assume that it exists on the default configurations until there are reports that it isn't

skiko/src/awtMain/kotlin/org/jetbrains/skiko/redrawer/AngleRedrawer.kt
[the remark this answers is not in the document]
igordmn: Checked, they pass

skiko/build-with-local-skia.sh
igordmn: I have done this in the new branch. It also downloads the binaries from https://github.com/JetBrains/angle-pack/releases

skiko/src/awtMain/cpp/windows/AngleSupport.cc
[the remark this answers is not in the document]
igordmn: > Does Skiko DLL get extracted somewhere in the Gradle case

Yes, it has 3-way logic:
- look in the skiko.library.path
- look in the jar resources, unpack it to ~/.skiko/hash
- look in %java.home%

Besides, it checks if  a second class loader tries to load the same library and copies it if it does (needed for plugins that use different Skiko versions).

I refactored it in the branch by reusing the existed logic. Please look and verify if it works for your case (it should be loaded the same way as Skiko does).

skiko/src/awtMain/cpp/windows/AngleRedrawer.cc
```diff
@@ -0,0 +1,283 @@
...
+
+            if (transparency)
+            {
+                enableTransparentWindow(hwnd);
+            }
+
+            AngleDevice *angleDevice = new AngleDevice();
+            angleDevice->window = hwnd;
+            angleDevice->device = hdc;
+            angleDevice->display = getAngleEGLDisplay(angleDevice->device);
```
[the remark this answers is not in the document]
igordmn: I tested on a Windows laptop with integrated AMD card:
- it doesn't change FPS between 120hz and 165hz displays
- though, the DIRECT3D/OPENGL don't change it either (previously on a different machine it changed it correctly)
- checking it on one monitor and just changing the refresh rate in the system settings changes the FPS correctly

So, we can assume, that it probably behaves correctly and any deviation is probably system related

skiko/src/jvmMain/kotlin/org/jetbrains/skiko/SkikoProperties.kt
igordmn: We have to keep it here

skiko/src/awtMain/cpp/windows/AngleSupport.cc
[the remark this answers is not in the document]
igordmn: I implemented it in https://github.com/JetBrains/skiko/pull/1082, it rewrites the current PR logic and reuses the logic above

skiko/src/awtMain/cpp/windows/AngleSupport.cc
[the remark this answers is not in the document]
Alovchin91: Okay, per [this blog post](https://walbourn.github.io/not-so-direct-setup/) d3dcompiler_47.dll is included in Windows 8.1 and later. As long as we don't aim to support Windows 7, we should be good to go.

skiko/src/awtMain/cpp/windows/AngleRedrawer.cc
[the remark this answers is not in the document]
Alovchin91: I would probably open an issue to see how it behaves when the system switches between a discrete and an integrated GPU, and also to make sure the window created by Skiko has the `CS_OWNDC` flag set.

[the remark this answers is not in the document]
igordmn: https://youtrack.jetbrains.com/projects/SKIKO/issues/SKIKO-1043/Investigate-behavior-of-switching-to-integrated-discrete-GPU-during-application-running

https://youtrack.jetbrains.com/issue/SKIKO-1044/ANGLE-renderer.-Support-skiko.gpu.priority

PR #1019, opened 2025-02-03
Implement SkImages::AdoptTextureFrom bindings

Luna0x01:
I've tested it and it works completely fine under OpenGL.

### DIFF
--- a/skiko/src/commonMain/kotlin/org/jetbrains/skia/Image.kt
+++ b/skiko/src/commonMain/kotlin/org/jetbrains/skia/Image.kt
@@ -138,6 +138,36 @@ class Image internal constructor(ptr: NativePointer) : RefCnt(ptr), IHasImageInf
             return Image(ptr)
         }
 
+        /** Creates GPU-backed [org.jetbrains.skia.Image] from backendTexture associated with context.
+        Skia will assume ownership of the resource and will release it when no longer needed.
+        A non-null SkImage is returned if format of backendTexture is recognized and supported.
+        Recognized formats vary by GPU backend.
+         */
+        fun adoptTextureFrom(
+            context: DirectContext,
+            textureId: Int,
+            target: Int,
+            width: Int,
+            height: Int,
+            format: Int,
+            origin: SurfaceOrigin = SurfaceOrigin.TOP_LEFT,
+            colorType: ColorType = ColorType.RGBA_8888,
+        ): Image {
+            Stats.onNativeCall()
+            val ptr = _nAdoptTextureFrom(
+                context._ptr,
+                textureId,
+                target,
+                width,
+                height,
+                format,
+                origin.ordinal,
+                colorType.ordinal,
+            )
+            require(ptr != NullPointer) { "Failed to Image::adoptTextureFrom" }
+            return Image(ptr)
+        }
+
         init {
             staticLoad()
         }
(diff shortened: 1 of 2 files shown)


MatkovIvan (changes requested):
Aside from JNI, it should contain K/N bindings to work outside JVM too

Luna0x01:
I've implemented the requested changes.

skiko/src/nativeJsMain/cpp/Image.cc
```diff
@@ -1,14 +1,45 @@
...
+
+    GrGLTextureInfo textureInfo;
+    textureInfo.fID = static_cast<GrGLuint>(textureId);
+    textureInfo.fTarget = static_cast<GrGLenum>(target);
+    textureInfo.fFormat = static_cast<GrGLenum>(format);
+
+    GrBackendTexture backendTexture = GrBackendTextures::MakeGL(
+        width,
+        height,
+        skgpu::Mipmapped::kYes,
```
MatkovIvan, line 30: Such a thing shouldn't be hardcoded

skiko/src/nativeJsMain/cpp/Image.cc
```diff
@@ -1,14 +1,45 @@
...
+SKIKO_EXPORT KNativePointer org_jetbrains_skia_Image__1nAdoptTextureFrom
+  (KNativePointer contextPtr, KInt textureId, KInt target, KInt width, KInt height, KInt format, KInt surfaceOrigin, KInt colorType) {
+    GrDirectContext* context = reinterpret_cast<GrDirectContext*>(contextPtr);
+
+    GrGLTextureInfo textureInfo;
+    textureInfo.fID = static_cast<GrGLuint>(textureId);
+    textureInfo.fTarget = static_cast<GrGLenum>(target);
+    textureInfo.fFormat = static_cast<GrGLenum>(format);
+
+    GrBackendTexture backendTexture = GrBackendTextures::MakeGL(
```
MatkovIvan, line 27: It seems that we need to expose this separately and pass it as an argument

MatkovIvan (commented):
Thanks for implementing the missing part.

Currently I'm not sure that creating a new `GrBackendTexture` should be a part of `Image` API, but I'm OK to leave this part as TODO for later.

PS I'll be on vacation next week so cc @igordmn for taking care of it

Luna0x01:
@igordmn Any idea when this will be reviewed?

skiko/src/nativeJsMain/cpp/Image.cc
igordmn: Let's provide it as a parameter (boolean suffice if it has 2 options)

skiko/src/nativeJsMain/cpp/Image.cc
igordmn: It is a GL object that is used in a non-GL method. Because of that we have to move it out of this method and create a real `BackendTexture` kotlin class with `BackendTexture.MakeGL` function.

skiko/src/commonMain/kotlin/org/jetbrains/skia/Image.kt
```diff
@@ -138,6 +138,36 @@ class Image internal constructor(ptr: NativePointer) : RefCnt(ptr), IHasImageInf
             return Image(ptr)
         }
 
+        /** Creates GPU-backed [org.jetbrains.skia.Image] from backendTexture associated with context.
```
igordmn, line 141: - Please use KDoc format of the comment:
```
/**
 *
 *
 */
```
- describe the parameters as in the original documentation in Skia

igordmn (changes requested):

PR #1020, opened 2025-02-03
Build Skiko with Clang-CL

Alovchin91:
Official Skia documentation suggests that it's [_highly recommended_](https://skia.org/docs/user/build/#highly-recommended-build-with-clang-cl) to build Skia with Clang-CL on Windows, and that this dramatically improves Skia performance with Software rendering and in other areas. This corresponds with my experience, so here it is.

Related Skia-pack PR: https://github.com/JetBrains/skia-pack/pull/64

Prerequisite: #1024

### DIFF
--- a/skiko/buildSrc/src/main/kotlin/properties.kt
+++ b/skiko/buildSrc/src/main/kotlin/properties.kt
@@ -43,13 +43,13 @@ fun compilerForTarget(os: OS, arch: Arch): String =
             Arch.Wasm -> "Unexpected combination: $os & $arch"
         }
         OS.Android -> "clang++"
-        OS.Windows -> "cl.exe"
+        OS.Windows -> "clang-cl.exe"
         OS.MacOS, OS.IOS, OS.TVOS -> "clang++"
         OS.Wasm -> "emcc"
     }
 
 fun linkerForTarget(os: OS, arch: Arch): String =
-    if (os.isWindows) "link.exe" else compilerForTarget(os, arch)
+    if (os.isWindows) "lld-link.exe" else compilerForTarget(os, arch)
 
 val OS.dynamicLibExt: String
     get() = when (this) {
@@ -74,22 +74,22 @@ enum class SkiaBuildType(
     val id: String,
     val flags: Array<String>,
     val clangFlags: Array<String>,
-    val msvcCompilerFlags: Array<String>,
-    val msvcLinkerFlags: Array<String>
+    val winCompilerFlags: Array<String>,
+    val winLinkerFlags: Array<String>
 ) {
     DEBUG(
-        "Debug",
+        id = "Debug",
         flags = arrayOf("-DSK_DEBUG"),
         clangFlags = arrayOf("-std=c++17", "-g", "-DSK_TRIVIAL_ABI=[[clang::trivial_abi]]"),
-        msvcCompilerFlags = arrayOf("/Zi /std:c++17"),
-        msvcLinkerFlags = arrayOf("/DEBUG"),
+        winCompilerFlags = arrayOf("/Zi", "/std:c++17"),
+        winLinkerFlags = arrayOf("/DEBUG"),
     ),
     RELEASE(
         id = "Release",
         flags = arrayOf("-DNDEBUG"),
         clangFlags = arrayOf("-std=c++17", "-O3"),
-        msvcCompilerFlags = arrayOf("/O2 /std:c++17"),
-        msvcLinkerFlags = arrayOf("/DEBUG"),
+        winCompilerFlags = arrayOf("/O2", "/std:c++17"),
+        winLinkerFlags = arrayOf("/DEBUG"),
     );
     override fun toString() = id
 }
--- a/skiko/src/awtMain/cpp/windows/directXRedrawer.cc
+++ b/skiko/src/awtMain/cpp/windows/directXRedrawer.cc
@@ -11,7 +11,7 @@
 #include "ganesh/GrDirectContext.h"
 #include "SkSurface.h"
 #include "include/gpu/ganesh/SkSurfaceGanesh.h"
-#include "../common/interop.hh"
+#include "interop.hh"
 #include "DCompLibrary.h"
 
 #include "ganesh/d3d/GrD3DTypes.h"
--- a/DEVELOPMENT.md
+++ b/DEVELOPMENT.md
@@ -9,14 +9,15 @@
   * `Windows`
     1. Download [Visual Studio Build Tools 2019](https://learn.microsoft.com/en-us/visualstudio/releases/2019/history) (search "BuildTools" on the page).
     2. During the installation, select "Desktop development with C++"
-    3. Add an environment variable SKIKO_VSBT_PATH=C:/Program Files (x86)/Microsoft Visual Studio/2019/BuildTools
-    ```
-    Control Panel|All Control Panel Items|System|Advanced system settings|Environment variables
-    ```
-    or by running `cmd` as administrator:
-    ```
-    setx /M SKIKO_VSBT_PATH "C:\Program Files (x86)\Microsoft Visual Studio\2019\BuildTools"
-    ```
+    3. Add an environment variable SKIKO_VSBT_PATH=C:\Program Files (x86)\Microsoft Visual Studio\2019\BuildTools
+       ```
+       Control Panel|All Control Panel Items|System|Advanced system settings|Environment variables
+       ```
+       or by running `cmd` as administrator:
+       ```
+       setx /M SKIKO_VSBT_PATH "C:\Program Files (x86)\Microsoft Visual Studio\2019\BuildTools"
+       ```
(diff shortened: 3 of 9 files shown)


skiko/buildSrc/src/main/kotlin/tasks/configuration/CommonTasksConfiguration.kt
```diff
@@ -90,7 +90,8 @@ fun skiaPreprocessorFlags(os: OS, buildType: SkiaBuildType): Array<String> {
             "-DWIN32_LEAN_AND_MEAN",
             "-DNOMINMAX",
             "-DSK_GAMMA_APPLY_TO_A8",
-            "-DSK_DIRECT3D"
+            "-DSK_DIRECT3D",
+            "-DSK_ANGLE"
```
Alovchin91, line 94: Remove from this branch

skiko/src/commonMain/kotlin/org/jetbrains/skia/BreakIterator.kt
```diff
@@ -305,14 +305,6 @@ class BreakIterator internal constructor(ptr: NativePointer) : Managed(ptr, _Fin
...
     }
 
-    /**
-     * Create a copy of this iterator
-     */
-    fun clone(): BreakIterator {
-        Stats.onNativeCall()
-        return BreakIterator(withErrorGuard("Failed to clone") { _nClone(_ptr, it) })
-    }
-
```
Alovchin91, line 315: I reckon this change would probably require bumping Skiko version to 0.9 🤔

skiko/buildSrc/src/main/kotlin/properties.kt
```diff
@@ -78,17 +78,17 @@ enum class SkiaBuildType(
...
-        msvcCompilerFlags = arrayOf("/Zi /std:c++17"),
+        msvcCompilerFlags = arrayOf("/Zi", "/std:c++17"),
         msvcLinkerFlags = arrayOf("/DEBUG"),
     ),
     RELEASE(
         id = "Release",
         flags = arrayOf("-DNDEBUG"),
         clangFlags = arrayOf("-std=c++17", "-O3"),
-        msvcCompilerFlags = arrayOf("/O2 /std:c++17"),
+        msvcCompilerFlags = arrayOf("/O2", "/std:c++17"),
```
igordmn, line 91: Let's remove msvc if it is not used now (from Development.md as well)

skiko/src/commonMain/kotlin/org/jetbrains/skia/BreakIterator.kt
igordmn: Okay, we'll build the next version as 0.9.0

igordmn (approved):

igordmn:
`Clang` gives a big increase in performance (~ 16% FPS from 2 runs on average) on Windows x64 / DirectX. Testing to post more precise results with different conditions. Software rendering will be more faster.

skiko/build-with-local-skia.sh
```diff
@@ -43,7 +43,7 @@ case $SKIA_TARGET in
     ;;
```
igordmn, line 43: Let's copy this
https://github.com/JetBrains/skia-pack/blob/8772c12c1d2881f9e663b0dfc8bc264a08a744aa/README.md#windows-specific

To DEVELOPMENT.md requirements

skiko/buildSrc/src/main/kotlin/properties.kt
Alovchin91: In fact, we still use MSVC flags, because on Windows, we use `clang-cl` and `lld-link` which are MSVC compatible compiler (cl) and linker (link) frontends, respectively. They understand the Windows .lib and .obj format, and they also use MSVC-like flags. We can rename these to something like `winCompilerFlags` and `winLinkerFlags` to make it less confusing.

skiko/build-with-local-skia.sh
Alovchin91: Done!

skiko/buildSrc/src/main/kotlin/properties.kt
igordmn: Sounds good.

Also, let's add TODO:
```
// TODO consider reusing `clangFlags` and test that it behaves the same or better:
//  - different optimisation levels (O3 instead O2 in Release)
//  - different Debug options
```

Alovchin91: I'm not entirely sure clang-cl supports the same flags as clang:

> To be compatible with cl.exe, clang-cl supports most of the same command-line options. Those options can start with either / or -. It also supports _some_ of Clang’s core options, such as the -W options.

I can still leave the TODO though. Theoretically the `/clang:` option can be used to pass Clang options unchanged. And the equivalent of Clang's `-O3` is `/Ot`, according to [the documentation](https://clang.llvm.org/docs/UsersManual.html#id12).

Alovchin91: Also note that I have [disabled the trivial ABI in Skia-pack](https://github.com/JetBrains/skia-pack/blob/3258ba9f76930af325f690dbb232261b0dbc3ffb/script/build.py#L95) to keep it compatible with MSVC, so if we decide to re-use Clang debug flags, we would have to remove that line from Skia-pack too.

skiko/src/commonMain/kotlin/org/jetbrains/skia/BreakIterator.kt
igordmn: Explanation why it is deleted in an internal channel: https://jetbrains.slack.com/archives/C02DDNREC77/p1738678000232729

- `breakIteratorCloneTest` is failing with an access violation in Debug mode with additional checks, so using it wasn't safe
- it is not used anywhery

skiko/buildSrc/src/main/kotlin/properties.kt
igordmn: Ok, let's keep without todo

igordmn:
[Benchmarks results](https://docs.google.com/spreadsheets/d/1ldcwBI80rqr49TIRaW0dqzaacBvZnC6S6yhz_pYXdCw/edit?gid=45262332#gid=45262332)

Binary size **27.38%** is smaller
DirectX rendering FPS  is **16.63%** faster
Software rendering FPS is **72.85%** faster

Schahen (changes requested):
I'd rather have two separate PRs but don't have any strong objections apart of changing the logic behind build-with-local-skia.sh

skiko/build-with-local-skia.sh
```diff
@@ -87,4 +87,6 @@ cd "$SCRIPT_DIR"
 
 rm -rf build/classes/kotlin/* # We need to drop old cache. We can do it with ./gradlew clean as well, but it tooks longer time to redownload dependencies dir.
 
-./gradlew publishToMavenLocal $SKIKO_TARGET_FLAGS -Pskia.dir="$(pwd)/skia-pack/skia" -Pskiko.debug=$SKIA_DEBUG_MODE
+for skikoMachine in ${skikoMachines[@]}; do
```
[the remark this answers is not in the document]
Alovchin91: It wasn't respecting the machine chosen by the logic above so it could fail because of discrepancy between the skia-pack and the current machine's architecture. But I'll drop this commit anyway.

skiko/src/commonMain/kotlin/org/jetbrains/skia/BreakIterator.kt
Alovchin91: I split this change off into a separate PR: #1024

Alovchin91:
> I'd rather have two separate PRs

I've split the BreakIterator part off into a separate PR: #1024 

It would be a prerequisite for this one.

skiko/src/awtMain/cpp/windows/directXRedrawer.cc
```diff
@@ -11,7 +11,7 @@
 #include "ganesh/GrDirectContext.h"
 #include "SkSurface.h"
 #include "include/gpu/ganesh/SkSurfaceGanesh.h"
-#include "../common/interop.hh"
+#include "../../../jvmMain/cpp/common/interop.hh"
```
MatkovIvan, line 14: It doesn't look right. Things like `../../../jvmMain/` shouldn't be a part of sources

Could you please add `skiko/src/jvmMain/cpp/common` to header search paths instead?

Alovchin91: Done! I've squashed it with the first commit.

MatkovIvan (approved):

igordmn (approved):

Schahen (approved):
Thanks for taking care of it!

igordmn:
I rechecked measurements of the Clang change, and realized that I wrongly calculated percentages.
I recalculated just by dividing FPS now, and the results are more [impressive](https://github.com/JetBrains/skiko/pull/1020#issuecomment-2649433785).

PR #1021, opened 2025-02-05
Build skiko with skia@m132

Schahen:


skiko/src/nativeJsMain/cpp/ImageFilter.cc
```diff
@@ -6,6 +6,8 @@
 #include "SkRect.h"
 #include "common.h"
 
+using SkRuntimeShaderBuilder = SkRuntimeEffectBuilder;
```
eymar, line 9: In Jvm bindings it was replaced directly to SkRuntimeEffectBuilder. 
Is it possible to replace here too, for consistency?

eymar (commented):
Looks good to me.
I think the failed CI on some targets is due to outdated imports?

skiko/src/nativeJsMain/cpp/ImageFilter.cc
Schahen: @eymar sure, it just that introducing alias was minimal intrusion but you are right - it's better to be homomorphic in that aspect, so let me do it

PR #1022, opened 2025-02-11
Pre-cache SkPicture snapshots for RenderNode

MatkovIvan:
[SKIKO-1000](https://youtrack.jetbrains.com/issue/SKIKO-1000) Pre-cache SkPicture snapshots for RenderNode

### DIFF
--- a/skiko/src/commonMain/cpp/common/include/node/RenderNode.h
+++ b/skiko/src/commonMain/cpp/common/include/node/RenderNode.h
@@ -1,5 +1,6 @@
 #pragma once
 #include <optional>
+#include <set>
 #include <SkBBHFactory.h>
 #include <SkCamera.h>
 #include <SkCanvas.h>
@@ -68,6 +69,8 @@ public:
 
     void drawInto(SkCanvas *canvas);
 
+    SkFlattenable::Type getFlattenableType() const override;
+
 protected:
     // SkDrawable
     SkRect onGetBounds() override;
@@ -76,7 +79,34 @@ protected:
     sk_sp<SkPicture> onMakePictureSnapshot() override;
 
 private:
+    // What changed about a node, as far as snapshot bookkeeping is concerned.
+    enum SnapshotChange : uint32_t {
+        // Drawing applied around the snapshot: transform, clip, layer, alpha. This
+        // node's own snapshot stays valid; the snapshots that inlined it do not.
+        kAppearance = 1 << 0,
+        // The recorded content, or the cull rect it is recorded with.
+        kContent = 1 << 1,
+        // The set of drawn nodes, or what they draw, may differ, so re-derive from
+        // the current dependencies whether this content can be snapshotted.
+        kDependencies = 1 << 2,
+    };
+
+    SkCanvas *beginRecordingInto(SkPictureRecorder& target);
+    // Whether the content may cover more than the node's own bounds, and so has to be
+    // measured while recording instead of being culled to them.
+    bool mayDrawOutOfBounds() const;
+    // Whether this node's own drawing only replays correctly at the transform it was
+    // recorded under, which bars every node that inlines it from snapshotting too.
+    bool drawsTransformDependentContent() const;
+    // Whether an observer that inlines this node cannot snapshot the result: true when this
+    // node draws a shadow, or when anything in its own content does. This is the single bit
+    // an observer reads from each of its dependencies.
+    bool preventsObserverSnapshot() const;
     void updateMatrix();
+    void updateDependencies();
+    void releaseDependencies();
+    void updateSnapshot();
+    void invalidateSnapshot(uint32_t changes);
     void drawShadow(SkCanvas *canvas, const LightGeometry& lightGeometry, const LightInfo& lightInfo);
     void setCameraLocation(float x, float y, float z);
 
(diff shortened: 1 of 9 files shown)


igordmn:
> need to investigate performance regression in part of our benchmark (another part shows the improvement)

What is the result of investigation? Do we have this regression now?

MatkovIvan:
> What is the result of investigation?

Not proven yet, but there were a few problems with old version that I've fixed now:
- missing "unref"/leak - growing walking tree with zombies
- over-invalidation - fixed by `SnapshotChange` to keep content's snapshot in more cases
- unconditional cacheing - now it's skipped for shadows and alpha modulation

> Do we have this regression now?

I don't have final numbers yet, and I want to run it not only locally. That's why I've added a feature flag disabled by default.

skiko/src/commonMain/cpp/common/include/node/RenderNode.h
```diff
@@ -85,6 +115,20 @@ class RenderNode : public SkDrawable {
...
+    // position relative to a light fixed in device space, so replaying the recording
+    // anywhere else yields a different shadow.
+    bool contentTransformInvariant;
+    // Whether the observers have already been told, since this node was last drawn, that
+    // their inlined copy of it is stale. Lets a node reached through several observer paths
+    // propagate once instead of once per path. Cleared when the node is drawn (they inline
+    // it afresh) and when preventsObserverSnapshot() flips (they must re-derive).
+    bool observersNotified;
+
+    std::set<RenderNode *> dependencies, observers;
```
hub-bla: AFAIR, sets in C++ are ordered. If we do not care about the order we might want to use `unordered_set`

skiko/src/commonMain/cpp/common/node/RenderNode.cpp
```diff
@@ -76,11 +83,39 @@ class AlphaModulateFilterCanvas : public SkPaintFilterCanvas {
...
+            if (dependencies->insert(renderNode).second) {
+                renderNode->ref();
+            }
+        } else {
+            drawable->draw(this, matrix);
+        }
+    }
+
+private:
+    std::set<RenderNode *> *dependencies;
```
hub-bla: Can be a reference to a set instead of creating a new pointer to it

hub-bla (changes requested):

skiko/src/commonMain/cpp/common/node/RenderNode.cpp
MatkovIvan: What's the problem with creating a pointer? It's literally a reference (address) to the structure without any allocations/copies

MatkovIvan: Fixed

skiko/src/commonMain/cpp/common/include/node/RenderNode.h
MatkovIvan: Fixed

igordmn:
Benchmarks results on Windows (this commit with `snapshotCache = true` vs the previous commit):
```
MultipleComponents - 3% faster
VisualEffects - 32% faster
LazyGrid-ItemLaunchedEffect, LazyGrid - 6% slower
LazyGrid-SmoothScroll-ItemLaunchedEffect - 3% faster
```
The rest of the benchmarks are within +-2% diff

MatkovIvan:
Benchmarks results on macOS

| Benchmark | JVM (CPU) | macOS native |
|---|---|---|
| **VisualEffects** | **−86.0%** | **−32.7%** |
| **MultipleComponents-NoVectorGraphics** | **−43.2%** | **−9.0%** |
| **MultipleComponents** | **−36.8%** | **−6.6%** |
| LazyList | −9.1% | −3.7% |
| LazyGrid-SmoothScroll | −5.1% | −1.3% |
| LazyGrid-SmoothScroll-ItemLaunchedEffect | −7.3% | +0.4% |
| AnimatedVisibility | +1.2% | −0.5% |
| CanvasDrawing | −2.3% | +1.2% |
| HeavyShader | −0.0% | +0.6% |
| TextLayout | +1.3% | +0.6% |
| LazyGrid | +1.7% | +1.0% |
| LazyGrid-ItemLaunchedEffect | +4.4% | +1.9% |

PS I'm still running more tests/repeats to have more confidence here

MatkovIvan:
The regression in `LazyGrid` benchmark comes from the content of every cell being re-recorded from scratch each frame (`LazyGrid` jumps ~50 items per frame, so almost the whole viewport is recomposed).
Because of that we pay to build the caches but they're literally never reused: a snapshot is built, drawn once, and thrown away by the very next frame. If a cell survived even 2 frames the cache would already pay for itself.
That's why `LazyGrid-SmoothScroll` has no regression (it's actually faster) - there the content scrolls and lives for many frames, not one.
The extra cost splits into two roughly equal halves:
  1. a pass to collect the dependent `RenderNode`s;
  2. recording the extra picture (the snapshot itself).

Both are small and, given the flag is opt-in, I think this is a reasonable cost to accept.

hub-bla (approved):

igordmn:
> Both are small and, given the flag is opt-in

We either need to have it opt-out now or have a plan/task to change it to opt-out.

Having it opt-in forever I see as a bad option, because we have to maintain code that can be used only by few.

From what you described, I see that it is worth having this optimization enabled by default. But when - is up to you.

MatkovIvan:
@igordmn It's already enabled by default in the current state of this PR - https://github.com/JetBrains/skiko/pull/1022/changes/93b06027b6bb9e6ae008557b4bae7f474b964cac

skiko/src/commonMain/cpp/common/include/node/RenderNode.h
```diff
@@ -68,15 +69,44 @@ class RenderNode : public SkDrawable {
 
     void drawInto(SkCanvas *canvas);
```
igordmn, line 71: Let's run Compose Snapshot tests before merging this PR.

If you have issues with `jbVerifyDependencyVersions` - you can comment it, or we can change the Skiko snapshot versioning now.

igordmn (approved):

skiko/src/commonMain/cpp/common/include/node/RenderNode.h
MatkovIvan: https://teamcity.jetbrains.com/buildConfiguration/JetBrainsPublicProjects_Compose_AllPersonalBuild/6379270

PR #1024, opened 2025-02-11
Remove BreakIterator.clone() method

Alovchin91:
This PR is split-off from and is a prerequisite of Clang-cl PR: #1020

BreakIterator.clone() is broken on JVM and seemingly leads to a use-after-free behaviour. Clang-cl makes this issue more visible, likely due to address sanitisation.

Explanation why it is deleted in an internal channel: https://jetbrains.slack.com/archives/C02DDNREC77/p1738678000232729
* `breakIteratorCloneTest` is failing with an access violation in Debug mode with additional checks, so using it wasn't safe
* it is not used anywhere

This is a breaking change and will require bumping Skiko version to 0.9.0.

### DIFF
--- a/skiko/src/commonMain/kotlin/org/jetbrains/skia/BreakIterator.kt
+++ b/skiko/src/commonMain/kotlin/org/jetbrains/skia/BreakIterator.kt
@@ -306,14 +306,6 @@ class BreakIterator internal constructor(ptr: NativePointer) : Managed(ptr, _Fin
     }
 
     /**
-     * Create a copy of this iterator
-     */
-    fun clone(): BreakIterator {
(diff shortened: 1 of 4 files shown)


MatkovIvan (approved):
> This is a breaking change and will require bumping Skiko version to 0.9.0.

We're still in 0.x versions and usually do not bump the second version because of removing some minor APIs. So I don't mind to bump (it doesn't really matter), but it's not about "will require"

igordmn:
>  So I don't mind to bump (it doesn't really matter), but it's not about "will require"

I am more for bumping because of changing the compiler on one of the platforms

PR #1029, opened 2025-02-27
Expose ICUs u_hasBinaryProperty

m-sasha:
This is needed to detect emojis in Compose Multiplatform.

Also, it looks like the native implementations of `String.intCodePoints()` were broken. I fixed them by copying the relevant code from Compose.

### DIFF
--- a/skiko/src/commonMain/kotlin/org/jetbrains/skia/CodePoint.kt
+++ b/skiko/src/commonMain/kotlin/org/jetbrains/skia/CodePoint.kt
@@ -1,0 +1,46 @@
+package org.jetbrains.skia
+
+// TODO Remove once it's available in common stdlib https://youtrack.jetbrains.com/issue/KT-23251
+internal typealias CodePoint = Int
+
+/**
+ * The minimum value of a supplementary code point, `\u0x10000`.
+ */
+private const val MIN_SUPPLEMENTARY_CODE_POINT: Int = 0x10000
+
+/**
+ * Converts a surrogate pair to a unicode code point.
+ */
+internal fun toCodePoint(high: Char, low: Char): CodePoint =
+    (((high - Char.MIN_HIGH_SURROGATE) shl 10) or (low - Char.MIN_LOW_SURROGATE)) + MIN_SUPPLEMENTARY_CODE_POINT
+
+internal fun Int.charCount(): Int = if (this >= MIN_SUPPLEMENTARY_CODE_POINT) 2 else 1
+
+internal val CharSequence.codePoints
+    get() = codePointsAt(0)
+
+internal fun CharSequence.codePointsAt(index: Int) = sequence {
+    var current = index
+    while (current < length) {
+        val codePoint = codePointAt(current)
+        yield(codePoint)
+        current += codePoint.charCount()
+    }
+}
+
+internal val CharSequence.codePointsAsIntArray: IntArray
+    get() = codePoints.toList().toIntArray()
+
+/**
+ * Returns the character (Unicode code point) at the specified index.
+ */
+internal fun CharSequence.codePointAt(index: Int): CodePoint {
+    val high = this[index]
+    if (high.isHighSurrogate() && index + 1 < this.length) {
+        val low = this[index + 1]
+        if (low.isLowSurrogate()) {
+            return toCodePoint(high, low)
+        }
+    }
+    return high.code
+}
--- a/skiko/src/commonMain/kotlin/org/jetbrains/skia/icu/Unicode.kt
+++ b/skiko/src/commonMain/kotlin/org/jetbrains/skia/icu/Unicode.kt
@@ -47,10 +47,13 @@ object CharDirection {
      */
     fun of(codePoint: Int): Int {
         Stats.onNativeCall()
-        return charDirection(codePoint)
+        return _nCharDirection(codePoint)
     }
 }
 
+/**
+ * Bundles functions to inspect Unicode character properties.
+ */
 @Suppress("MemberVisibilityCanBePrivate", "unused", "SpellCheckingInspection")
 object CharProperties {
     init {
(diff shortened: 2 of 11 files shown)


skiko/src/commonMain/kotlin/org/jetbrains/skia/CodePoint.kt
```diff
@@ -0,0 +1,51 @@
...
+ * Converts a surrogate pair to a unicode code point.
+ */
+internal fun toCodePoint(high: Char, low: Char): CodePoint =
+    (((high - Char.MIN_HIGH_SURROGATE) shl 10) or (low - Char.MIN_LOW_SURROGATE)) + MIN_SUPPLEMENTARY_CODE_POINT
+
+// Copied from CharHelpers.skiko.kt
+internal fun Int.charCount(): Int = if (this >= MIN_SUPPLEMENTARY_CODE_POINT) 2 else 1
+
+internal val String.codePointsAsIntArray: IntArray
+    get() = codePoints.toList().toIntArray()
```
MatkovIvan, line 24: Doesn't it create an extra copy?

m-sasha: It does (actually more than once, inside `toList()`), but it's not easy to avoid this copy, as the number of codepoints is not known upfront. And if we just return `List<Int>` here, the callers will still need to convert it to an int array.

skiko/src/commonMain/kotlin/org/jetbrains/skia/CodePoint.kt
```diff
@@ -0,0 +1,51 @@
+package org.jetbrains.skia
+
+// Copied from CharHelpers.skiko.kt
```
MatkovIvan, line 3: Since it's from another repo, it's not super relevant. I'd suggest remove or replace with full link

skiko/src/commonMain/kotlin/org/jetbrains/skia/CodePoint.kt
```diff
@@ -0,0 +1,51 @@
...
+internal fun Int.charCount(): Int = if (this >= MIN_SUPPLEMENTARY_CODE_POINT) 2 else 1
+
+internal val String.codePointsAsIntArray: IntArray
+    get() = codePoints.toList().toIntArray()
+
+internal val String.codePoints
+    get() = codePointsAt(0)
+
+internal fun String.codePointsAt(index: Int) = sequence {
```
MatkovIvan, line 29: The code in `CharHelpers.skiko.kt` mentioned is actually like:
```suggestion
internal val CharSequence.codePoints get() = sequence {
```

skiko/src/commonMain/kotlin/org/jetbrains/skia/icu/Unicode.kt
```diff
@@ -44,9 +45,149 @@ object CharDirection {
...
      */
-    fun of(codePoint: Int): Int = charDirection(codePoint)
+    fun of(codePoint: Int): Int {
+        Stats.onNativeCall()
+        return charDirection(codePoint)
+    }
+}
+
+@Suppress("MemberVisibilityCanBePrivate", "unused", "SpellCheckingInspection")
+object CharProperties {
```
MatkovIvan, line 55: Missing KDoc for public API

skiko/src/commonMain/kotlin/org/jetbrains/skia/icu/Unicode.kt
```diff
@@ -44,9 +45,149 @@ object CharDirection {
...
+
 }
 
 @ExternalSymbolName("org_jetbrains_skia_icu_Unicode_charDirection")
 @ModuleImport("./skiko.mjs", "org_jetbrains_skia_icu_Unicode_charDirection")
 private external fun charDirection(codePoint: Int): Int
+
+@ExternalSymbolName("org_jetbrains_skia_icu_Unicode_nCodePointHasBinaryProperty")
+@ModuleImport("./skiko.mjs", "org_jetbrains_skia_icu_Unicode_nCodePointHasBinaryProperty")
+private external fun nCodePointHasBinaryProperty(codePoint: Int, property: Int): Boolean
```
MatkovIvan, line 192: Let's keep prefixes consistent (one way or another)

MatkovIvan (commented):
LGTM, just a few nitpicks

skiko/src/commonMain/kotlin/org/jetbrains/skia/CodePoint.kt
m-sasha: Removed.

skiko/src/commonMain/kotlin/org/jetbrains/skia/CodePoint.kt
m-sasha: Yes, the copy is after my changes in https://github.com/JetBrains/compose-multiplatform-core/pull/1869
This is more generic and is not worse, so why change it?

skiko/src/commonMain/kotlin/org/jetbrains/skia/icu/Unicode.kt
m-sasha: Added.

skiko/src/commonMain/kotlin/org/jetbrains/skia/CodePoint.kt
MatkovIvan: `String`/`CharSequence` consistency

skiko/src/commonMain/kotlin/org/jetbrains/skia/icu/Unicode.kt
m-sasha: Changed.

skiko/src/commonMain/kotlin/org/jetbrains/skia/CodePoint.kt
m-sasha: We have it in more than one place; the one I copied it from uses `String`, but sure, `CharSequence` is better.

MatkovIvan:
@m-sasha you forgot to push the changes, right?

m-sasha:
> @m-sasha you forgot to push the changes, right?

Yes, sorry, pushed now.

PR #1030, opened 2025-03-01
add BackendTexture and Image.adoptTextureFrom

MaapuH:
### New:  
(related #1019)
1. Adds bindings to GrBackendTextures::MakeGL 
2. Adds bindings to SkImages::AdoptTextureFrom
3. Adds bindings to GrBackendTextures::GLTextureParametersModdified

### Changes in build scripts:
(related #742)
1. Add file paths treating if building on Windows machine
2. Use "emcc.bat" instead of "emcc" if building on Windows machine, as described on Emscripten website

### Use-case: 
Allows skia to draw images, rendered externally on gpu.
Compose would benefit greatly, allowing to embed VideoPlayer or Browser without any friction or need to use Swing.


MaapuH:
At the moment I am working on these bindings:
DirectContext.createBackendTexture() // this is common for all backends
DirectContext.deleteBackendTexture() // this is common for all backends

BackendTexture.backend // allows to check which backend is texture created on
BackendTexture.getGLInfo // allows to get id of the texture if the backend is GL, this we can use for external rendering

We can also getDirect3DInfo for DirectX backend, getMetalInfo for Metal, etc
These I'll be doing when I'm done with GL bindings

All of that will allow for easy integration with skia, like I wrote in the beginning

Should I convert this to draft, or just submit another PR when all the upcoming bindigs are implemented?

skiko/src/nativeJsMain/cpp/BackendTexture.cc
```diff
@@ -0,0 +1,38 @@
...
+
+static void deleteBackendTexture(GrBackendTexture* rt) {
+    delete rt;
+}
+SKIKO_EXPORT KNativePointer org_jetbrains_skia_BackendTexture__1nGetFinalizer() {
+    return reinterpret_cast<KNativePointer>(&deleteBackendTexture);
+}
+
+SKIKO_EXPORT KNativePointer org_jetbrains_skia_BackendTexture__1nMakeGL
+  (KInt textureId, KInt target, KInt width, KInt height, KInt format, KInt surfaceOrigin, KInt colorType, KBoolean isMipmapped) {
```
igordmn: - the arguments are not equal to the arguments of `_nMakeGL`
- `surfaceOrigin` and `colorType` are unused, we should remove them
- the order of the arguments here (and in the public Kotlin function) is better to make the same in in the Skia function:
```
KInt width, KInt height, KBoolean isMipmapped, KInt textureId, KInt target, KInt format
```

skiko/src/commonMain/kotlin/org/jetbrains/skia/BackendTexture.kt
```diff
@@ -0,0 +1,79 @@
...
+            val ptr = _nMakeGL(
+                textureId,
+                textureTarget,
+                textureFormat,
+                width,
+                height,
+                isMipmapped
+            )
+            return if (ptr == NullPointer) throw RenderException("Can't create OpenGL BackendTexture")
+            else BackendTexture(ptr)
```
igordmn:
```suggestion
            BackendTexture(ptr)
```
According the implementation of `_nMakeGL`, ptr is never null. The Kdoc should also be changed

skiko/src/commonMain/kotlin/org/jetbrains/skia/BackendTexture.kt
```diff
@@ -0,0 +1,79 @@
...
+import org.jetbrains.skia.impl.*
+import org.jetbrains.skia.impl.Library.Companion.staticLoad
+import org.jetbrains.skia.impl.getPtr
+import org.jetbrains.skia.impl.reachabilityBarrier
+import org.jetbrains.skiko.RenderException
+
+class BackendTexture internal constructor(ptr: NativePointer) : Managed(ptr, _FinalizerHolder.PTR) {
+    companion object {
+        /**
+         * Creates BackendTexture from GL texture.
```
igordmn, line 9: Please, describe all arguments

igordmn (changes requested):
Thanks! Sorry for a long review.

It needs small corrections, and it is good to merge.

MaapuH:
Corrections applied!

skiko/src/nativeJsMain/cpp/BackendTexture.cc
```diff
@@ -0,0 +1,38 @@
...
+    delete rt;
+}
+SKIKO_EXPORT KNativePointer org_jetbrains_skia_BackendTexture__1nGetFinalizer() {
+    return reinterpret_cast<KNativePointer>(&deleteBackendTexture);
+}
+
+SKIKO_EXPORT KNativePointer org_jetbrains_skia_BackendTexture__1nMakeGL
+  (KInt width, KInt height, KBoolean isMipmapped, KInt textureId, KInt target, KInt format) {
+    GrGLTextureInfo textureInfo;
+    textureInfo.fID = static_cast<GrGLuint>(textureId);
```
serjsysoev, line 19: AFAIK WebGL textures have no ids, so emscripten provides the mapping for skia. This means that without registering a texture in the emscripten internal map it is not possible to adopt textures created outside skia (e.g. using js interop). Should we include a [workaround](https://github.com/emscripten-core/emscripten/issues/6103#issuecomment-358916669) for this to allow registering external webgl textures?
It probably should be done as a separate PR, just bringing this up as it would be useful for my use case (adopting textures created using js interop).

igordmn: This workaround looks like a code that should be called outside of Skia, not in Skia itself. Skia only provides an API which can be used, including requiring an id that can be obtained by the API you provided.

PR #1032, opened 2025-03-04
Fix the background flashing when closing a window/dialog.

m-sasha:
Flushing the transaction causes the window background to be visible for a frame when closing the window. 

Fixes: https://youtrack.jetbrains.com/issue/CMP-5651

This reverts https://github.com/JetBrains/skiko/pull/552. I have tested the IDE sample from Jewel, and that issue does not seem to reproduce anymore, even without flushing the transaction.

https://github.com/user-attachments/assets/a3c7c0ab-2342-4cae-8c52-669faeb1f516

### DIFF
--- a/skiko/src/awtMain/objectiveC/macos/MetalRedrawer.mm
+++ b/skiko/src/awtMain/objectiveC/macos/MetalRedrawer.mm
@@ -269,7 +269,6 @@ JNIEXPORT void JNICALL Java_org_jetbrains_skiko_redrawer_MetalRedrawer_disposeDe
         env->DeleteGlobalRef(device.layer.javaRef);
         [[NSNotificationCenter defaultCenter] removeObserver:device.occlusionObserver];
         [device.layer removeFromSuperlayer];
-        [CATransaction flush];
     }
 }
 


elijah-semyonov (approved):
Just for more context:
What `CATransaction.flush` does is committing all changes of `CALayer` properties and trees to be rendered to be visible during next frame.

CoreAnimation rendering is a blackbox so grouping transactions explicitly or implicitly is the only way to make multiple changes in the tree visually simultaneous.

Implicit transactions are always scoped within main runloop work items so this invocation here introduced a state barrier between before and after potentially allowing CoreAnimation to render two different frames with states from before and after the barrier, which is perceived as flickering. 

In order to avoid flickering with guarantees we somehow need to ensure that both changes of `CALayer` hierarchy by AWT and us are confined within one transaction, which this PR is required for but doesn't guarantee as far as I see it.

m-sasha:
@elijah-semyonov Thanks for the input.

What I understood, but it clearly wrong, is that flushing the transaction is optional. I know it's wrong because removing the flush from other places in `MetalRedrawer.mm`, such as in `resizeLayers`, breaks Compose drawing completely (nothing is drawn).

> In order to avoid flickering with guarantees we somehow need to ensure that both changes of CALayer hierarchy by AWT
> and us are confined within one transaction, which this PR is required for but doesn't guarantee as far as I see it.

It seems to me that in this particular instance, we're flushing a transaction that isn't our own (there's no `[CATransaction begin];` in `disposeDevice`). Is that even reasonable?

igordmn (approved):

elijah-semyonov:
As per [doc](https://developer.apple.com/documentation/quartzcore/catransaction/flush()) flush always happens in the end of runloop work item. In other words, any job executed on the macOS executable main thread will be confined within `CATransaction`.

PR #1033, opened 2025-03-08
Call the listeners of PropertyKind.Renderer on render API fallback

m-sasha:
Currently, these listeners are only called when the `renderApi` is set manually, not when it changes to a fallback.


MatkovIvan (commented):
Probably it would be better to split the fix and cleaning/refactor, but it's not too big so LGTM.
Please wait for Igor's review before the merge

igordmn (approved):

PR #1039, opened 2025-03-25
[Metal] Fix window background flash on dispose/hide while animation is running

m-sasha:
The issue with flashing on dispose is fixed by disabling `displaySyncEnabled` on the native layer in `MetalRedrawer_disposeDevice`.

The issue with flashing on hide is fixed by avoiding a call to `MetalRedrawer.setVisible(false)` in the particular scenario of the window becoming hidden.


m-sasha:
@elijah-semyonov If you could say whether this makes sense...

PR #1040, opened 2025-03-27
Update emscripten version in docker

eymar:
emscripten 3.1.50 supports mimalloc - https://github.com/emscripten-core/emscripten/blob/main/ChangeLog.md#3150---112923 

We want to try it.

### DIFF
--- a/skiko/docker/linux-emscripten-amd64/Dockerfile
+++ b/skiko/docker/linux-emscripten-amd64/Dockerfile
@@ -10,7 +10,7 @@ RUN apt-get update -y && \
 RUN apt-get install git python wget -y && \
     apt-get install fontconfig libfontconfig1-dev libglu1-mesa-dev libxrandr-dev libdbus-1-dev curl zip -y
 ENV EMSDK_DIR=/usr/emsdk
-ENV EMSDK_VER=3.1.49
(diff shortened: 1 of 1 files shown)


Schahen (approved):

skiko/docker/linux-emscripten-amd64/Dockerfile
```diff
@@ -10,7 +10,7 @@ RUN apt-get update -y && \
 RUN apt-get install git python wget -y && \
     apt-get install fontconfig libfontconfig1-dev libglu1-mesa-dev libxrandr-dev libdbus-1-dev curl zip -y
 ENV EMSDK_DIR=/usr/emsdk
-ENV EMSDK_VER=3.1.49
+ENV EMSDK_VER=3.1.50
```
MatkovIvan, line 13: Why 3.1.50?

Can we update it to the more recent one?
3.1.57 had a breaking change (see https://github.com/emscripten-core/emscripten/issues/21756), so it probably makes sense to use newer version to avoid such incompatibilities

eymar: our Skia is built using 3.1.44 - it's incompatible with 3.1.57, but compatible with 3.1.50

PR #1041, opened 2025-03-28
SkiaMultiplatformSample. Fix showing render API after fallback

igordmn:
renderApi can change if there was a fallback (DIRECT3D -> OpenGL -> Software). We need to always call it in the frame.

Noticed it during testing.

### DIFF
--- a/samples/SkiaMultiplatformSample/src/awtMain/kotlin/org/jetbrains/skiko/sample/AwtClocks.kt
+++ b/samples/SkiaMultiplatformSample/src/awtMain/kotlin/org/jetbrains/skiko/sample/AwtClocks.kt
@@ -8,7 +8,7 @@ import java.awt.event.MouseMotionListener
 import java.awt.event.MouseWheelEvent
 import java.awt.event.MouseWheelListener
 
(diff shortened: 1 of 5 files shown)


samples/SkiaMultiplatformSample/src/awtMain/kotlin/org/jetbrains/skiko/sample/AwtClocks.kt
```diff
@@ -8,7 +8,7 @@ import java.awt.event.MouseMotionListener
 import java.awt.event.MouseWheelEvent
 import java.awt.event.MouseWheelListener
 
-class AwtClocks(private val layer: SkiaLayer) : Clocks(layer.renderApi), MouseMotionListener, MouseWheelListener {
+class AwtClocks(private val layer: SkiaLayer) : Clocks({ layer.renderApi }), MouseMotionListener, MouseWheelListener {
```
m-sasha, line 11: Maybe `layer::renderApi` instead?

MatkovIvan (approved):

m-sasha (approved):

PR #1042, opened 2025-04-04
Add configurable clearColor to SkiaLayer to avoid white background during resize on Windows.

bezsahara:
This adds a `clearColor` field to `SkiaLayer`, supported by a new SkiaLayerProperties.clearColor value.
It defaults to white but can be overridden via a system property (`SKIKO_CLEAR_COLOR`) or by setting `SkiaLayerProperties.defaultClearColor` early at startup.

This helps eliminate the white flicker during window resizing in Compose Desktop (observed on Windows).
The change is fully backward-compatible and has no effect unless explicitly configured.

I’ve tested it, and the white background disappears completely. Here are links to issues on YouTrack that describe the issue:
https://youtrack.jetbrains.com/issue/CMP-7700/Desktop-Synchronize-Compose-and-Swing-layout-and-rendering-phases
https://youtrack.jetbrains.com/issue/CMP-7919/White-background-flicker-during-window-resize-in-Compose-Desktop


m-sasha:
I'm currently investigating the same issue on macOS, and after that I'll look into Windows as well.

JonnyTsGitHub:
Please review and accept this PR.  I need this as well.

m-sasha:
Is this fixed by https://github.com/JetBrains/skiko/pull/1141 ?

Issue #740, opened 2023-07-01
Undefined symbols for architecture arm64


ARTI1208:
Trying to use compose in KMM (kotlin 1.9-RC, 1.5.0-dev1080, skiko 0.7.63, also checked skiko 0.7.67), got error linking for IOS

```
Undefined symbols for architecture arm64:
  "_org_jetbrains_skia_TextBlobBuilderRunHandler__1nGetFinalizer", referenced from:
      _kfun:org.jetbrains.skia.shaper.TextBlobBuilderRunHandler._FinalizerHolder.$init_global#internal in result.o
  "_org_jetbrains_skia_TextBlobBuilderRunHandler__1nMake", referenced from:
      _kfun:org.jetbrains.skia.shaper.TextBlobBuilderRunHandler#<init>(org.jetbrains.skia.ManagedString?;kotlin.Boolean;kotlin.Float;kotlin.Float){} in result.o
  "_org_jetbrains_skia_TextBlobBuilderRunHandler__1nMakeBlob", referenced from:
      _objc2kotlin_kfun:org.jetbrains.skia.shaper.TextBlobBuilderRunHandler#makeBlob(){}org.jetbrains.skia.TextBlob? in result.o
  "_org_jetbrains_skia_svg_SVGCanvasKt__1nMake", referenced from:
      _kfun:org.jetbrains.skia.svg.SVGCanvas#make(org.jetbrains.skia.Rect;org.jetbrains.skia.WStream;kotlin.Boolean;kotlin.Boolean){}org.jetbrains.skia.Canvas in result.o
ld: symbol(s) not found for architecture arm64
```

I checked binaries and see that there are:
 _org_jetbrains_skia_svg_SVGCanvas__1nMake (not _org_jetbrains_skia_svg_SVGCanvas**Kt**__1nMake), 

_org_jetbrains_skia\_**shaper**_TextBlobBuilderRunHandler__1nGetFinalizer (not _org_jetbrains_skia_TextBlobBuilderRunHandler__1nGetFinalizer), 

_org_jetbrains_skia\_**shaper**_TextBlobBuilderRunHandler__1nMake (not _org_jetbrains_skia_TextBlobBuilderRunHandler__1nMake), 

_org_jetbrains_skia\_**shaper**_TextBlobBuilderRunHandler__1nMakeBlob (not _org_jetbrains_skia_TextBlobBuilderRunHandler__1nMakeBlob)

I guess @ExternalSymbolName values were not specified correctly / were not updated

[...]

PR #1049, opened 2025-04-10
Point `TextBlobBuilderRunHandler` to the correct native symbol

JakeWharton:
https://github.com/JetBrains/skiko/blob/10a91fc95a7fe2c3a6aa98f4001a065406eaf371/skiko/src/nativeJsMain/cpp/shaper/TextBlobBuilderRunHandler.cc#L9
https://github.com/JetBrains/skiko/blob/10a91fc95a7fe2c3a6aa98f4001a065406eaf371/skiko/src/nativeJsMain/cpp/shaper/TextBlobBuilderRunHandler.cc#L14
https://github.com/JetBrains/skiko/blob/10a91fc95a7fe2c3a6aa98f4001a065406eaf371/skiko/src/nativeJsMain/cpp/shaper/TextBlobBuilderRunHandler.cc#L21

Fixes [SKIKO-740](https://youtrack.jetbrains.com/issue/SKIKO-740) (#740) with #1050.

### DIFF
--- a/skiko/src/commonMain/kotlin/org/jetbrains/skia/shaper/TextBlobBuilderRunHandler.kt
+++ b/skiko/src/commonMain/kotlin/org/jetbrains/skia/shaper/TextBlobBuilderRunHandler.kt
@@ -80,14 +80,14 @@ class TextBlobBuilderRunHandler<T> internal constructor(
     }
 }
 
-@ExternalSymbolName("org_jetbrains_skia_TextBlobBuilderRunHandler__1nGetFinalizer")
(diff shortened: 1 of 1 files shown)


MatkovIvan (commented):
Sounds valid. It's weird that it's not covered by tests (`skiko/skiko/src/commonTest/kotlin/org/jetbrains/skia
/TextBlobBuilderTest.kt`), just a call from there will catch such issues (even without any asserts)

eymar:
Fixed in https://github.com/JetBrains/skiko/pull/1060

PR #1050, opened 2025-04-10
Point `SVGCanvas` to the correct native symbol

JakeWharton:
https://github.com/JetBrains/skiko/blob/10a91fc95a7fe2c3a6aa98f4001a065406eaf371/skiko/src/nativeJsMain/cpp/svg/SVGCanvas.cc#L7


MatkovIvan (commented):
Thanks! Could you please add some simple call of this function to `commonTest` to verify this on CI?

PR #1052, opened 2025-04-12
Add support for EGL

Thomas-Vos:
Requires skia-pack version update before merging: https://github.com/JetBrains/skia-pack/pull/68

### DIFF
--- a/skiko/src/commonMain/kotlin/org/jetbrains/skia/DirectContext.kt
+++ b/skiko/src/commonMain/kotlin/org/jetbrains/skia/DirectContext.kt
@@ -7,6 +7,13 @@ import org.jetbrains.skiko.loadOpenGLLibrary
 
 class DirectContext internal constructor(ptr: NativePointer) : RefCnt(ptr) {
     companion object {
+        fun makeEGL(): DirectContext {
+            Stats.onNativeCall()
+            val ptr = _nMakeEGL()
+            if (ptr == NullPointer) throw RenderException("Can't create EGL DirectContext")
+            return DirectContext(ptr)
+        }
+
         fun makeGL(): DirectContext {
             Stats.onNativeCall()
             loadOpenGLLibrary()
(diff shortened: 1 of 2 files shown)


eymar:
@Thomas-Vos I see you've udpated the Docker files. Do I understand correctly, that we have to rebuild the docker image first before running the CI? 

(@MatkovIvan has prepared an update for DOcker publishing - https://github.com/JetBrains/skiko/pull/1152), so running CI should get easier after it.

skiko/src/commonTest/kotlin/org/jetbrains/skia/SurfaceTest.kt
```diff
@@ -105,8 +111,12 @@ class SurfaceTest {
 
     @Test
     fun canMakeRenderTarget() {
-        if (!TestGlContext.isAvailabale()) return
+        if (!TestGlContext.isAvailable()) return
 
+        if (hostOs == OS.Linux && kotlinBackend == KotlinBackend.Native && hostArch == Arch.Arm64) {
+            // TODO: fix test on Linux arm64 using EGL
```
Thomas-Vos, line 117: I'm not sure how to get the canMakeRenderTarget working on CI (egl on linux arm64). Any ideas?

https://github.com/JetBrains/skiko/actions/runs/21452220344/job/61784849069

As EGL does work fine on my Linux ARM64 smart watch, it seems to be just a testing issue on CI, so I disabled this test for now.

MatkovIvan: Ignoring one test instead of all on arm64 lgtm

Thomas-Vos: Ok, it seems JVM tests on arm64 CI also fail with GL and fall back to software.

PR #1054, opened 2025-04-22
build: fix local machine build scripts

nikola-kocic-jetbrains:
* WASM was trying to double-link the libraries (on macOS)
* The minimal macOS version was not alignd between Skia and Skiko
* ARM Linux was misdetected (`uname -a` reports "aarch64" in that case)

### DIFF
--- a/skiko/buildSrc/src/main/kotlin/tasks/configuration/WasmTasksConfiguration.kt
+++ b/skiko/buildSrc/src/main/kotlin/tasks/configuration/WasmTasksConfiguration.kt
@@ -80,7 +80,9 @@ fun SkikoProjectContext.createWasmLinkTasks(): LinkWasmTasks = with(this.project
         if (outputES6) buildSuffix.set("es6")
         if (isForD8) buildSuffix.set("d8")
 
-        libFiles = project.fileTree(unpackedSkia) { include("**/*.a") }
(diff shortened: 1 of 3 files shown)


skiko/buildSrc/src/main/kotlin/tasks/configuration/WasmTasksConfiguration.kt
```diff
@@ -80,7 +80,9 @@ fun SkikoProjectContext.createWasmLinkTasks(): LinkWasmTasks = with(this.project
         if (outputES6) buildSuffix.set("es6")
         if (isForD8) buildSuffix.set("d8")
 
-        libFiles = project.fileTree(unpackedSkia) { include("**/*.a") }
+        val skiaBinSubdir = "$unpackedSkia/out/${buildType.id}-wasm-wasm"
+
+        libFiles = project.fileTree(skiaBinSubdir) { include("**/*.a") }
```
SergeevPavel, line 85: do I understand it right that before it was trying to link some unused libraries? Does it lead to decrease of resulting binary size?

nikola-kocic-jetbrains: Previously the build was just failing with the "duplicate symbols" error, as it was also trying to link the Skia third-party libraries, that were already linked.

PR #1057, opened 2025-05-07
[Metal] Decouple onRender from draw

m-sasha:
Allow the `Redrawer` to call `SkikoRenderDelegate.onRender` separately from drawing the resulting Picture to the screen. 
Then in `MetalRedrawer`:
- On `needRedraw(throttledToVsync=false)` call `onRender` immediately, but schedule drawing at the next vsync. This option will be used for regular Compose invalidation.
- On `needRedraw(throttledToVsync=true)` schedule both `onRender` and drawing at the next vsync. This option will be used to update animations in Compose.

This results in several improvements:
- If Compose invalidation happens several times between consecutive vsync events:
    - Only the latest (most up-to-date) state will be shown on the screen.
    - The latest state will be shown a frame sooner. 
- In response to Swing's `paint` call, there is no need to render (Compose), as we already have a picture ready to be drawn. So this can happen faster, and without blocking the EDT thread.


skiko/src/commonMain/kotlin/org/jetbrains/skiko/redrawer/Redrawer.kt
```diff
@@ -1,10 +1,15 @@
...
+private val initialTime = TimeSource.Monotonic.markNow()
+
 internal interface Redrawer {
     fun dispose()
-    fun needRedraw()
-    fun redrawImmediately()
+    fun needRedraw(throttledToVsync: Boolean)
+    fun redrawImmediately(updateNeeded: Boolean)
     fun syncBounds() = Unit
+    fun update(nanoTime: Long = initialTime.elapsedNow().inWholeNanoseconds)
```
MatkovIvan, line 12: ~Might be `TimeSource.Monotonic.markNow().elapsedNow().inWholeNanoseconds` without extra global val~

skiko/src/awtMain/kotlin/org/jetbrains/skiko/redrawer/MetalRedrawer.kt
```diff
@@ -73,63 +72,82 @@ internal class MetalRedrawer(
...
-            draw()
+            update()
+            drawOnlyFrameDispatcher.scheduleFrame()
         }
     }
 
+    private val throttledFrameDispatcher = FrameDispatcher(MainUIDispatcher) {
+        if (layer.isShowing) {
+            update()
+            drawOnlyFrameDispatcher.scheduleFrame()
```
igordmn, line 90: I have concerns that we schedule drawing here, not calling it immediately. We might skip some draw's, making jumpy pacing `update - update - draw - update - update - draw` instead of `update - draw - update - draw`:

```
frameDispatcher1
  frameDispatcher.update
    ...
    frameDispatcher.needRedraw(throttledToVsync= false) // schedules frameDispatcher2
  drawOnlyFrameDispatcher.scheduleFrame() // will be called after frameDispatcher2

frameDispatcher2
drawOnlyFrameDispatcher // lag, to late for draw, we had to draw before frameDispatcher2
frameDispatcher3
...
```

skiko/src/awtMain/kotlin/org/jetbrains/skiko/redrawer/MetalRedrawer.kt
```diff
@@ -73,63 +72,82 @@ internal class MetalRedrawer(
...
+        }
+        waitForVSyncIfNeeded()
```
igordmn, line 92: It looks like we wait for additional vsync:

```
vsync // waitForVSyncIfNeeded() in throttledFrameDispatcher
update()
drawOnlyFrameDispatcher.scheduleFrame()

vsync // waitForVSyncIfNeeded() in drawOnlyFrameDispatcher
draw()
```


What is the FPS for `draw` calls? I have suspicion that it is halved in half if that is the case.

igordmn (changes requested):
Thanks!

The idea to split update/draw looks good, but there are major issues we need to solve.

skiko/src/awtMain/kotlin/org/jetbrains/skiko/redrawer/VSyncer.kt
```diff
@@ -0,0 +1,73 @@
...
+import org.jetbrains.skiko.RendezvousBroadcastChannel
+
+
+/**
+ * A utility allowing several coroutines to wait for the next vsync.
+ */
+internal class VSyncer(windowPtr: Long) {
+
+    // The underlying throttler that blocks a thread
+    private val displayLinkThrottler = DisplayLinkThrottler(windowPtr)
```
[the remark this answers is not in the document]
m-sasha: I'm hoping to add this behavior to other renderers as well, but sure, for now I'll rename it to `MetalVSyncer`.

skiko/src/awtMain/kotlin/org/jetbrains/skiko/SkiaLayer.awt.kt
```diff
@@ -1,5 +1,6 @@
 package org.jetbrains.skiko
```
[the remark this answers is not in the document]
m-sasha: You can see it in the last [two](https://github.com/JetBrains/skiko/compare/975891a7f34386d1ec566e68312c8d72badee0d2..79eec5d603d01093bc9b041420ddffe7041c1c82) [changes](https://github.com/JetBrains/skiko/compare/79eec5d603d01093bc9b041420ddffe7041c1c82..7da9fef5791554f02d228df34199694bf9819919).

Calling `contentPane.paint` was causing an update+draw to happen immediately.

Now it's enough to call `pack()` on the window in order to get the first update+draw to happen. When `pack` is called, it causes `reshape`, but not `paint` to be called. In `reshape` I check whether the component is showing, and if not, paint immediately.

skiko/src/awtMain/kotlin/org/jetbrains/skiko/SkiaLayer.awt.kt
```diff
@@ -365,41 +399,6 @@ actual open class SkiaLayer internal constructor(
...
-    override fun setBounds(x: Int, y: Int, width: Int, height: Int) {
-        super.setBounds(x, y, width, height)
-
-        // To avoid visual artifacts on Windows/Direct3D,
-        // redrawing should be performed immediately, without scheduling to "later".
-        // Subscribing to events instead of overriding this method won't help too.
-        //
-        // Please note that calling redraw during layout might break software renderers,
-        // so applying this fix only for Direct3D case.
-        if (renderApi == GraphicsApi.DIRECT3D && isShowing) {
```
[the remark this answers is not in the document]
m-sasha: It's strange, because `SkiaLayer.reshape` (which is the same as `SkiaLayer.setBounds`, calls `validate()` which in turn calls `doLayout`, which in turn calls `HardwareLayer.reshape`, where `redrawer?.syncBounds` and `redrawer?.redrawImmediately` are called. So it seems like it should be the same.

I'll investigate.

skiko/src/awtMain/kotlin/org/jetbrains/skiko/redrawer/MetalRedrawer.kt
```diff
@@ -73,63 +72,82 @@ internal class MetalRedrawer(
         }
         _device = initDevice
         contextHandler = MetalContextHandler(layer, initDevice, adapter)
-        setDisplaySyncEnabled(initDevice.ptr, properties.isVsyncEnabled)
+        setDisplaySyncEnabled(initDevice.ptr, false)  // VSync is managed manually (see vsyncedFrameDispatcher)
```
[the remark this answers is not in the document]
m-sasha: I did it on purpose, but I don't remember why anymore. I'll revert it and see if it breaks something. The tests pass with it.

skiko/src/awtMain/kotlin/org/jetbrains/skiko/redrawer/MetalRedrawer.kt
```diff
@@ -73,63 +72,82 @@ internal class MetalRedrawer(
...
     }
 
     override val renderInfo: String get() = contextHandler.rendererInfo()
 
     private val frameDispatcher = FrameDispatcher(MainUIDispatcher) {
```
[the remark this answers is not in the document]
m-sasha: That's the plan, when/if I implement it for other renderers.

skiko/src/awtMain/kotlin/org/jetbrains/skiko/redrawer/MetalRedrawer.kt
m-sasha: If we draw immediately here, the `temporary change is not visible` test fails because a running animation causes the temporary change to be drawn.

I'm not sure I understand your concern, though. This whole PRs purpose is to allow a sequence like `update - update - draw`. It makes sure that drawing only happens on vsync.

skiko/src/awtMain/kotlin/org/jetbrains/skiko/redrawer/MetalRedrawer.kt
m-sasha: I'm getting exactly 60fps.

The sequence of events for an animation is:
1. [time=0] Compose calls `withFrameNanos` which calls `onNewAwaiters`, which calls `needRedraw(throttledToVsync=true)`, which schedules a frame on `throttledFrameDispatcher`.
2. [time=~0] `throttledFrameDispatcher` lambda is called on the next available EDT iteration.
3. [time=~0] `update()` is called from the `throttledFrameDispatcher` lambda. Compose calls `withFrameNanos` again, which calls `needRedraw(throttledToVsync=true)` which schedules a frame on `throttledFrameDispatcher`.
4. [time=~0] `drawOnlyFrameDispatcher.scheduleFrame()` is called from the `throttledFrameDispatcher` lambda.
5. The coroutine for `throttledFrameDispatcher` suspends until the next vsync.
6. [time=~0] `drawOnlyFrameDispatcher` lambda is called on the next available EFT iteration and suspends until the next vsync.
7. [time=~16ms] `throttledFrameDispatcher` coroutine wakes up due to vsync and suspends in `channel.receive()`.
8. [time=~16ms]  `drawOnlyFrameDispatcher` coroutine wakes up due to vsync, calls `draw` and suspends in `channel.receive()`.
9. [time=~16ms] `throttledFrameDispatcher` coroutine wakes up (because of 3) and we go back to 3.

skiko/src/awtMain/kotlin/org/jetbrains/skiko/redrawer/VSyncer.kt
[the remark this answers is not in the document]
igordmn: A similar code isn't written at the moment for other renderers. The sync happens during buffer swapping a blocking way, in the end of the frame.

It can be difficult to find an appropriate similar API.

skiko/src/awtMain/kotlin/org/jetbrains/skiko/SkiaLayer.awt.kt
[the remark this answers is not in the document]
igordmn: > In reshape I check whether the component is showing, and if not, draw immediately.

I see, it should fix the issue indeed.

skiko/src/awtMain/kotlin/org/jetbrains/skiko/redrawer/MetalRedrawer.kt
igordmn: > temporary change is not visible

Looking at the test, I see issues it is incorrect requirement that we shouldn't draw/show temporary updates. If something is updated, and it requested a draw, we should draw it ASAP, to avoid a large latency between the state update and the actual change on the screen.

in this code:
```
renderDelegate.color = tempColor
renderDelegate.layer.needRedraw(throttledToVsync = true)
renderChannel.receive()  // Wait until render was actually called
renderDelegate.color = color
renderDelegate.layer.needRedraw(throttledToVsync = true)
```
if there was a vsync after `renderChannel.receive()`, we should draw it.

igordmn: > This whole PRs purpose is to allow a sequence like update - update - draw

The is also another change in this PR - not waiting for vsync for some updates. It is good change.

But `update - update - draw` as I said above adds latencies, and can introduce frame drops. Ideally every update should be followed by a draw, so we can see the actual state on the screen (it can be not achievable on macOs because of lost frames issue)

igordmn: > (it can be not achievable on macOs because of lost frames issue

There is a theoretical way (don't sent the last buffer until 2 others are in flight, and reuse it for new content). But this is out of scope of this PR

skiko/src/awtMain/kotlin/org/jetbrains/skiko/redrawer/MetalRedrawer.kt
igordmn: > I'm getting exactly 60fps.

Did you check draw calls? You should move `fpsCounter?.tick()` from `update` to `draw` to measure that.

skiko/src/awtMain/kotlin/org/jetbrains/skiko/redrawer/MetalRedrawer.kt
igordmn: >  to avoid a large latency between the state update and the actual change on the screen.

I remember we discussed online that it is to avoid issues for blinking. But I believe that it should be fixed not by the renderer logic, but by the application logic. Because it is not achievable in the renderer logic, we can only add a latency that just reduce frequency of blinking, but not eliminate fully.

skiko/src/awtMain/kotlin/org/jetbrains/skiko/redrawer/MetalRedrawer.kt
m-sasha: Of course.

skiko/src/awtMain/kotlin/org/jetbrains/skiko/SkiaLayer.awt.kt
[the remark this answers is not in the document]
m-sasha: I've put the hack back in (in `SkiaLayer.reshape`).
I guess it takes a little longer for the execution to get to `HardwareLayer.reshape`, and that makes the difference.

I will need to investigate alternatives when I get to improving skiko on Windows.

skiko/src/awtMain/kotlin/org/jetbrains/skiko/SkiaLayer.awt.kt
```diff
@@ -100,18 +103,29 @@ actual open class SkiaLayer internal constructor(
...
         backedLayer = object : HardwareLayer(externalAccessibleFactory) {
-            override fun paint(g: java.awt.Graphics) {
-                Logger.debug { "Paint called on $this" }
-                checkContentScale()
 
-                // 1. JPanel.paint is not always called (in rare cases).
-                //    For example if we call 'jframe.isResizable = false` on Ubuntu
-                //
-                // 2. HardwareLayer.paint is also not always called.
-                //    For example, on macOs when we resize window or change DPI
```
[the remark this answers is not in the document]
m-sasha: In the last commit, I made it so `update` is called from `paint` when the scale changes, just in case:
```
val updateNeeded = checkContentScale()
redrawer?.redrawImmediately(updateNeeded = updateNeeded)
```

skiko/src/awtMain/kotlin/org/jetbrains/skiko/redrawer/MetalRedrawer.kt
```diff
@@ -78,11 +77,38 @@ internal class MetalRedrawer(
...
+        }
+    }
+
     private val frameDispatcher = FrameDispatcher(MainUIDispatcher) {
         if (layer.isShowing) {
-            update(System.nanoTime())
+            // Schedule drawing before calling update so that if update
+            // schedules this frameDispatcher again, and update is a long
+            // operation, the frame is drawn before update is called again.
+            drawOnlyFrameDispatcher.scheduleFrame()
```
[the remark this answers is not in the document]
igordmn: > The point of this PR is to separate update from draw, so naturally other code could be executed between them.

During our discussion, I tried to emphasis that update/draw are the same thing, just on different levels, and ideally shouldn't be separated, and should always go one after another. Update - prepares the buffer in canvas draw call way (picture), draw - prepares buffer in GPU draw call way, and flushes it to the screen. All other work (recomposition, layout) is just a preparation work is needed to make these draw calls. If there are no draw calls (window is invisible), there is no point to recompose/layout the window content.

Separating update/draw can cause different kind of desync issues. One of the potential things that can cause issues - handling events between update/draw, I am currently thinking about it.

But we can call them more frequently, this is also done in this PR, and this reduces the input lag, which is good.

There is also the macOs-only issue with 3 buffers, but it shouldn't affect the overall semantics. Also, there is a possible solution (I wrote earlier about it)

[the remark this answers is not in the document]
igordmn: > I am currently thinking about it.

Thoughts:
- the events can't affect the draw state, as a new `update` call still is not scheduled, and only `update` can affect `draw`. So semantically, it is okay.
- still, handling the events earlier wastes the CPU time and postpones the draw call. We are risking to not perform it in time and cause a frame drop. So, performance-wise, it it probably not okay.

I have some thoughts of using `launch(start = UNDISPATCHED`, will try to solve it via it, not via `FrameDispatcher`

skiko/src/awtMain/kotlin/org/jetbrains/skiko/redrawer/MetalRedrawer.kt
```diff
@@ -91,46 +117,48 @@ internal class MetalRedrawer(
...
+    private suspend fun waitForVSyncIfNeeded() {
+        vSyncer?.waitForVSync()
     }
 
-    override fun redrawImmediately() {
-        check(!isDisposed) { "MetalRedrawer is disposed" }
-        inDrawScope {
-            update(System.nanoTime())
-            if (!isDisposed) { // Redrawer may be disposed in user code, during `update`
-                performDraw(waitVsync = SkikoProperties.macOSWaitForPreviousFrameVsyncOnRedrawImmediately)
```
igordmn, line 112: @SergeevPavel, we are removing `skiko.rendering.macos.waitForPreviousFrameVsyncOnRedrawImmediately`, replacing the waiting for vsync in the main thread by scheduling it to a background coroutine that also waits for vsync (but non-blocking).

Is it fine, or do you prefer restoring immediate redraw? Just for information, if it is called 3-4 times in one vsync, you can lose the last frame.

skiko/src/awtMain/kotlin/org/jetbrains/skiko/redrawer/MetalRedrawer.kt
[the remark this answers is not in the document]
igordmn: @m-sasha, please review my changes in:
https://github.com/JetBrains/skiko/pull/1088

After them and [this](https://github.com/JetBrains/skiko/pull/1057#discussion_r2301582284), this PR is good to merge

igordmn (commented):
Approved from me, but let's wait for @SergeevPavel answer before merge

skiko/src/awtMain/kotlin/org/jetbrains/skiko/redrawer/MetalRedrawer.kt
igordmn: Pavel answered that it is okay in DM

skiko/src/commonMain/kotlin/org/jetbrains/skiko/SkiaLayer.kt
```diff
@@ -55,9 +55,13 @@ expect open class SkiaLayer {
...
     /**
-     * Force redraw.
+     * Request redrawing of the content; The [renderDelegate] will be asked to re-render, and the result will be drawn
+     * on the screen.
+     *
+     * @param throttledToVsync Whether to throttle calling [renderDelegate]'s [SkikoRenderDelegate.onRender] to at most
+     * once between vsync signals (if vsync is enabled).
      */
-    fun needRedraw()
+    fun needRedraw(throttledToVsync: Boolean = true)
```
pjBooms, line 64: It seems to be public API change. Don't we have public API checkers for Skio?

igordmn: We don't, also we don't support backward compatibility due to Skiko is in 0.*.* version, and due to dependency on Skia that also doesn't support backward compatibility.

So far we haven't required having 2 reviewers for changes of API in Skiko

PR #1061, opened 2025-05-26
Update linux-amd64/Dockerfile with dependencies needed for cross compiling linuxArm64 target

Thomas-Vos:
https://github.com/JetBrains/skiko/pull/1051#discussion_r2107305732

### DIFF
--- a/skiko/docker/linux-amd64/Dockerfile
+++ b/skiko/docker/linux-amd64/Dockerfile
@@ -10,9 +10,9 @@ RUN apt-get update -y && \
     apt-get update && \
     apt-get install gcc-9 g++-9 -y && \
     update-alternatives --install /usr/bin/gcc gcc /usr/bin/gcc-9 60 --slave /usr/bin/g++ g++ /usr/bin/g++-9 && \
-    update-alternatives --config gcc \
-    apt-get install gcc-9-aarch64-linux-gnu g++-9-aarch64-linux-gnu -y \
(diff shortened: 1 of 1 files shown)


Thomas-Vos:
@eymar I am not that familiar with docker. I tried to build the Dockerfile locally on my Linux machine but it also fails on the same missing package. I am unsure why this package is missing in docker but works fine outside docker.

Thomas-Vos:
It seems that `gcc-aarch64-linux-gnu` is available but `gcc-9-aarch64-linux-gnu` is not for some reason. But we have to use version 9 to be able to build skiko, it fails with a newer version. The linux x64 build also uses gcc-9.

Thomas-Vos:
Ok, found the cause of missing dependency. The first line in the Dockerfile sets the ubuntu version to 14.04. This is very old and seems to be missing packages that are required. It can be fixed by updating this to for example 22.04 (same as GitHub CI).

@eymar  What's the reason behind this old version and is it safe to update it?

eymar:
afair, the old version was used for backward compatibility, so that skiko is compiled against an older stdlib. 

Would setting ubuntu 18.04 or 20.xy work?

Thomas-Vos:
That should be fine (image build is successful). Tomorrow I can prepare a PR for you with the updated Dockerfile.

Thomas-Vos:
@eymar Hey, please see https://github.com/JetBrains/skiko/pull/1062

PR #1062, opened 2025-05-28
Update linux-amd64 Dockerfile to Ubuntu 20.04

Thomas-Vos:
Needed to add `ARG DEBIAN_FRONTEND=noninteractive` so installation does not get stuck, copied this from another Dockerfile.
Also `binutils-2.26` was missing so replaced it with `binutils`, this is in line with the other docker files, but I am unsure if this could break anything.

On my PC the image build is successful (using `build.sh` script). The `run.sh` script is also successful after adapting its paths to my local setup. Unsure about publishing.

### DIFF
--- a/skiko/docker/linux-amd64/Dockerfile
+++ b/skiko/docker/linux-amd64/Dockerfile
@@ -1,11 +1,11 @@
-FROM ubuntu:trusty-20191217
-
+FROM ubuntu:20.04
+ARG DEBIAN_FRONTEND=noninteractive
 # Update binutils, g++, gcc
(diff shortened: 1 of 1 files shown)


eymar:
A comment for repository maintainers:

I believe that updating Ubuntu version to 20.04 here is fine, since we already use this version in skia-pack https://github.com/JetBrains/skia-pack/blob/main/.github/workflows/build.yml#L56

eymar (approved):

eymar:
The new docker image has been built successfully.

PR #1067, opened 2025-06-27
Upgrade Skiko to new build-helpers:publishing version

igordmn:
New sources are in https://github.com/JetBrains/compose-multiplatform/pull/5343

Was published in https://teamcity.jetbrains.com/buildConfiguration/JetBrainsPublicProjects_Compose_PublishBuildHelpers/5279403 from a branch

### DIFF
--- a/skiko/settings.gradle.kts
+++ b/skiko/settings.gradle.kts
@@ -12,13 +12,7 @@ pluginManagement {
             google()
         }
         dependencies {
-            // TODO Removing this makes publishing module below crash android internal plugin
-            classpath("org.ow2.asm:asm:9.6")
-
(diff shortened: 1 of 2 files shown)


skiko/settings.gradle.kts
```diff
@@ -12,13 +12,7 @@ pluginManagement {
             google()
         }
         dependencies {
-            // TODO Removing this makes publishing module below crash android internal plugin
-            classpath("org.ow2.asm:asm:9.6")
```
igordmn, line 16: This can still be needed. I'll return it in another PR, if CI will fail. It was added only because of some strange CI failure

terrakok (approved):

PR #1079, opened 2025-08-08
Fix Android sample crash.

colinrtwhite:
Fixes https://youtrack.jetbrains.com/issue/SKIKO-761/SkikoAndroidSample-crashes-with-skiko-v0.7.68

The main fix is adding `-latomic` ([reference](https://developer.android.com/ndk/guides/common-problems#undefined_reference_to_atomic_)). I'm not 100% why this fixes it, but I think the previous crash was due to some missing atomic operation support.

I also bumped the Android Gradle plugin, Gradle, and compile and target SDK versions to get it to work in Android Studio.


Siedlerchr:
I think you need to update the docker files as well https://github.com/JetBrains/skiko/blob/master/skiko/docker/linux-android-amd64/Dockerfile

PR #1082, opened 2025-08-13
Pack ANGLE into a separate jar

igordmn:
An addition to [the feature PR](https://github.com/JetBrains/skiko/pull/1017) that allows including ANGLE the same way as Skiko native libraries included:
- unpack it from an additional jar `org.jetbrains.skiko:skiko-awt-runtime-angle-$target:$version`
- find them in `skiko.library.path` directoty (a system property)
- find them in `java.home`

It reuses the same logic that is used to load `Skiko.dll`.

Currently only Windows supported, the same way as in the original PR.

No CI changes needed, as building of this jar is added as a dependency for `publishSkikoJvmRuntimeWindowsX64PublicationToComposeRepoRepository`/`publishSkikoJvmRuntimeWindowsArm64PublicationToComposeRepoRepository`

### Features - Desktop
A new experimental renderer is introduced for Windows. It uses [the ANGLE library](https://github.com/google/angle) which has proven to be stable because it is used inside Chromium, compared to the `DIRECT3D` renderer that has issues on a few machines. Note that `ANGLE` still uses `DIRECT3D` API under the hood.

1. Add this to the code to enable it:
```
System.setProperty("skiko.rendering.angle.enabled", "true")
```

2. Add the ANGLE library into the dependencies:
  - If you use Gradle:
  ```
  if (System.getProperty("os.name").startsWith("Win")) {
      implementation("org.jetbrains.skiko:skiko-awt-runtime-angle-$target:$version")
  }
  ```
  - If you set `skiko.library.path`. Extract `libEGL.dll`, `libGLESv2.dll` from https://github.com/JetBrains/angle-pack/releases into this directory.


samples/SkiaAwtSample/build.gradle.kts
```diff
@@ -40,6 +40,9 @@ dependencies {
     implementation("org.jetbrains.kotlinx:kotlinx-coroutines-test:1.5.0")
```
igordmn, line 40: Should be merged after https://github.com/JetBrains/skiko/pull/1017 and rebased on master

skiko/gradle.properties
```diff
@@ -1,10 +1,13 @@
 kotlin.code.style=official
 deploy.version=0.0.0
 
-
+# a tag from https://github.com/JetBrains/skia-pack
 dependencies.skia=m138-80d088a-1
 
-# you can override general skia dependencies by passing platform-specific property: 
+# a tag from https://github.com/JetBrains/angle-pack
```
igordmn, line 12: @Alovchin91, could you give Write access to https://github.com/JetBrains/angle-pack for https://github.com/orgs/JetBrains/teams/compose-multiplatform, so we can maintain it as well?

skiko/src/jvmMain/kotlin/org/jetbrains/skiko/LibraryLoader.kt
```diff
@@ -0,0 +1,131 @@
+package org.jetbrains.skiko
+
+import java.io.File
+import java.nio.file.Files
+import java.nio.file.StandardCopyOption
+
+internal class LibraryLoader(
```
igordmn, line 7: The code for this class is copied from the old Library.kt, except the signatures and places mentioned below

skiko/src/jvmMain/kotlin/org/jetbrains/skiko/LibraryLoader.kt
```diff
@@ -0,0 +1,131 @@
...
+    // A native library cannot be loaded in several classloaders, so we have to clone
+    // the native library to allow Skiko loading to work properly in complex cases, i.e.,
+    // several IDEA plugins.
+    private fun loadLibraryOrCopy(library: File): File? {
+        try {
+            System.load(library.absolutePath)
+            return null
+        } catch (e: UnsatisfiedLinkError) {
+            if (e.message?.contains("already loaded in another classloader") == true) {
+                val copyDir = Files.createTempDirectory("skiko").toFile()
```
igordmn, line 34: `copyDir` is now passed as a return value, not stored in the field. It gives us less error prone code.

skiko/src/jvmMain/kotlin/org/jetbrains/skiko/LibraryLoader.kt
```diff
@@ -0,0 +1,131 @@
...
+                file.deleteOnExit()
+            Library::class.java.getResourceAsStream("/$resourceName").use { input ->
+                Files.copy(input, tempFile.toPath(), StandardCopyOption.REPLACE_EXISTING)
+            }
+            Files.move(tempFile.toPath(), file.toPath(), StandardCopyOption.ATOMIC_MOVE)
+        }
+        return file
+    }
+
+    private var isLoaded = false
```
igordmn, line 60: Now it is not atomic (no need because it is inside Synchronized)

skiko/src/jvmMain/kotlin/org/jetbrains/skiko/LibraryLoader.kt
```diff
@@ -0,0 +1,131 @@
...
+
+    /**
+     * Load a native library finding it in multiple sources:
+     * - from SkikoProperties.libraryPath
+     * - java.home
+     * - jar resources
+     *
+     * @throws LibraryLoadException if library wasn't loaded successfully.
+     *         Calling this function again retries the loading.
```
igordmn, line 69: This is a behavior change for Skiko.dll. Previously any error during loading fails the loading forever.

Now if we call it again and there was an error previously, it tries to load it again.

skiko/src/awtTest/kotlin/org/jetbrains/skiko/util/UiTest.kt
```diff
@@ -19,7 +19,7 @@ internal fun uiTest(
 
     runBlocking(MainUIDispatcher) {
         if (renderApiProperty == "all") {
-            for (renderApi in SkikoProperties.fallbackRenderApiQueue(SkikoProperties.renderApi)) {
+            for (renderApi in SkikoProperties.fallbackRenderApiQueue(initialApi = null)) {
```
igordmn, line 24: This ensures that we test all the defined renderers, not only starting from the default one

igordmn:
CI checked here: https://github.com/JetBrains/skiko/pull/1083

skiko/gradle.properties
Alovchin91: Yup, done 👍

PR #1085, opened 2025-08-14
Update SkikoJsSample and rename it to SkikoWebSample

Schahen:
This PR updates SkikoJsSample (which is renamed to SkikoWebSample) in following aspects:

- By default sample built against release version so that user can check the sample without local build at all
- When build against local skiko, `skiko.version` parameter is ignored
- Sample code commonized and prepared for future introduction of common web source sets in Kotlin 2.2.20


samples/SkiaWebSample/build.gradle.kts
```diff
@@ -8,34 +8,37 @@ repositories {
...
 val resourcesDir = "$buildDir/resources/"
 
-val skikoWasm by configurations.creating
-
 val isCompositeBuild = extra.properties.getOrDefault("skiko.composite.build", "") == "1"
 
+// When we build skiko locally we have no say in setting skiko.version in the included build.
+// That said, it is always built as "0.0.0-SNAPSHOT" and setting any other version not only is misleading,
+// but even can create conflict due to incompatibility of skiko runtime and skiko libs
+val skikoVersion = if (isCompositeBuild) "0.0.0-SNAPSHOT" else project.properties["skiko.version"]
```
igordmn: When I set `skiko.composite.build`, `gradlew wasmJsBrowserDevelopmentRun` fails on my machine (Windows):
```
> Failed to query the value of task ':kotlinNpmInstall' property 'packageJsonFiles'.
   > Could not resolve all dependencies for configuration ':jsNpmAggregated'.
      > Could not resolve org.jetbrains.skiko:skiko:0.9.23.
        Required by:
            root project :
         > No matching variant of project :skiko was found. The consumer was configured to find a library for use during 'kotlin-runtime', preferably optimized for non-jvm, as well as attribute 'org.jetbrains.kotlin.js.compiler' with value 'ir', attribute 'org.jetbrains.kotlin.js.public.package.json' with value 'public-package-json', attribute 'org.jetbrains.kotlin.platform.type' with value 'js' but:
```

When I run via IDE:
```
Could not determine the dependencies of task ':wasmJsPackageJson'.
> Task with path ':wasmRootPackageJson' not found in project ':skiko'.
```

Though, the other samples (old `SkiaJsSample`, `SkiaMultiplatformSample` has the same error)

Did you test it, does it work for you?

Schahen: I tested it but not on window machine, will recheck tomorrow and let you know

Schahen: @igor, what exactly command are you running (just to double check)

igordmn (approved):

skiko/gradle/wrapper/gradle-wrapper.properties
```diff
@@ -1,6 +1,6 @@
 distributionBase=GRADLE_USER_HOME
 distributionPath=wrapper/dists
-distributionUrl=https\://services.gradle.org/distributions/gradle-8.13-bin.zip
+distributionUrl=https\://services.gradle.org/distributions/gradle-8.14.3-all.zip
```
MatkovIvan, line 3: I'd keep it 8.13-bin across the projects for now (in case if there is no explicit reason)

Schahen: @MatkovIvan sure, let me double check both this and kotlin version after it was updated in seiko

PR #1088, opened 2025-08-26
Decouple onRender from draw. Additional changes

igordmn:
- restore `withContext(dispatcherToBlockOn)`. It performs rendering in background without blocking the main thread. SkiaLayerPerformanceTest caught a regression in FPS cadence
- Change scheduling of frames to fix [these issues](https://github.com/JetBrains/skiko/pull/1057#discussion_r2121014191)

### DIFF
--- a/skiko/src/awtMain/kotlin/org/jetbrains/skiko/redrawer/MetalRedrawer.kt
+++ b/skiko/src/awtMain/kotlin/org/jetbrains/skiko/redrawer/MetalRedrawer.kt
@@ -121,6 +121,9 @@ internal class MetalRedrawer(
 
     private suspend fun draw() {
         inDrawScope {
+            // Move drawing to another thread to free the main thread
+            // It can be expensive to run it in the main thread and FPS can become unstable.
+            // This is visible by running [SkiaLayerPerformanceTest], standard deviation is increased significantly.
             withContext(dispatcherToBlockOn) {
                 performDraw()
             }


skiko/src/awtMain/kotlin/org/jetbrains/skiko/redrawer/MetalRedrawer.kt
```diff
@@ -220,4 +189,49 @@ internal class MetalRedrawer(
      * @note see https://developer.apple.com/documentation/quartzcore/cametallayer/2887087-displaysyncenabled
      */
     private external fun setDisplaySyncEnabled(device: Long, enabled: Boolean)
+
+    private class MetalFrameDispatcher(
+        private val update: () -> Unit,
+        private val draw: suspend CoroutineScope.() -> Unit,
+        private val isShowing: () -> Boolean,
+        private val waitForVSync: suspend CoroutineScope.() -> Unit,
+    ) {
```
m-sasha: I like the grouping of related functions and properties this class creates, but I don't like
1. The name - it's confusing as we already have `FrameDispatcher`, and this class doesn't implement it. Also, there's no need to repeat the word "Metal", as we're already inside `MetalRedrawer`. Maybe `FrameScheduler`?
2. The arguments create an abstraction, but it's not needed here, as the class is only used in one place. Any abstraction complicates understanding. I suggest making it `inner` instead, and calling the functions in `MetalRedrawer` directly.

skiko/src/awtMain/kotlin/org/jetbrains/skiko/redrawer/MetalRedrawer.kt
```diff
@@ -149,16 +112,24 @@ internal class MetalRedrawer(
...
+                    performDraw()
+                }
+            }
         }
     }
 
     private suspend fun draw() {
-        performDraw()
+        inDrawScope {
+            withContext(dispatcherToBlockOn) {
```
m-sasha, line 127: Are you sure we need this? `draw` never waits for vsync anymore, so why do we need `dispatcherToBlockOn`? It seems like an unnecessary context (thread) switch, which could be quite expensive, no?
Is it because the drawing itself takes a long time?

igordmn: Yes, because the draw calls before vsync can be expensive too and block the main thread (even if they just schedule work for GPU). It was revealed by `SkiaLayerPerformanceTest`. The FPS isn't stable without this change. Instead of `16,17,18,16,16,17,18` we had `16,20,10,16,21,9`.

A similar experimental improvement was developed for iOS that also uses the Metal API

skiko/src/awtMain/kotlin/org/jetbrains/skiko/redrawer/MetalRedrawer.kt
igordmn: Done

skiko/src/awtMain/kotlin/org/jetbrains/skiko/redrawer/MetalRedrawer.kt
m-sasha: I see. I suggest adding a comment to explain that, then.

igordmn: Added

PR #1094, opened 2025-09-10
Enable wasm skiko artefact building whenever we are building it for a sample in composite mode

Schahen:
As of now whenever we run web samples with `skiko.composite.build=1` from Idea we enforce enabling relevant targets -but this was not the case for running from Terminal or any other non-Idea environment (without Idea present)

This can confuse users and require from them an additional level of mental effort to figure out what's going on.

The goal of this PR is to solve this problem via approaching the problem similarly to IDEA-backed configuration - that is - enabled the relevant property when build is composite


Schahen:
Closing till furhter  improvements

PR #1103, opened 2025-09-15
Fix "ComposePanel doesn't use hardware acceleration with RenderSettings.SwingGraphics on Linux"

igordmn:
Fixes https://youtrack.jetbrains.com/issue/CMP-8936

`glXChooseFBConfig` fails to find a config with `GLX_DOUBLEBUFFER, False` on my machine, it only has double buffered configs. But double/single buffering doesn't matter as we use pixel buffers.

Also choose a correct config that is required by https://registry.khronos.org/OpenGL-Refpages/gl2.1/xhtml/glXCreatePbuffer.xml:
```
BadMatch is generated if config does not support rendering to pixel buffers (e.g., GLX_DRAWABLE_TYPE does not contain GLX_PBUFFER_BIT)
```

## Testing
- manually, in the clocks example
- the modified test failed before the fix, now it doesn't fail

## Release Notes (Skiko)
Fix "SkiaSwingLayer doesn't use hardware acceleration on Linux"

### Fixes - Desktop 
Fix "ComposePanel doesn't use hardware acceleration with RenderSettings.SwingGraphics on Linux"

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -286,7 +286,7 @@ jobs:
           java-version: '21'
           cache: 'gradle'
 
-      # OPENGL is ignored as it doesn't exist on GitHub actions agents
+      # OPENGL is ignored as it doesn't exist on the Windows GitHub agent
       - shell: bash
         name: 'Compile and run AWT tests'
         run: ./gradlew --stacktrace --info -Pskiko.test.onci=true -Pskiko.test.ui.renderApi.ignoreAssertsFor=OPENGL :skiko:awtTest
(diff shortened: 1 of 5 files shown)


skiko/src/awtTest/kotlin/org/jetbrains/skiko/SkiaLayerTest.kt
```diff
@@ -386,6 +398,10 @@ class SkiaLayerTest {
             window3.toFront()
             delay(1000)
             screenshots.assert(window3.bounds, "window3")
+
+            assertRenderApiFor(window1.layer)
```
igordmn, line 402: This is added to not miss the case when the test fallbacks to the next API due to not supporting it

.github/workflows/ci.yml
```diff
@@ -238,7 +238,7 @@ jobs:
           sudo apt-get remove g++ -y
           sudo update-alternatives --install /usr/bin/clang clang /usr/bin/clang-11 100
           sudo update-alternatives --install /usr/bin/clang++ clang++ /usr/bin/clang++-11 100
-          sudo Xvfb :0 -screen 0 1280x720x24 &
+          sudo Xvfb :0 -screen 0 1280x720x24 +extension GLX &
```
MatkovIvan, line 241: Most likely you'll need to add this in compose CI after this change

igordmn: https://jetbrains.team/p/ui/reviews/53/timeline

PR #1104, opened 2025-09-18
Fix drawing issues and add API to request render-and-draw explicitly

m-sasha:
1. Align `SkiaLayer` and `HardwareLayer`'s paint implementations to only call `needRedraw`
2. Remove `updateNeeded` parameter from `Redrawer.redrawImmediately`, as it's now always `true`
3. Add a public API, `SkiaLayer.renderAndDrawImmediately`
4. Fix `MetalRedrawer.rerdawImmediately` to actually update and draw synchronously.

### DIFF
--- a/skiko/src/awtMain/kotlin/org/jetbrains/skiko/SkiaLayer.awt.kt
+++ b/skiko/src/awtMain/kotlin/org/jetbrains/skiko/SkiaLayer.awt.kt
@@ -557,6 +557,8 @@ actual open class SkiaLayer internal constructor(
         redrawer?.needRender(throttledToVsync)
     }
 
+    actual fun needRedraw() = needRender()
+
     internal fun update(nanoTime: Long) {
         check(isEventDispatchThread()) { "Method should be called from AWT event dispatch thread" }
         check(!isDisposed) { "SkiaLayer is disposed" }
--- a/skiko/src/androidMain/kotlin/org/jetbrains/skiko/SkiaLayer.android.kt
+++ b/skiko/src/androidMain/kotlin/org/jetbrains/skiko/SkiaLayer.android.kt
@@ -63,6 +63,8 @@ actual open class SkiaLayer {
         }
     }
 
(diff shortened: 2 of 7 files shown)


skiko/src/awtMain/kotlin/org/jetbrains/skiko/SkiaLayer.awt.kt
```diff
@@ -416,24 +416,29 @@ actual open class SkiaLayer internal constructor(
...
+                latestReceivedGraphicsContextScaleTransform,
+                currentGraphicsContextScaleTransform
+            )
         }
     }
 
+    /**
+     * Updates the layer and redraws synchronously.
+     */
+    fun updateAndDrawImmediately() {
```
igordmn, line 438: Please, call it inside the Skiko update PR instead of `paint()` to pass the Compose tests

igordmn, line 438: The name should be the same as in `needRedraw`, so we need to rename it to `needUpdateAndDraw` (with deprecation).

Or rename both to `renderImmediatelly`/`needRender` (because we have `onRender` callback)

m-sasha: I renamed both. Do I need to keep backwards compatibility?

igordmn: We don't support it for Skiko, but deprecation for a popular method (`needRedraw`) is useful. Set TODO to remove it after 2026-07

m-sasha: `SkiaLayer` is an `expect` class, so I was forced to put 
```
actual fun needRedraw() = needRender()
```
in every `actual` class.

PR #1106, opened 2025-09-20
Use clang instead of clang-11 in linux-arm CI

m-sasha:


igordmn (approved):

MatkovIvan (commented):
Changing it ONLY here is incorrect. Please update [Dockerfile](https://github.com/JetBrains/skiko/blob/master/skiko/docker/linux-arm64/Dockerfile#L9) too and rebuild the image that is used to build the relase

m-sasha:
> use something like clang-12
I don't know about 12, but [tried clang-17 and clang-19](https://github.com/JetBrains/skiko/pull/1105) and they didn't work either.

PR #1108, opened 2025-09-29
Fix invocation of tests that only run on CI

m-sasha:


skiko/src/awtTest/kotlin/org/jetbrains/skiko/SkiaLayerPerformanceTest.kt
```diff
@@ -179,6 +188,7 @@ class SkiaLayerPerformanceTest {
     }
 
     @Test
+    @Ignore
```
igordmn: - why we ignore only one test? the test below is more heavy, why does it pass?
- it it also fails, it is better to just to pass `skiko.test.performance.enabled`, because it is often needed to run these tests manually

m-sasha: > why we ignore only one test? the test below is more heavy, why it pass?

I don't know, but it does pass.

> it it also fails, it is better to just to pass skiko.test.performance.enabled, because it is often needed to run these tests manually

Let me know what you prefer.

igordmn: Let's use `skiko.test.performance.enabled` and remove `Ignore`

igordmn (approved):

.github/workflows/ci.yml
```diff
@@ -170,21 +170,25 @@ jobs:
...
         name: 'Compile and run AWT tests'
         run: |
           export DISPLAY=:0
-          ./gradlew --no-daemon --stacktrace --info -Pskiko.native.enabled=true -Pkotlin.native.cacheKind.linuxX64=none -Pskiko.test.onci=true :skiko:awtTest
+          ./gradlew --no-daemon --stacktrace --info -Pskiko.native.enabled=true -Pkotlin.native.cacheKind.linuxX64=none -Dskiko.test.onci=true -Dskiko.test.performance.enabled=false :skiko:awtTest
         timeout-minutes: 25
 
+      - name: Clean Gradle cache to recover disk space
+        run: |
+          rm -rf ~/.gradle/caches
```
MatkovIvan, line 190: It's totally wrong approach to do that. It invalidates the purpose of uploading gradle caches entirely.

It had to be rejected, not merged. Please don't keep it like that in master

PR #1110, opened 2025-10-07
Fix 16kb page size load alignment for Android

Siedlerchr:
As of November Google requires 16kb page sizes for native libraries.

https://developer.android.com/guide/practices/page-sizes

Updated the ndk version in the gradle and docker files

Tested this with the sample app and also with a more complex project.

Fixes [SKIKO-1064](https://youtrack.jetbrains.com/issue/SKIKO-1064)

### DIFF
--- a/samples/SkiaAndroidSample/settings.gradle.kts
+++ b/samples/SkiaAndroidSample/settings.gradle.kts
@@ -6,5 +6,4 @@ pluginManagement {
     }
 
 }
-
 rootProject.name = "SkiaAndroidSample"


samples/SkiaAndroidSample/gradle.properties
```diff
@@ -1 +1,2 @@
 # android.useAndroidX=true
+#skiko.version=0.9.26
```
kropp: Seems to be a leftover

skiko/buildSrc/src/main/kotlin/tasks/configuration/JvmTasksConfiguration.kt
```diff
@@ -327,6 +327,9 @@ fun SkikoProjectContext.createLinkJvmBindings(
                 "-llog",
                 "-landroid",
                 "-latomic",
+                "-Wl",
+                "-z",
+                "max-page-size=16384",
```
kropp: Documentation on the linked page https://developer.android.com/guide/practices/page-sizes#compile-r28 states that NDK r28+ always link with 16kb pages, do we still need to set these flags?

kropp:
Thank you for your contribution!
We will soon be requiring a YouTrack ticket for all PRs, I created one for this: [SKIKO-1064](https://youtrack.jetbrains.com/issue/SKIKO-1064) Support 16 KB page sizes on Android
It is intended to be used for all feature discussions

skiko/buildSrc/src/main/kotlin/tasks/configuration/JvmTasksConfiguration.kt
Siedlerchr: You are right, the flags are not needed. I tested it again now and it seems I had some mix-up with the ndk version resolution order on my system

samples/SkiaAndroidSample/gradle.properties
Siedlerchr: fixed

samples/SkiaAndroidSample/build.gradle.kts
```diff
@@ -39,7 +39,7 @@ val unzipTaskArm64 = tasks.register("unzipNativeArm64", Copy::class) {
 }
 
 kotlin {
-    jvmToolchain(11)
+    jvmToolchain(17)
```
kropp: Please also update `DEVELOPMENT.md` in the root folder.

kropp: Is it required by the Android NDK r29?

samples/SkiaAndroidSample/settings.gradle.kts
```diff
@@ -6,4 +6,7 @@ pluginManagement {
     }
 
 }
+plugins {
+    id("org.gradle.toolchains.foojay-resolver-convention") version "0.8.0"
+}
```
kropp: I'd prefer to avoid adding new dependencies unless strictly necessary.
Building Skiko already requires a sophisticated environment setup described in the DEVELOPMENT.md, so I believe it is enough to just update the instruction there with the new JDK version.

samples/SkiaAndroidSample/build.gradle.kts
Siedlerchr: This is/was just from the sample app. Independent of skiko. But I am happy to revert this.

samples/SkiaAndroidSample/settings.gradle.kts
Siedlerchr: The plugin can download the Java toolchain specified by jvmToolchain. However, as it was only in the sample app I reverted it.  https://docs.gradle.org/current/userguide/toolchains.html#sub:download_repositories

kropp: Yeah, thanks, I know the plugin. Just wanted to avoid adding unnecessary things to the build.

kropp (approved):
So in the end this PR just updates the NDK. LGTM

PR #1184, opened 2026-03-24
[DO NOT MERGE]: Introduce Graphite Native Metal

hub-bla:
[SKIKO-1128](https://youtrack.jetbrains.com/issue/SKIKO-1128)

This PR introduces the initial version of the Graphite API in Skiko using native Metal.
Please note that a separate PR for Graphite packaging must be merged before this one can be finalized

### DIFF
--- a/skiko/src/commonMain/kotlin/org/jetbrains/skia/Surface.kt
+++ b/skiko/src/commonMain/kotlin/org/jetbrains/skia/Surface.kt
@@ -1,7 +1,10 @@
 package org.jetbrains.skia
 
+import org.jetbrains.skia.gpu.graphite.BackendTexture
+import org.jetbrains.skia.gpu.graphite.Recorder
 import org.jetbrains.skia.impl.*
 import org.jetbrains.skia.impl.Library.Companion.staticLoad
+import org.jetbrains.skiko.ExperimentalSkikoApi
 
 class Surface : RefCnt {
     companion object {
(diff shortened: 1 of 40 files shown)


skiko/src/commonMain/kotlin/org/jetbrains/skia/Surface.kt
```diff
@@ -221,6 +224,37 @@ class Surface : RefCnt {
                 reachabilityBarrier(imageInfo.colorInfo.colorSpace)
             }
         }
+
+        @ExperimentalSkikoApi
+        fun makeFromBackendTexture(
```
MatkovIvan: An option to avoid using `graphite` package here - make it an extension function inside `graphite` package

```kt
fun Surface.Companion.makeFromBackendTexture(...)
```

skiko/src/commonMain/kotlin/org/jetbrains/skiko/context/ContextHandler.kt
```diff
@@ -48,3 +52,47 @@ internal abstract class ContextHandler(
...
+        renderTarget?.close()
+    }
+
+    override fun rendererInfo(): String {
+        return "GraphicsApi: Ganesh ${layer.renderApi}\n" +
+                "OS: ${hostOs.id} ${hostArch.id}\n"
+    }
+}
+
+internal abstract class GraphiteContextHandler(
```
MatkovIvan, line 78: Should we move it into `graphite` package?

skiko/src/commonMain/kotlin/org/jetbrains/skiko/SkikoFlags.kt
```diff
@@ -0,0 +1,7 @@
+package org.jetbrains.skiko
+
+import org.jetbrains.skia.gpu.SkiaGpuBackend
+
+object SkikoFlags {
```
MatkovIvan, line 5: Currently flags are platform-dependent (built-in properties on Desktop).

I see it's used only in `MacOsMetalRedrawer`, so maybe it might be regular argument there

skiko/src/commonMain/kotlin/org/jetbrains/skia/Surface.kt
hub-bla: Good idea, I'll move it

skiko/src/commonMain/kotlin/org/jetbrains/skiko/context/ContextHandler.kt
hub-bla: We can but ContextHandler can't be sealed then. If thats fine then sure

skiko/src/commonMain/kotlin/org/jetbrains/skiko/SkikoFlags.kt
hub-bla: It's currently used only in `MacOsMetalRedrawer` because I only made the Graphite-based context handler for native macOS for now. In general I think that flag (or flags) related to GPU backend should be available to all platforms so we know at runtime which backend is available

PR #1185, opened 2026-03-27
Update Kotlin to 2.3.20

MatkovIvan:
[SKIKO-1067](https://youtrack.jetbrains.com/issue/SKIKO-1067) Build Skiko with Kotlin 2.3.0


Schahen (commented):
I was going  to introduce some web-related changes but actually absolutely nothing stops me doing so after we'll have this in master, so thank you for taking care of it

eymar (approved):

PR #1186, opened 2026-03-31
Patch skia symbols for iOS and tvOS targets

ASalavei:
Add the step to patch all C/C++ symbols for iOS and tvOS targets to prevent symbol clashes when the user's app uses another skia library, or one of its dependencies.

Fixes: https://youtrack.jetbrains.com/issue/CMP-6845/How-to-use-compose-multiplatform-on-iOS-in-app-with-lib-skia-dependency


skiko/buildSrc/src/main/kotlin/tasks/configuration/NativeTasksConfiguration.kt
```diff
@@ -231,7 +229,20 @@ fun SkikoProjectContext.configureNativeTarget(os: OS, arch: Arch, target: Kotlin
 
     val bridgesLibrary = layout.buildDirectory.file("nativeBridges/static/$targetString/skiko-native-bridges-$targetString.a")
     val bridgesLibraryPath = bridgesLibrary.get().asFile.absolutePath
-    val allLibraries = skiaStaticLibraries(skiaDir, targetString, buildType) + bridgesLibraryPath
+
+    // For iOS/tvOS we patch every library so that all public Skia symbols have
+    // postfix with "_skiko", preventing conflicts when multiple Skia copies
```
eymar: How does it affect the binary size?

.github/actions/setup-native/action.yml
```diff
@@ -0,0 +1,18 @@
+name: 'Setup Native Build Environment'
```
MatkovIvan: Why is it not a part of `actions/setup-prerequisites`?

skiko/buildSrc/scripts/patch_skia_symbols.py
```diff
@@ -0,0 +1,311 @@
+#!/usr/bin/env python3
+"""
+Patches public symbols in Skia static libraries for iOS/tvOS to avoid symbol
+conflicts when multiple copies of Skia (or libraries depending on Skia) are
+linked into the same app binary.
```
MatkovIvan: Why should it be in this repo instead of in skia build scripts directly? Since we are already handling some suffixes there, it seems it might be just an addition there

eymar: as I understand: the reason is that not only the functions should be renamed, but also the call sites in skiko bindings. 

If the symbols were renamed on skia side, then we'll need one more source set with ios/tvos-only bindings using _suffixed names. We don't really want that.

ASalavei: The script should rename also symbols in c++/kotlin bindings library

skiko/buildSrc/src/main/kotlin/tasks/configuration/NativeTasksConfiguration.kt
ASalavei: Binary increased for 128Kb (131945 bytes, 0.25%) for iOS Arm64 target in klib and 2,85Mb (2994968 bytes, 1,9%) when extracted.
I tried other ways, like replacing "_skiko" with something like "_skk" does its job, but still increases the size for 100Kb. Also, we can use a very simple solution like adding an underscore, but we will increase change of unwanted clash of symbols.
So, if we want to keep the stacktrace readable - that's we need to stick to some kind of this solution. Otherwise we can just freely rename all the symbols and safe ~1Mb.

eymar: thanks for checking. it's good to know how such changes affect the size.

.github/actions/setup-prerequisites/action.yml
```diff
@@ -14,6 +14,11 @@ runs:
         sudo rm -rf /opt/hostedtoolcache/CodeQL
         df -h
 
+    - name: Install LLVM
+      if: runner.os == 'macOS'
+      shell: bash
+      run: brew install llvm
```
eymar, line 20: Perhaps, pinning the llvm version could help with CI stability (to avoid unwanted updates and unexpected issues). WDYT?

Besides that, do you plan to update the TeamCity build config for skiko too? (TC is responsible for publication)

eymar (approved):

skiko/buildSrc/scripts/patch_skia_symbols.py
MatkovIvan: Why cannot we compile it (including bindings) with prefixes in the first place? I mean documented compilation flags sound much safer than patching the result with a custom script

ASalavei: What compilation flags you're talking about? I didn't find other approaches then the current one.

MatkovIvan: Preprocessor defines like we already use here

https://github.com/JetBrains/skia/blob/a42fe4521e89fd52c11edb2456636c55c36a71a3/bazel/external/icu/BUILD.bazel#L1060

ASalavei: That seems like a feature of the icu library. Does similar option present in other libraries except ICU?

.github/actions/setup-prerequisites/action.yml
ASalavei: Good point. Usually LLVM got installed with XCode, but it is not true for GH actions. I will check publications on TC and pin the version here.

skiko/buildSrc/scripts/patch_skia_symbols.py
```diff
@@ -0,0 +1,312 @@
+#!/usr/bin/env python3
```
MatkovIvan: The main thing that worries me is that we’re adding a custom post-link binary rewrite step in `buildSrc`, with Python (we don't have it at all in this repo now), `llvm-objcopy`, and ABI-specific symbol rewriting, and making normal iOS/tvOS builds depend on it. That feels pretty heavy and brittle to maintain for something that is essentially a packaging/compatibility workaround.

I'd like to hear @igordmn opinion before merging it

skiko/buildSrc/scripts/patch_skia_symbols.py
MatkovIvan: Probably in some, but not in all.

One more thing to explore with compilation flags is:
- Generate a header with renaming macros
- Force-include it via compiler flags
- Let the compiler produce the renamed symbols directly

ASalavei: Already tried - does not work because several libraries use own macros to define joined symbol names on the compilation phase. In this case macros substitution does not work.

skiko/buildSrc/scripts/patch_skia_symbols.py
ASalavei: > we don't have it at all in this repo now

There is a `skia-pack` dependency repository which uses python. Technically we can move the script there, but this script must be applied after c++ bridging library is built in the `skiko` repository. So it's not very correct to have it there.

> That feels pretty heavy and brittle to maintain for something that is essentially a packaging/compatibility workaround

I would tread it as a part of `skiko` build requirement for iOS/tvOS platforms, including the `llvm-objcopy`. The script itself can be re-written on Kotlin. WDYT?

MatkovIvan: > There is a skia-pack dependency repository

Not anymore. It's archived. Scripts are in skia fork repo now

> The script itself can be re-written on Kotlin. WDYT?

It's probably a better fit for this repo 🤔

igordmn: > I'd like to hear @igordmn opinion before merging it

Heavy build code worries me to. Though the issue is probably serious enough to justify it.

Before proceeding with it, are there any more simple alternatives? Maybe passing some arguments to the compiler to hide symbols?

If we still have to have it, better to rewrite it on Kotlin and embed it into the buildSrc code.

MatkovIvan: I like the idea of hiding instead of renaming symbols, that would seem closer to the actual intent: only the bindings layer needs to stay visible.
AI suggest this way to implement this:
> 1. Build the bridge normally, the way Skiko already does
> 2. Take all Skia archives plus the bridge archive and partially link them into one relocatable Mach-O object with `ld -r`
> 3. Keep only the bridge API symbols global, and turn everything else local/private with `nmedit -s exported_symbols.txt` or possibly `llvm-objcopy --keep-global-symbols`.
> 4. Put that relocatable object into a final `.a` and ship only that one to Kotlin/Native.

It looks much simpler than the current solution, because it avoids custom symbol rewriting logic, manual parsing of mangled names, and ABI-specific patching rules.

ASalavei: Made some tests:
The solution proposed by @igordmn also works, but increases application binary size from 25.3 Mb to 28.3 Mb which is not acceptable. From another side, current solution does not affect the iOS application size (but increases skiko klib)
Here is the branch with prototype: https://github.com/JetBrains/skiko/tree/andrei.salavei/single-skiko-lib

igordmn: >  but increases application binary size from 25.3 Mb to 28.3 Mb which is not acceptable.

Do you know what causes the increase?

Reading the code, we do linking twice in 2 tasks via `libtool`:

```
    private fun archiveLibrary(outDir: File, mergedObj: File) {
        execOperations.exec {
            executable = "libtool"
            args = listOf(
                "-static",
                "-o", File(outDir, outputFileName.get()).absolutePath,
                mergedObj.absolutePath,
            )
        }
    }
...

    val linkTask = project.registerSkikoTask<Exec>(actionName, os, arch) {
...
            OS.MacOS, OS.IOS, OS.TVOS -> {
                executable = "libtool"
                argumentProviders.add { listOf("-static", "-o", staticLib) }
            }

...
    dependsOn(mergeTask ?: linkTask)
```

Not sure if this is the cause, but worth to try linking only once.

igordmn: Also, at some point in the build pipeline, we should do unused code stripping. In theory, by hiding symbols we should strip more code, and make binary size smaller. But I don't know if this is actually possible.

ASalavei: @igordmn , I checked if we can remove the first linkage, but it didn't reduce the binary size.
Why it so - the final linker (those one who build iOS app) is able to strip only exposed symbols. In case of the single library approach, the linker cannot effectively optimize dead symbols.

igordmn: > the final linker (those one who build iOS app) is able to strip only exposed symbols.

Can we strip these symbols on this stage, during hiding?

If it requires long investigation though, we can proceed with the current approach. Which disadvantages we can have, besides more difficulties in maintaining it?

ASalavei: > Can we strip these symbols on this stage, during hiding?

On this stage - no. Compiler already strips as much as it can. I tried several ways to minimize the size, but the issue is that that the best optimization is possible only at the final stage.

>  Which disadvantages we can have, besides more difficulties in maintaining it?

Mismatch between symbols in code and debugger / crashlog. The solution can be simplified if we won't take mangled symbols into account, but then the automatic un-mangling may cause issues.
Also there is a very low chance that we missed some symbols that potentially may clash.

igordmn: Ok, then it looks fine for me to have this approach under our maintenance. 

I let the original reviewers to review the implementation.

skiko/buildSrc/src/main/kotlin/PatchSkiaSymbolsTask.kt
```diff
@@ -0,0 +1,337 @@
...
+        //     their original names and could clash with a second copy of Skia
+        //     linked into the same binary.  We rename ONLY symbols whose Mach-O
+        //     name starts with "__Z".
+        logger.lifecycle("Extracting C++ symbols from skiko bridge …")
+        val bridgeSyms = extractGlobalDefinedSymbols(bridgeFile)
+        val bridgeCxxSyms = bridgeSyms.filterTo(mutableSetOf()) {
+            it.startsWith("__Z") && !it.endsWith("_skiko")
+        }
+        val newInBridge = bridgeCxxSyms - allSymbols
+        allSymbols.addAll(bridgeCxxSyms)
```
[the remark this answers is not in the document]
ASalavei: it should be no such symbols in the skia/bridge codebase.

.github/actions/setup-prerequisites/action.yml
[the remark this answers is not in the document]
ASalavei: Disagree with the fix. If we need proper LLVM configuration only for ios/tvos tasks it must be done in a separate action. cc @MatkovIvan

MatkovIvan (approved):

eymar (approved):

sargunv:
I suspect this fix will resolve https://github.com/maplibre/maplibre-native/issues/3777, https://github.com/maplibre/maplibre-compose/issues/551, and https://youtrack.jetbrains.com/issue/CMP-8882/ when it's included in Compose Multiplatform.

PR #1188, opened 2026-04-09
Optimize Classes handling between Skia C++ and Compose consumers

ApoloApps:
Migrates  most Enum classes to use Value classes to avoid object allocations between Compose and Skia Native. Compose uses value classes a lot and most of the times (at least I have not yet found the opposite), it shares the representing object with Skiko (in most cases Integer ordinals). Compose could map to these value classes and since Integer is the same object, we would save an object allocation between Compose and Skia Native (Skiko creates lots of Skiko enum classes redundantly) 
Cons: 
- we lose exhaustiveness in few places -> In fact, we already have `else` branches since Compose also uses Value classes so this is not a problem. Also, I have checked almost all public usages of these APIs using https://grep.app/search?f.lang=Kotlin and I have not seen actual when statements. Only used in Compose. Most public repos only use just 1 value of the classes.

I maintain names in places where maybe are needed for compatibility. I do think there should be no problem.

We avoid: object allocations between Compose -> Skia (reusing Compose's objects

Future Steps
1. Compose PR to replace `when `statements with a single inline value class constructor (value classes have internal constructors so no outside not-specified param value)
2. When Kotlin/JS enables Long -> BigInt conversion by default, optimize classes with 2 Integers or 2 Floats to instead be stored in a single Long using shifting bits and value classes (`org.jetbrains.skia.Point` , `org.jetbrains.skia.FilterMipmap`, `org.jetbrains.skia.shaper.ScriptRun`, `org.jetbrains.skia.paragraph.PositionWithAffinity`, `org.jetbrains.skia.paragraph.TextIndent`, .....)


m-sasha:
I'm not sure I understand the problem being solved here.

> to avoid object allocations between Compose and Skia Native.

Enum objects are only ever allocated once...

ApoloApps:
> only ever allocated once...

This way it is never allocated in the first place. It also allows to simplify redundant logic in Compose (toSkiaWhateverObject functions could be simplified by making it inline and just passing Enum(composeenum.value), which would eliminate all that logic in between since it represents same underlying Integer value) 
```kotlin
    private fun StrokeCap.toSkia() = when (this) {
        StrokeCap.Butt -> SkPaintStrokeCap.BUTT
        StrokeCap.Round -> SkPaintStrokeCap.ROUND
        StrokeCap.Square -> SkPaintStrokeCap.SQUARE
        else -> SkPaintStrokeCap.BUTT
    }
//Change to
  private inline fun StrokeCap.toSkia() = SkPaintStrokeCap(SkPaintStrokeCap.value)

//Which would be eliminated/inlined when compiling because it is pointing to same value/class

//(the ordinal or value backing value class param)
``` 
Removes redundant intermediate logic and prevents allocation in the first place. It also aligns this behaviour with Compose (use value class whenever it can be done to avoid heavy allocations in unnecessary places). Please note that this PR also change places that are not enum classes only. Some interfaces/classes (redundant) are also migrated to value classes to avoid all above stated

**Edit:**
In addition to all this above, as stated in **Future Steps**, this would be used in other Data classes that take 2 previously Enum classes (org.jetbrains.skia.Point , org.jetbrains.skia.FilterMipmap, org.jetbrains.skia.shaper.ScriptRun, org.jetbrains.skia.paragraph.PositionWithAffinity, org.jetbrains.skia.paragraph.TextIndent, .....) and transform them also to Value classes which would avoid allocation by packing them into a single Long

**Edit 2:**
it also simplifies an Enum Entries operator get fun call (enum entries are cached but not its get function call)
FontEdging.entries[_nGetFontEdging(_ptr)] -> FontEdging.(_nGetFontEdging(_ptr))

m-sasha:
I'm not sure a one-time allocation of a few objects is worth the loss of expressiveness of enums.

AFAIK, the reason Compose avoids enums (in public APIs) is that they're impossible to extend (add new values) without breaking compatibility. Not because of the overhead of object allocation.

ApoloApps:
> I'm not sure a one-time allocation of a few objects is worth the loss of expressiveness of enums.
> 
> AFAIK, the reason Compose avoids enums (in public APIs) is that they're impossible to extend (add new values) without breaking compatibility. Not because of the overhead of object allocation.

this `expressiveness of enums` is only used inside Skiko (no public repos use this expressiveness outside Skiko, in this case exhaustiveness in `when` expressions, which is the main advantage of enums over value classes). In Compose, this `when` check functions also suffer from this "lack of expressiveness" in the first place and they could be avoided as stated here:
```kotlin
    private fun StrokeCap.toSkia() = when (this) {
        StrokeCap.Butt -> SkPaintStrokeCap.BUTT
        StrokeCap.Round -> SkPaintStrokeCap.ROUND
        StrokeCap.Square -> SkPaintStrokeCap.SQUARE
        else -> SkPaintStrokeCap.BUTT
    }
//Change to
  private inline fun StrokeCap.toSkia() = SkPaintStrokeCap(SkPaintStrokeCap.value)

//Which would be eliminated/inlined when compiling because it is pointing to same value/class

//(the ordinal or value backing value class param)
``` 
I still think this is a good improvement, we really don't lose that much if at all. Still, yours and your team's call!! :)

m-sasha:
> In Compose, this when check functions also suffer from this "lack of expressiveness" in the first place and they could be avoided as stated here:

This can be solved (and is already solved in most places, I think) by matching the enum ordinal to the skia value. For example:
```
    var strokeCap: PaintStrokeCap
        get() = try {
            Stats.onNativeCall()
            PaintStrokeCap.entries[_nGetStrokeCap(_ptr)]
        } finally {
            reachabilityBarrier(this)
        }
        set(value) = try {
            Stats.onNativeCall()
            _nSetStrokeCap(_ptr, value.ordinal)
        } finally {
            reachabilityBarrier(this)
        }
```

m-sasha:
> I still think this is a good improvement, we really don't lose that much if at all. Still, yours and your team's call!! :)

I can't speak for the entire team, but my opinion on this is that the cost/benefit trade-off is not worth it.

ApoloApps:
>This can be solved (and is already solved in most places, I think) by matching the enum ordinal to the skia value. For example:

not exactly this usecase you are mentioning :) (which the one you are mentioning is actually this one specified in **Edit 2:**)
>in the first place and they could be avoided as stated here:

It is the one code block after this previous quote

ApoloApps:
>I'm not sure a one-time allocation of a few object

still, you are still invoking them frequently (even if singleton objects) which must not be as cheap as just reusing Compose's value class underlying value (no invocation of Integer cause it is reusing the one in Value Class of Compose)

ApoloApps:
https://android.googlesource.com/platform/frameworks/support/+/refs/changes/56/4095556/5/compose/docs/compose_collection_selection_mode_design.md#:~:text=Enums%20introduce%20class%2Dloading%20and%20memory%20overhead

PR #1194, opened 2026-04-24
Fix PathSegment.hashCode() not handling PathVerb.CLOSE

Bogdanmitrovic:
PathSegment.hashCode() threw RuntimeException("Unreachable") for PathVerb.CLOSE segments when used in hash-based collections such as intersect(). Added CLOSE case consistent with what equals() checks for CLOSE segments. Also removed the else branch since all PathVerb values are now handled. Added a test which fails unless CLOSE is handled in hashCode().

## Testing
Ran test on awt locally — the new test `closedPathIntersectTest` fails 
before the fix and passes after.

## Release Notes
Fix crash when using closed paths in hash-based collections such as `intersect()`

Fixes: [SKIKO-1038](https://youtrack.jetbrains.com/issue/SKIKO-1038)


MatkovIvan (approved):
Thanks!

PR #1200, opened 2026-05-26
Extract skiko test utils into separate module

hub-bla:
This PR introduces the extraction of test utilities into a separate module.

This is a required preparation step for extracting `Skottie` into a separate module, as it uses some of these test utilities.

Fixes [SKIKO-1136](https://youtrack.jetbrains.com/issue/SKIKO-1136)


skiko/src/commonMain/kotlin/org/jetbrains/skia/impl/Native.kt
```diff
@@ -1,10 +1,12 @@
 package org.jetbrains.skia.impl
 
 import org.jetbrains.skia.ManagedString
+import org.jetbrains.skiko.InternalSkikoApi
 
 expect class NativePointer
 
-internal expect class InteropPointer
+@OptIn(InternalSkikoApi::class)
```
MatkovIvan: `@OptIn` is for usage, not declaration

```suggestion
@InternalSkikoApi
```

hub-bla: I guess I'm too used to using `@OptIn` without noticing 😅 

It's fixed now. I’m opting in the entire `core` so it’s easier to review what actually needs to be exposed. In the end, I'll opt in only what's necessary individually

MatkovIvan (approved):

igordmn (approved):

hub-bla:
Tested [here](https://github.com/hub-bla/skiko/actions/runs/26499563967)  because yesterday there were issues with the workflows.

PR #1201, opened 2026-05-27
Implement support for Skia's Vulkan backend

ohlunasa:
Associated issue: [SKIKO-1138](https://youtrack.jetbrains.com/issue/SKIKO-1138/Add-Vulkan-DirectContext-support)

This PR adds:

- The ability to create a Vulkan `DirectContext`
- Support for passing a custom allocator to the Vulkan `DirectContext` (tested on JVM only)

I've verified the implementation works correctly, testing code is available [here](https://github.com/LunasaVT/skiko-vk-test)

**Note:** The Skia pack URL has been changed to point to my fork, which builds Skia with Vulkan enabled. A corresponding PR to upstream (JetBrains/skia) is open -- https://github.com/JetBrains/skia/pull/19. Once that PR is merged, the pack URL in this PR should be updated back to the upstream source.


skiko/buildSrc/src/main/kotlin/SkikoProjectContext.kt
```diff
@@ -102,7 +102,7 @@ fun SkikoProjectContext.declareSkiaTasks() {
 
             val skiaReleaseTag = project.skiaVersion(target)
 
-            val skiaBaseUrl = "https://github.com/JetBrains/skia/releases/download/$skiaReleaseTag"
+            val skiaBaseUrl = "https://github.com/LunasaVT/skia/releases/download/$skiaReleaseTag"
```
hub-bla, line 105: Reminder, revert before merging

skiko/src/jvmMain/cpp/common/DirectContext.cc
```diff
@@ -50,6 +50,437 @@ extern "C" JNIEXPORT jlong JNICALL Java_org_jetbrains_skia_DirectContextKt__1nMa
...
+#ifdef SK_VULKAN
+#include "include/gpu/ganesh/vk/GrVkDirectContext.h"
+#include "include/gpu/ganesh/vk/GrVkTypes.h"
+#include "include/gpu/vk/VulkanBackendContext.h"
+#include "include/gpu/vk/VulkanExtensions.h"
+#include "include/gpu/vk/VulkanMemoryAllocator.h"
+
+// simple built-in per-allocation allocator
+// this is always expected to be functional
+class SimpleVkAllocator : public skgpu::VulkanMemoryAllocator {
```
hub-bla: It looks like this class can be moved to `commonMain` cpp folder as it's declaration is in both jvm and native bindings

skiko/src/jvmMain/cpp/common/DirectContext.cc
```diff
@@ -50,6 +50,437 @@ extern "C" JNIEXPORT jlong JNICALL Java_org_jetbrains_skia_DirectContextKt__1nMa
...
+        if (  props & VK_MEMORY_PROPERTY_HOST_VISIBLE_BIT)     flags |= skgpu::VulkanAlloc::kMappable_Flag;
+        if (  props & VK_MEMORY_PROPERTY_LAZILY_ALLOCATED_BIT) flags |= skgpu::VulkanAlloc::kLazilyAllocated_Flag;
+
+        *out = reinterpret_cast<skgpu::VulkanBackendMemory>(new Alloc{mem, reqs.size, flags});
+        return VK_SUCCESS;
+    }
+};
+
+// calls back to the jvm
+class JvmVkAllocator : public skgpu::VulkanMemoryAllocator {
```
hub-bla, line 62: Could you tell a bit more what's the intended use case behind this?

It looks like a wrapper that exposes creation of VkAllocator to Kotlin, but I'm not sure what a subclass of `VulkanMemoryAllocator` on the Kotlin side would actually contain in the override of abstract methods.

The abstract methods only take handles and there are no Vulkan APIs on the skiko Kotlin side to work with. Is it expected to be paired with an external dependency, or am I missing something?

skiko/gradle.properties
```diff
@@ -7,7 +7,7 @@ kotlin.mpp.enableCInteropCommonization=true
 deploy.version=0.0.0
 
 # a tag from https://github.com/JetBrains/skia/releases
-dependencies.skia=m148-8967a2e80c
+dependencies.skia=m149-397e8b2ac7
```
hub-bla, line 10: reminder, change it after merging https://github.com/JetBrains/skia/pull/19

hub-bla (changes requested):
Thanks for the contribution :)

I have left a few points that need to be addressed/answered to move forward

skiko/src/jvmMain/cpp/common/DirectContext.cc
ohlunasa: The intended use case is to allow integrators to connect Skiko with their own memory allocation systems rather than relying on a built-in allocator implementation.

I've put together a simple example allocator here:
https://github.com/LunasaVT/skiko-vk-test/blob/master/src/main/kotlin/block-allocator.kt

In practice, this API is expected to be used alongside an external Vulkan binding library such as LWJGL (which the example uses). The abstract methods expose Vulkan handles so that implementations can perform allocation and memory-management operations through the Vulkan APIs provided by those external dependencies.

skiko/src/jvmMain/cpp/common/DirectContext.cc
ohlunasa: Good point. Since there is a common source set and the declaration is exposed by both the JVM and Native bindings, this implementation could probably live there instead.

I don't think there's anything platform-specific in `SimpleVkAllocator` itself, so moving it to the common C++ sources would avoid duplicating it across targets.

Will do and get back to you.

hub-bla: I noticed that Skia already comes with a predefined Vulkan memory allocator.
See: https://github.com/JetBrains/skia/blob/7861df630d7a46a49c1f90cc80f7972a2b3b4e7d/src/gpu/vk/vulkanmemoryallocator/VulkanAMDMemoryAllocator.cpp#L276

Maybe it's worth using this one instead of creating a naive implementation in Skiko. WDYT?

skiko/src/jvmMain/cpp/common/DirectContext.cc
hub-bla: In that case, can we annotate it with `ExperimentalSkikoApi` and document that it's intended for use with external libraries? I'm not sure we'll have Vulkan memory allocator bindings in Skiko, at least for now.

skiko/src/jvmMain/cpp/common/DirectContext.cc
ohlunasa: Thanks for pointing this out. The original simple allocator was only intended as a minimal built-in implementation to ensure Vulkan contexts could be created without requiring users to provide a custom allocator.

I agree it would make more sense to use Skia's allocator as the default implementation. It already provides the required features and is also maintained alongside Skia itself.

The custom allocator API would still be useful for applications that already have their own Vulkan memory management and want Skia to integrate with it, but I think using Skia's allocator as the built-in/default path is the better approach.

 I'll investigate replacing it and update the PR accordingly.

skiko/src/jvmMain/cpp/common/DirectContext.cc
ohlunasa: That sounds reasonable to me.

The custom allocator API is primarily intended for advanced integrations that already use a Vulkan binding library (e.g. LWJGL) and have their own memory management strategy. It's not particularly useful on its own without access to Vulkan APIs from another library.

I'll mark the allocator API as `@ExperimentalSkikoApi`.

skiko/src/commonMain/cpp/common/include/VulkanMemoryAllocators.hh
```diff
@@ -0,0 +1,22 @@
...
+
+#ifdef SK_VULKAN
+
+#include "include/core/SkRefCnt.h"
+#include "include/gpu/vk/VulkanBackendContext.h"
+#include "include/gpu/vk/VulkanMemoryAllocator.h"
+
+namespace skgpu {
+
+enum class ThreadSafe : bool {
```
MatkovIvan, line 11: Do we really need an enum for it?

skiko/src/commonMain/kotlin/org/jetbrains/skia/VulkanMemoryAllocator.kt
```diff
@@ -0,0 +1,181 @@
...
+ */
+@ExperimentalSkikoApi
+abstract class VulkanMemoryAllocator {
+
+    /**
+     * Describes how Skia intends to use a buffer allocation.
+     *
+     * This may or may not be acknowledged by the underlying allocator implementation.
+     */
+    enum class BufferUsage(internal val ordinalValue: Int) {
```
MatkovIvan: Prefer `@JvmInline value class`

skiko/src/commonMain/kotlin/org/jetbrains/skia/VulkanMemoryAllocator.kt
```diff
@@ -0,0 +1,181 @@
...
+     * The result of a successful allocation.
+     *
+     * @property deviceMemory Vulkan `VkDeviceMemory` handle.
+     * @property offset Byte offset within `deviceMemory`.
+     *                  For dedicated allocations this is typically `0`.
+     * @property size Allocation size in bytes.
+     * @property memoryTypeIndex Index into
+     * `VkPhysicalDeviceMemoryProperties.memoryTypes`.
+     */
+    data class Allocation(
```
MatkovIvan: [Avoid using data classes in your API﻿](https://kotlinlang.org/docs/api-guidelines-backward-compatibility.html#avoid-using-data-classes-in-your-api)

skiko/src/commonMain/kotlin/org/jetbrains/skia/VulkanMemoryAllocator.kt
```diff
@@ -0,0 +1,181 @@
...
+        val usage = BufferUsage.entries.getOrNull(usageOrdinal) ?: return null
+        return allocateBufferMemory(buffer, usage, allocationPropertyFlags)
+    }
+
+    companion object {
+
+        /**
+         * Requests a dedicated `VkDeviceMemory` allocation.
+         */
+        const val ALLOCATION_FLAG_DEDICATED: Int = 0b0001
```
MatkovIvan: Should we group it into another value class to make it type-safe?

skiko/src/jvmTest/kotlin/org/jetbrains/skia/VulkanInteropTest.jvm.kt
```diff
@@ -0,0 +1,49 @@
...
+import org.jetbrains.skiko.ExperimentalSkikoApi
+import org.jetbrains.skiko.Library
+import kotlin.test.Test
+import kotlin.test.assertFailsWith
+
+@OptIn(ExperimentalSkikoApi::class)
+class VulkanInteropTest {
+    private companion object {
+        init {
+            Library.load()
```
MatkovIvan: It should be in production code, not in tests. It highlights the issue

skiko/src/commonMain/cpp/common/include/VulkanMemoryAllocators.hh
ohlunasa: This mirrors skia's own definition

skiko/src/commonMain/kotlin/org/jetbrains/skia/VulkanMemoryAllocator.kt
ohlunasa: I've converted it into an inlined value class.

skiko/src/commonMain/kotlin/org/jetbrains/skia/VulkanMemoryAllocator.kt
ohlunasa: Good idea, this makes the API much safer. I've already pushed the change.

skiko/src/commonMain/kotlin/org/jetbrains/skia/VulkanMemoryAllocator.kt
ohlunasa: I've converted `Allocation` into a regular class.

Wyvest:
Any updates on this?

Wyvest:
Fyi, Lunasa (the original maintainer of the fork) is busy, so I am now maintaining the fork / this PR for them in their absence. I just merged upstream changes + updated to use upstream skia as their PR was merged there.

hub-bla (approved):

hub-bla:
Approved


In `skiko-graphite` please switch 
```
#if defined(SK_VULKAN)
``` 
 to 

```
#if defined(SK_VULKAN) && (defined(_WIN32) || defined(__linux__))
```
 It should resolve CI failures

PR #1203, opened 2026-05-28
Make getCurrentSystemTheme on Linux thread-safe

m-sasha:
1. Makes initialization thread-safe using [C++ magic statics](https://blog.mbedded.ninja/programming/languages/c-plus-plus/magic-statics/)
2. Makes dbus use thread-safe by calling [`dbus_threads_init_default()`](https://dbus.freedesktop.org/doc/api/html/group__DBusThreads.html#ga33b6cf3b8f1e41bad5508f84758818a7)

Created with:
Claude Code Sonnet 4.6
  Total cost:            $1.20
  Total duration (API):  7m 14s
  Total duration (wall): 21m 14s
  Total code changes:    36 lines added, 86 lines removed

### DIFF
--- a/skiko/src/awtMain/cpp/linux/theme.cc
+++ b/skiko/src/awtMain/cpp/linux/theme.cc
@@ -5,33 +5,28 @@
 
 #include "jni_helpers.h"
 
+static bool initDbusThreads(void* lib) {
+    auto func = (dbus_bool_t(*)()) dlsym(lib, "dbus_threads_init_default");
+    if (func) func();
+    return true;
+}
+
 static void* loadLibDbus() {
-    static void* result = nullptr;
-    if (result != nullptr) return result;
-    result = dlopen("libdbus-1.so", RTLD_LAZY | RTLD_LOCAL);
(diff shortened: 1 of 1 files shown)


skiko/src/awtMain/cpp/linux/theme.cc
```diff
@@ -5,33 +5,28 @@
...
+    if (func) func();
+    return true;
+}
+
 static void* loadLibDbus() {
-    static void* result = nullptr;
-    if (result != nullptr) return result;
-    result = dlopen("libdbus-1.so", RTLD_LAZY | RTLD_LOCAL);
+    static void* result = dlopen("libdbus-1.so", RTLD_LAZY | RTLD_LOCAL);
+    static bool _ = result && initDbusThreads(result);
```
MatkovIvan: Should we handle return status of `dbus_threads_init_default` call?

MatkovIvan (approved):

skiko/src/awtMain/cpp/linux/theme.cc
m-sasha: Sure

PR #1205, opened 2026-05-31
Transform Matrix* to value classes

ApoloApps:
It transforms Matrix* classes to be value classes so that we avoid double allocation in these simple classes (1 fo FloatArray and 1 for the Matrix class itself)
https://github.com/JetBrains/skiko/pull/1137 followup (at least partly)

Fixes [SKIKO-1140](https://youtrack.jetbrains.com/issue/SKIKO-1140/Change-Matrix-to-value-classes)

**Next**
Follow-up PR in Compose Core to make use of the internal float array constructor to reuse if coming from another FLoatArray

### DIFF
--- a/skiko/src/commonMain/kotlin/org/jetbrains/skia/Matrix33.kt
+++ b/skiko/src/commonMain/kotlin/org/jetbrains/skia/Matrix33.kt
@@ -1,11 +1,14 @@
 package org.jetbrains.skia
 
+import kotlin.jvm.JvmInline
 import kotlin.math.PI
 import kotlin.math.abs
 import kotlin.math.cos
 import kotlin.math.sin
 
-internal fun Float.toRadians(): Double = this.toDouble() / 180 * PI
+internal const val radiansToDegrees = 180.0 / PI
+internal const val tolerance = (1.0f / (1 shl 12)).toDouble()
+internal inline fun Float.toRadians(): Double = this.toDouble() * radiansToDegrees
 
 /**
  *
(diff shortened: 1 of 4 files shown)


skiko/src/commonMain/kotlin/org/jetbrains/skia/Matrix33.kt
```diff
@@ -1,11 +1,14 @@
...
 
+import kotlin.jvm.JvmInline
 import kotlin.math.PI
 import kotlin.math.abs
 import kotlin.math.cos
 import kotlin.math.sin
 
-internal fun Float.toRadians(): Double = this.toDouble() / 180 * PI
+internal const val radiansToDegrees = 180.0 / PI
+internal const val tolerance = (1.0f / (1 shl 12)).toDouble()
```
MatkovIvan:
```suggestion
private const val radiansToDegrees = 180.0 / PI
private const val tolerance = (1.0f / (1 shl 12)).toDouble()
```

skiko/src/commonMain/kotlin/org/jetbrains/skia/Matrix44.kt
```diff
@@ -11,14 +12,31 @@ import org.jetbrains.skia.impl.withResult
  * Skia assumes a right-handed coordinate system:
  * +X goes to the right
  * +Y goes down
- * +Z goes into the screen (away from the viewer)
+ * +Z goes into the screen (away from the viewer)ç
```
MatkovIvan:
```suggestion
 * +Z goes into the screen (away from the viewer)
```

MatkovIvan:
Thanks, LGTM in general, just a few nitpicks

skiko/src/commonMain/kotlin/org/jetbrains/skia/Matrix33.kt
ApoloApps: done

skiko/src/commonMain/kotlin/org/jetbrains/skia/Matrix44.kt
ApoloApps: done

MatkovIvan (approved):

SergeVL:
const val radiansToDegrees = 180.0 / PI
 - inline fun Float.toRadians(): Double = this.toDouble() * radiansToDegrees
 + inline fun Float.toRadians(): Double = this.toDouble() / radiansToDegrees

ApoloApps:
> const val radiansToDegrees = 180.0 / PI
> 
>  - inline fun Float.toRadians(): Double = this.toDouble() * radiansToDegrees
> 
>  + inline fun Float.toRadians(): Double = this.toDouble() / radiansToDegrees

yeah, the code is wrong. I must have confused with Kotlin's operation precedence somehow when I changed the code. 
To get radians using radiansToDegrees, degrees must be divided by radiansToDegrees to get the inverse, which is what our operation really is: degrees to radians. I'll be submitting a PR with a fix, thanks for catching it!!

PR: https://github.com/JetBrains/skiko/pull/1269

PR #1206, opened 2026-06-02
GitHub checks. Checking that AWT runtimes can be built

igordmn:
Previously we didn't check JVM Linux x64/arm64, macOs x64, Windows arm64 AWT runtimes

### DIFF
--- a/.github/workflows/publish-dry-run.yml
+++ b/.github/workflows/publish-dry-run.yml
@@ -137,7 +137,7 @@ jobs:
             -Pskiko.arch=arm64 \
             :skiko:publishToMavenLocal
 
-      # Build Arm64 separately (arm64 is built in publishToMavenLocal)
+      # Build x64 separately (arm64 is built in publishToMavenLocal)
       - shell: bash
         name: 'Publish AWT Runtime x64 Maven Local'
         run: |
--- a/.github/workflows/publish-dry-run.yml
+++ b/.github/workflows/publish-dry-run.yml
@@ -62,26 +62,15 @@ jobs:
         name: 'Setup Prerequisites'
 
       - uses: ./.github/actions/docker-skiko-run
-        name: 'Publish KMP and AWT Maven Local'
+        name: 'Publish Maven Local x64'
         with:
           image_name: linux-compat
           command: |
             ./gradlew --no-daemon --stacktrace \
               -Pskiko.native.enabled=true \
               -Pskiko.awt.enabled=true \
-              :skiko:publishKotlinMultiplatformPublicationToMavenLocal \
-              :skiko:publishAwtPublicationToMavenLocal \
-              :skiko:publishAwtRuntimeElementsPublicationToMavenLocal
-
-      - uses: ./.github/actions/docker-skiko-run
-        name:  'Publish AWT Runtime x64 Maven Local'
-        with:
-          image_name: linux-compat
-          command: |
-            ./gradlew --no-daemon --stacktrace \
-              -Pskiko.awt.enabled=true \
-              :skiko:publishSkikoJvmRuntimeLinuxX64PublicationToMavenLocal \
-              -Pskiko.arch=x64
+              -Pskiko.arch=x64 \
+               :skiko:publishToMavenLocal
 
       - uses: ./.github/actions/docker-skiko-run
         name: 'Publish Native Maven Local'
@@ -112,6 +101,7 @@ jobs:
       - uses: ./.github/actions/setup-prerequisites
         name: 'Setup Prerequisites'
 
+      # We run it on a separate runner because cross-compilation on x64 fails
       - uses: ./.github/actions/docker-skiko-run
         name:  'Publish AWT Runtime Arm64 Maven Local'
         with:
@@ -139,12 +129,15 @@ jobs:
         name: 'Setup Prerequisites'
 
       - shell: bash
-        name: 'Publish Maven Local'
+        name: 'Publish Maven Local arm64'
         run: |
           ./gradlew --no-daemon --stacktrace \
             -Pskiko.native.enabled=true \
+            -Pskiko.awt.enabled=true \
+            -Pskiko.arch=arm64 \
             :skiko:publishToMavenLocal
 
+      # Build Arm64 separately (arm64 is built in publishToMavenLocal)
       - shell: bash
         name: 'Publish AWT Runtime x64 Maven Local'
         run: |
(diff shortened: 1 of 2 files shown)


.github/workflows/publish-dry-run.yml
igordmn: The alternative to the current checks is to include it here, with `skiko.awt.enabled=true`. But I decided to separate it, to mimic the build server, and because we need to build 2 architectures

igordmn:
Sorry, it looks like publishToMavenLocal should publish some part of it, need to recheck

.github/workflows/publish-dry-run.yml
igordmn: Outdated

igordmn:
New times

Windows
image

macOs
image

Linux x64
image

Linux arm64
image

.github/workflows/publish-dry-run.yml
```diff
@@ -62,16 +62,15 @@ jobs:
...
         with:
           image_name: linux-compat
           command: |
             ./gradlew --no-daemon --stacktrace \
               -Pskiko.native.enabled=true \
               -Pskiko.awt.enabled=true \
-              :skiko:publishKotlinMultiplatformPublicationToMavenLocal \
-              :skiko:publishAwtPublicationToMavenLocal \
-              :skiko:publishAwtRuntimeElementsPublicationToMavenLocal
+              -Pskiko.arch=x64 \
```
igordmn, line 72: Example of dry-run of publishToMavenLocal, filtering publish tasks:
```
gradlew -Pskiko.awt.enabled=true publishToMavenLocal --dry-run
:publishAwtPublicationToMavenLocal SKIPPED
:publishAwtRuntimeElementsPublicationToMavenLocal SKIPPED
:publishKotlinMultiplatformPublicationToMavenLocal SKIPPED
:publishSkikoJvmRuntimeAngleWindowsX64PublicationToMavenLocal SKIPPED
:publishSkikoJvmRuntimeWindowsX64PublicationToMavenLocal SKIPPED
```

.github/workflows/publish-dry-run.yml
```diff
@@ -92,6 +91,27 @@ jobs:
...
+    name: 'Linux arm64 Publish Dry Run'
+    runs-on: ubuntu-24.04-arm
+    steps:
+      - uses: actions/checkout@v4
+        name: 'Check out code'
+
+      - uses: ./.github/actions/setup-prerequisites
+        name: 'Setup Prerequisites'
+
+      # We run it on a separate runner because cross-compilation on x64 fails
```
hub-bla, line 104: This is because [linux-compat ](https://github.com/JetBrains/skiko/blob/master/skiko/docker/linux-compat/Dockerfile )does not download arm versions of some libraries like [linux-amd64](https://github.com/JetBrains/skiko/blob/master/skiko/docker/linux-amd64/Dockerfile#L20-L32) does

I'm fine with separate job though

hub-bla (approved):

.github/workflows/publish-dry-run.yml
igordmn: @MatkovIvan does it make sense to install them for linux-compat too?

MatkovIvan: Yes, we already have some cross-compilation setup there. If it misses something it's better to fix it

igordmn: I didn't succeed with few attempts to enable cross-compilation on linux-compact, so, let's just run it as a separate job.

MatkovIvan: Do we do the same on TeamCity?

.github/workflows/publish-dry-run.yml
```diff
@@ -92,6 +91,27 @@ jobs:
...
+      - uses: ./.github/actions/docker-skiko-run
+        name:  'Publish AWT Runtime arm64 Maven Local'
+        with:
+          image_name: linux-compat
+          command: |
+            ./gradlew --no-daemon --stacktrace \
+              -Pskiko.awt.enabled=true \
+              :skiko:publishSkikoJvmRuntimeLinuxArm64PublicationToMavenLocal \
+              -Pskiko.arch=arm64
```
MatkovIvan: Nitpick alignement (all flags before tasks)
```suggestion
              -Pskiko.arch=arm64 \
              :skiko:publishSkikoJvmRuntimeLinuxArm64PublicationToMavenLocal
```

.github/workflows/publish-dry-run.yml
```diff
@@ -109,12 +129,23 @@ jobs:
...
             :skiko:publishToMavenLocal
 
+      # Build x64 separately (arm64 is built in publishToMavenLocal)
+      - shell: bash
+        name: 'Publish AWT Runtime x64 Maven Local'
+        run: |
+          ./gradlew --no-daemon --stacktrace \
+            -Pskiko.awt.enabled=true \
+            :skiko:publishSkikoJvmRuntimeMacosX64PublicationToMavenLocal \
+            -Pskiko.arch=x64
```
MatkovIvan:
```suggestion
            -Pskiko.arch=x64 \
            :skiko:publishSkikoJvmRuntimeMacosX64PublicationToMavenLocal
```

.github/workflows/publish-dry-run.yml
```diff
@@ -142,12 +173,23 @@ jobs:
...
              :skiko:publishToMavenLocal
 
+      # Build Arm64 separately (x64 is built in publishToMavenLocal)
+      - shell: bash
+        name: 'Publish AWT Runtime arm64 Maven Local'
+        run: |
+          ./gradlew --no-daemon --stacktrace \
+            -Pskiko.awt.enabled=true \
+            :skiko:publishSkikoJvmRuntimeWindowsArm64PublicationToMavenLocal \
+            -Pskiko.arch=arm64
```
MatkovIvan:
```suggestion
            -Pskiko.arch=arm64 \
            :skiko:publishSkikoJvmRuntimeWindowsArm64PublicationToMavenLocal
```

MatkovIvan (approved):

.github/workflows/publish-dry-run.yml
igordmn: yes, Linux builds on x86+Arm, macOs on Arm only, Windows on x86 only

hub-bla: linux-amd64 is a different distro so it's not possible to map the same command for getting these libs. The easiest way would be to get them through multi stage build where we add something like this

```
FROM --platform=linux/arm64 amazonlinux:2 AS arm64-libs
RUN yum install -y \
    fontconfig fontconfig-devel \
    freetype freetype-devel \
    mesa-libGL mesa-libGL-devel \
    mesa-libGLU mesa-libGLU-devel \
    libX11 libX11-devel \
    libXrandr libXrandr-devel \
    dbus-libs dbus-devel && \
    yum clean all
```

and then later
```
FROM amazonlinux:2

COPY --from=arm64-libs /usr/lib64/ $ARM_TOOLCHAIN_SYSROOT/usr/lib64/
COPY --from=arm64-libs /usr/lib/ $ARM_TOOLCHAIN_SYSROOT/usr/lib/
```

I can make a PR to fix it but TBH since the JVM target can be compiled on Linux ARM64 and doesn't require cross compilation we probably shouldn't cross compile it?

igordmn: It looks like it doesn't worth it because we have Arm64 agents on GitHub/TeamCity.

PR #1208, opened 2026-06-05
Control Skia symbols visibility across platforms

hub-bla:
To extract Skottie and other extensions from Skiko core, we need to expose certain private symbols from skia. This requires creating tasks to control the visibility of Skia symbols.

To control visibility of symbols the following things were applied:

### JVM
We are generating a lists of symbols that we want to export. To obtain the symbols we're doing the following:

1. Collecting all globally defined symbols from the core Skiko/Skia inputs: Skia static libraries plus core object files.
2. Collecting all undefined symbols from extension/module inputs. These represent symbols that extension modules need the core library to provide.
3. Adding JNI infrastructure symbols from the core exports, such as `Java_`, `JNI`, and `jvm`, so the JVM can still find the native entry points.
4. Intersecting the extension imports with the core exports. The result is the list of core symbols that should stay exported.

Generated symbol list it then passed to the linker which hide symbols we don't need to have visible.

### Native 
For Native targets, Skia is linked into the final produced binary rather than exposed as a shared "core" library that extension modules link against. Because of that, Skia symbols do not need to remain externally visible for other Skiko modules to resolve against them.
Since we are going to build Skia from our fork with Skia symbols public by default we need to hide them on native.

### WASM
This PR does not include wasm parser as we don't need it. We can first create the wasm binary for extension functions and then provide them when compiling the core. This will allow to export only symbols that are needed by the extension modules automatically. This will be introduced in the PR with Skottie extraction.

Closes [SKIKO-1142](https://youtrack.jetbrains.com/issue/SKIKO-1142)

### DIFF
--- a/skiko/build.gradle.kts
+++ b/skiko/build.gradle.kts
@@ -479,6 +479,39 @@ afterEvaluate {
     }
 }
 
+fun configureSymbolsFor(os: OS, arch: Arch) {
+    val suffix = joinToTitleCamelCase(os.id, arch.id)
+    val skiaBindingsDir = skikoProjectContext.registerOrGetSkiaDirProvider(os, arch)
+    val coreCompile = tasks.named<CompileSkikoCppTask>("compileJvmBindings$suffix")
+    val coreObjcCompile = if (os.isMacOs) tasks.named<CompileSkikoObjCTask>("objcCompile$suffix") else null
+
+    skikoProjectContext.configureGenerateSymbolsList(
+        os, arch, skiaBindingsDir, coreCompile, coreObjcCompile
+    )
+
+    tasks.named("linkJvmBindings$suffix") {
+        dependsOn("generateSymbolsList$suffix")
+    }
+}
+
+if (supportAwt) {
+    afterEvaluate {
+        configureSymbolsFor(targetOs, targetArch)
+
+        if (targetOs == OS.MacOS && targetArch == Arch.Arm64) {
+            configureSymbolsFor(OS.MacOS, Arch.X64)
+        }
+    }
+}
+
+if (supportAndroid) {
+    afterEvaluate {
+        for (arch in arrayOf(Arch.X64, Arch.Arm64)) {
+            configureSymbolsFor(OS.Android, arch)
+        }
+    }
+}
+
 skikoProjectContext.declarePublications()
 
 val mavenCentral = MavenCentralProperties(project)
--- a/skiko/buildSrc/src/main/kotlin/symbols/SymbolExtractor.kt
+++ b/skiko/buildSrc/src/main/kotlin/symbols/SymbolExtractor.kt
@@ -8,7 +8,7 @@ import java.io.File
 internal class SymbolExtractor(
     private val execOperations: ExecOperations,
     private val os: OS,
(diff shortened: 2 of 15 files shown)


MatkovIvan (approved):

skiko/build.gradle.kts
```diff
@@ -479,6 +479,39 @@ afterEvaluate {
...
+        dependsOn("generateSymbolsList$suffix")
+    }
+}
+
+if (supportAwt) {
+    afterEvaluate {
+        configureSymbolsFor(targetOs, targetArch)
+
+        if (targetOs == OS.MacOS && targetArch == Arch.Arm64) {
+            configureSymbolsFor(OS.MacOS, Arch.X64)
```
igordmn, line 502: Why do we configure symbols for the other architecture? (the same for Android)

When we publish, we call the publishing tasks separately passing arm64 in one task, and x64 in the other.

skiko/buildSrc/src/main/kotlin/symbols/SymbolExtractor.kt
```diff
@@ -0,0 +1,82 @@
...
+            OS.Linux, OS.Android, OS.MacOS, OS.IOS, OS.TVOS -> {
+                val nmCommand = nmCommand(type)
+                val output = run(nmCommand.executable, nmCommand.args + actualFiles.map { it.absolutePath })
+                parseNmPosix(output)
+                    .filter { it.type == type }
+                    .mapTo(mutableSetOf()) { it.name }
+            }
+
+            OS.Windows -> {
+                val output = run("dumpbin", listOf("/SYMBOLS") + actualFiles.map { it.absolutePath })
```
igordmn: This is not in the PATH on my machine.

Please, use `windowsSdkPaths.kt` to define the path to this tool, relative to Visual Studio Tools

igordmn:
Please, run https://buildserver.labs.intellij.net/buildConfiguration/Skiko_PublishSnapshot for checking the TeamCity publication

igordmn:
> Please, run https://buildserver.labs.intellij.net/buildConfiguration/Skiko_PublishSnapshot for checking the TeamCity publication

It actually may fail because it may need a new docker 🤔. Then I suggest to be ready to fix/revert after merging

skiko/build.gradle.kts
hub-bla: For macOS because of https://github.com/JetBrains/skiko/blob/6bcc77c4caab15a66039674e148b354d9feef050/skiko/buildSrc/src/main/kotlin/tasks/configuration/JvmTasksConfiguration.kt#L434

For Android because of https://github.com/JetBrains/skiko/blob/6bcc77c4caab15a66039674e148b354d9feef050/skiko/build.gradle.kts#L424

I didn't add it to `createSkikoJvmJarTask` because only the core module will configure symbols and extension module will reuse that

skiko/buildSrc/src/main/kotlin/symbols/SymbolExtractor.kt
hub-bla: Should be working now

igordmn (approved):

PR #1209, opened 2026-06-05
Convert non-enum classes to value classes to avoid allocations

ApoloApps:
Fixes [SKIKO-1143](https://youtrack.jetbrains.com/issue/SKIKO-1143/Convert-class-FontStyle-interface-FontWeight-interface-FontWidth-and-interface-PathSegmentMask-to-value-classes)

Non-enum classes (interfaces or bare classes) can be optimized if they're just a wrapper of a primitive by using value classes.
Convert also FontMetrics, which in the boundary layer already uses FloatArray under the hood so transforming it into value class wrapping that same FloatArray, saves a few allocations (class allocation, since array is reused)

For more info about the approach of this PR, see issue (I explain what the changes are and the why of it)

### DIFF
--- a/skiko/src/commonMain/kotlin/org/jetbrains/skia/FontMetrics.kt
+++ b/skiko/src/commonMain/kotlin/org/jetbrains/skia/FontMetrics.kt
@@ -21,14 +21,27 @@ value class FontMetrics(internal val metrics: FloatArray) {
         xMax: Float,
         xHeight: Float,
         capHeight: Float,
-        underlineThickness: Float,
-        underlinePosition: Float,
-        strikeoutThickness: Float,
-        strikeoutPosition: Float
+        underlineThickness: Float?,
+        underlinePosition: Float?,
+        strikeoutThickness: Float?,
+        strikeoutPosition: Float?
     ) : this(
         floatArrayOf(
-            top, ascent, descent, bottom, leading, avgCharWidth, maxCharWidth, xMin, xMax, xHeight, capHeight,
-            underlineThickness, underlinePosition, strikeoutThickness, strikeoutPosition
+            top,
+            ascent,
+            descent,
+            bottom,
+            leading,
+            avgCharWidth,
+            maxCharWidth,
+            xMin,
+            xMax,
+            xHeight,
+            capHeight,
+            underlineThickness ?: Float.NaN,
+            underlinePosition ?: Float.NaN,
+            strikeoutThickness ?: Float.NaN,
+            strikeoutPosition ?: Float.NaN
         )
     )
 
--- a/skiko/src/commonMain/kotlin/org/jetbrains/skia/FontMetrics.kt
+++ b/skiko/src/commonMain/kotlin/org/jetbrains/skia/FontMetrics.kt
@@ -8,7 +8,25 @@ import kotlin.jvm.JvmInline
 @JvmInline
 value class FontMetrics(internal val metrics: FloatArray) {
 
-    //Maintained for binary compatibility
+    /**
+     * Maintained for binary compatibility and previous construction with its named values
+     *
+     * @param top greatest extent above origin of any glyph bounding box, typically negative; deprecated with variable fonts
+     * @param ascent distance to reserve above baseline, typically negative
+     * @param descent distance to reserve below baseline, typically positive
+     * @param bottom greatest extent below origin of any glyph bounding box, typically positive; deprecated with variable fonts
+     * @param leading distance to add between lines, typically positive or zero
+     * @param avgCharWidth average character width, zero if unknown
+     * @param maxCharWidth maximum character width, zero if unknown
+     * @param xMin greatest extent to left of origin of any glyph bounding box, typically negative; deprecated with variable fonts
+     * @param xMax greatest extent to right of origin of any glyph bounding box, typically positive; deprecated with variable fonts
+     * @param xHeight height of lower-case 'x', zero if unknown, typically negative
+     * @param capHeight height of an upper-case letter, zero if unknown, typically negative
+     * @param underlineThickness underline thickness
+     * @param underlinePosition distance from baseline to top of stroke, typically positive
+     * @param strikeoutThickness strikeout thickness
+     * @param strikeoutPosition distance from baseline to bottom of stroke, typically negative
+     */
     constructor(
         top: Float,
         ascent: Float,
@@ -54,7 +72,7 @@ value class FontMetrics(internal val metrics: FloatArray) {
      * distance to reserve above baseline, typically negative
      */
     val ascent: Float
-         get() = metrics[1]
+        get() = metrics[1]
     /**
      * distance to reserve below baseline, typically positive
      */
--- a/skiko/src/commonMain/kotlin/org/jetbrains/skia/FontMetrics.kt
+++ b/skiko/src/commonMain/kotlin/org/jetbrains/skia/FontMetrics.kt
@@ -3,157 +3,113 @@ package org.jetbrains.skia
 import org.jetbrains.skia.impl.InteropPointer
 import org.jetbrains.skia.impl.InteropScope
 import org.jetbrains.skia.impl.withResult
+import kotlin.jvm.JvmInline
 
-class FontMetrics(
+@JvmInline
+value class FontMetrics(internal val metrics: FloatArray) {
+
+    //Maintained for binary compatibility
+    constructor(
+        top: Float,
+        ascent: Float,
+        descent: Float,
+        bottom: Float,
+        leading: Float,
+        avgCharWidth: Float,
+        maxCharWidth: Float,
+        xMin: Float,
+        xMax: Float,
+        xHeight: Float,
+        capHeight: Float,
+        underlineThickness: Float,
+        underlinePosition: Float,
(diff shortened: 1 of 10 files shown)


ApoloApps:
@MatkovIvan

skiko/src/webMain/kotlin/org/jetbrains/skiko/Actuals.web.kt
```diff
@@ -1,12 +1,11 @@
 package org.jetbrains.skiko
 
-import kotlinx.browser.window
 import org.w3c.dom.HTMLElement
 
 internal actual inline fun <R> maybeSynchronized(lock: Any, block: () -> R): R =
     block()
 
-actual fun currentNanoTime(): Long = (window.performance.now() * 1_000_000).toLong()
+actual fun currentNanoTime(): Long = currentNanoTimeWindowPerformance().toLong()
```
ApoloApps, line 8: More performant due to https://youtrack.jetbrains.com/issue/KT-87142

skiko/src/commonMain/kotlin/org/jetbrains/skia/FontWeight.kt
```diff
@@ -1,17 +1,23 @@
 package org.jetbrains.skia
 
-interface FontWeight {
```
MatkovIvan, line 3: Since it changes the public API in a breaking way, we need to
- Prepare CMP adoption in advance
- Merge it to the next non-patch version (0.151.0)

skiko/src/webMain/kotlin/org/jetbrains/skiko/CanvasRenderer.kt
```diff
@@ -69,24 +85,26 @@ internal abstract class CanvasRenderer(
...
 
+@OptIn(ExperimentalWasmJsInterop::class)
+private fun windowRequestAnimationFrame(callback: JsAny) : Int =
+    //language=JavaScript
+    js("window.requestAnimationFrame(callback)")
+
+@OptIn(ExperimentalWasmJsInterop::class)
+private fun rAfCallbackToJs(callback: (Double) -> Unit): JsAny =
+    //language=JavaScript
+    js("(timestamp) => callback(timestamp)")
```
MatkovIvan: cc @Schahen @eymar

MatkovIvan (commented):
No blockers from my side except merging into a non-patch version

skiko/src/commonMain/kotlin/org/jetbrains/skia/FontMetrics.kt
```diff
@@ -3,158 +3,114 @@ package org.jetbrains.skia
...
+        avgCharWidth: Float,
+        maxCharWidth: Float,
+        xMin: Float,
+        xMax: Float,
+        xHeight: Float,
+        capHeight: Float,
+        underlineThickness: Float,
+        underlinePosition: Float,
+        strikeoutThickness: Float,
+        strikeoutPosition: Float
```
hub-bla: Previously these were nullable. Is it expected?

skiko/src/webMain/kotlin/org/jetbrains/skiko/CanvasRenderer.kt
ApoloApps: https://github.com/JetBrains/compose-multiplatform-core/pull/2928#discussion_r3457658548 explains this change better. Basically, instead of creating a new lambda object on each call of requestAnimationFrame, I first of all cache it in class. Now we only have 1 allocation instead. Then, there is `window.requestAnimationFrame(lambda)`. Problem with this is that since it is an external declaration, an external adapter for it is created and the lambda argument also creates another adapter because Js doesn't know how to handle Kotlin lambdas. This happens on each call of rAf. Instead, I cache the callback adapter (by adding the js block that returns a JsAny, which is the Js object we need for rAf callback), which is the js object i pass to rAf (we avoid creating the lambda and adapter of it on each call). Then there is the extraction of `window.requestAnimationFrame(lambda)` to a js block. This is due to https://youtrack.jetbrains.com/issue/KT-87142 which creates several calls back and forth in the interop border (first gets the js window object and then passes it to wasm which then uses it in the external rAf declaration, which crosses the border once again to Js). This way we avoid the interop border cost as all of this code compiles to a few calls in Wasm compared to previous impl

skiko/src/commonMain/kotlin/org/jetbrains/skia/FontMetrics.kt
ApoloApps: Yep, forgot to change them to nullable in the constructor. In the underlying FloatArray they are represented as Float.NaN though

ApoloApps: done

skiko/src/webMain/kotlin/org/jetbrains/skiko/CanvasRenderer.kt
ApoloApps: I'll be reverting to only caching the lambda passed to `window.requestAnimationFrame(lambda)` as the code I'm referring to adds little to no overhead (it is insignificant). I still keep the rAf lambda cached as a way to save on allocations on each rAf call
Edit: I'll keep the `private fun windowRequestAnimationFrame(callback: JsAny) : Int =
    //language=JavaScript
    js("window.requestAnimationFrame(callback)")` as the [youtrack.jetbrains.com/issue/KT-87142](https://youtrack.jetbrains.com/issue/KT-87142) issue is there

skiko/src/commonMain/kotlin/org/jetbrains/skia/FontWeight.kt
ApoloApps: Done. Follow up: https://github.com/JetBrains/compose-multiplatform-core/pull/3157

skiko/src/commonTest/kotlin/org/jetbrains/skia/StrutStyleTests.kt
```diff
@@ -12,11 +12,11 @@ class StrutStyleTests {
     @Test
     fun strutStyleTest() {
         StrutStyle().use { strutStyle ->
-            assertEquals(FontStyle(400, 5, FontSlant.UPRIGHT), strutStyle.fontStyle)
+            assertEquals(FontStyle(FontWeight(400), FontWidth(5), FontSlant.UPRIGHT), strutStyle.fontStyle)
 
-            strutStyle.fontStyle = FontStyle(300, 4, FontSlant.ITALIC)
+            strutStyle.fontStyle = FontStyle(FontWeight(300), FontWidth(4), FontSlant.ITALIC)
```
MatkovIvan: Should it be
```suggestion
            strutStyle.fontStyle = FontStyle(FontWeight.LIGHT, FontWidth.SEMI_CONDENSED, FontSlant.ITALIC)
```
?

skiko/src/commonMain/kotlin/org/jetbrains/skia/FontMetrics.kt
```diff
@@ -3,158 +3,127 @@ package org.jetbrains.skia
 import org.jetbrains.skia.impl.InteropPointer
 import org.jetbrains.skia.impl.InteropScope
 import org.jetbrains.skia.impl.withResult
+import kotlin.jvm.JvmInline
+
+@JvmInline
+value class FontMetrics(internal val metrics: FloatArray) {
+
+    //Maintained for binary compatibility
```
MatkovIvan: It seems it's not only for binary compatibility, but for construction it from named values in general.

skiko/src/commonMain/kotlin/org/jetbrains/skia/FontMetrics.kt
```diff
@@ -3,158 +3,127 @@ package org.jetbrains.skia
...
      */
-    val strikeoutThickness: Float?,
+    val strikeoutThickness: Float?
+        get() = metrics[13].asNumberOrNull()
     /**
      * distance from baseline to bottom of stroke, typically negative
      */
     val strikeoutPosition: Float?
-) {
-    /**
```
MatkovIvan, line 69: Restore as constructor parameters KDoc

MatkovIvan (approved):
LGTM in general, just some minor comments.

@hub-bla Please review the final state and handle merging/followups

skiko/src/commonMain/kotlin/org/jetbrains/skia/FontMetrics.kt
ApoloApps: done

skiko/src/commonMain/kotlin/org/jetbrains/skia/FontMetrics.kt
ApoloApps: done

skiko/src/commonTest/kotlin/org/jetbrains/skia/StrutStyleTests.kt
ApoloApps: done

ApoloApps:
@hub-bla everything has been adressed and the tests pass

hub-bla (approved):

PR #1211, opened 2026-06-09
Fix BreakIterator.setText Android crash

hub-bla:
Fixes:
```
03:57:36.872  A  java_vm_ext.cc:598] JNI DETECTED ERROR IN APPLICATION: incompatible array type short[] expected char[]: 0x7fd297fa00
                 java_vm_ext.cc:598]     in call to GetCharArrayRegion
                 java_vm_ext.cc:598]     from long org.jetbrains.skia.BreakIteratorKt._nSetText(long, java.lang.Object, int, java.lang.Object)
```

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -3,6 +3,9 @@ name: Skiko Tests
 on:
   workflow_call: # Allow being called by other workflows
 
+# TODO: Set up Android tests
+# Some tests can fail on Android while passing in other JVM targets.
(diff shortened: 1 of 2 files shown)


MatkovIvan (commented):
Please make sure that it's called from tests and verified by CI

MatkovIvan (approved):
Since we don't have infra for run it in Android env, we agreed that it's out of the scope here and we just need TODO/task for it for future follow up

PR #1213, opened 2026-06-15
Extract skottie into separate module

hub-bla:
This PR includes:

- Skottie extension module for each target
- Exposing some internal parts of skiko so that they can be used in extension modules

Required external changes :

- Compile Skia static libraries with `visibility=default` instead of `hidden`. I found it difficult to make symbols tagged as `hidden` visible again. Going from `visible` to `hidden` was much more straightforward
- Each static library that Skia produces is self-contained and thus, might have Skia core object files inside. This means we have to remove such files from Skottie during the Skia build

Closes [SKIKO-1146](https://youtrack.jetbrains.com/issue/SKIKO-1146)


skiko/gradle.properties
```diff
@@ -8,7 +8,7 @@ kotlin.js.yarn.lock.file=false
 deploy.version=0.0.0
 
 # a tag from https://github.com/JetBrains/skia/releases
-dependencies.skia=m148-fc4d63be29
+dependencies.skia=m148-9c7d55a254
```
hub-bla: Reminder to update it once the necessary PR in skia is merged

skiko/buildSrc/src/main/kotlin/SkikoProjectContext.kt
```diff
@@ -102,7 +105,7 @@ fun SkikoProjectContext.declareSkiaTasks() {
 
             val skiaReleaseTag = project.skiaVersion(target)
 
-            val skiaBaseUrl = "https://github.com/JetBrains/skia/releases/download/$skiaReleaseTag"
+            val skiaBaseUrl = "https://github.com/hub-bla/skia/releases/download/$skiaReleaseTag"
```
hub-bla: Reminder to update it once the necessary PR in skia is merged

ApoloApps:
what is the change in size without the module? Just curious

hub-bla:
Just a few:
| Library Name               | Size before (MB) | Size after (MB) |
|---------------------------|------------------|-----------------|
| libskiko-macos-arm64.dylib | 21.47            | 20.84           |
| libskiko-macos-x64.dylib   | 22.23            | 21.41           |
| libskiko-linux-x64.so      | 29.62            | 28.07           |
| skiko-windows-x64.dll      | 14.07            | 13.67           |
| skiko.wasm                 | 8.63             | 8.79            |

wasm has a slight increase because it required changing the skiko core to support dynamically linked side modules, which seems to introduce an unavoidable overhead caused by [internal data relotation function](https://github.com/llvm/llvm-project/issues/55608#issuecomment-1133059537)

.github/workflows/tests.yml
```diff
@@ -33,7 +33,8 @@ jobs:
             -Pskiko.native.enabled=true \
             -Dskiko.test.onci=true \
             -Dskiko.test.performance.enabled=false \
-            :skiko:awtTest
+            :skiko:awtTest \
```
igordmn: As there will be more modules in the future, a better way to run all tasks in `skiko` project is:
```
 ./gradlew -p skiko ...
            ...
            awtTest \
```

.github/workflows/tests.yml
```diff
@@ -332,14 +379,19 @@ jobs:
...
+            ./skiko/skiko-skottie/hs_err_pid*.log
           retention-days: 5
 
       - uses: test-summary/action@v2
         name: 'Test Summary'
         with:
-          paths: "./skiko/build/test-results/**/TEST-*.xml"
+          paths: |
+            ./skiko/build/test-results/**/TEST-*.xml
+            ./skiko/skiko-skottie/build/test-results/**/TEST-*.xml
```
igordmn:
```suggestion
            ./skiko/**/build/test-results/**/TEST-*.xml
```

skiko/buildSrc/src/main/kotlin/tasks/configuration/NativeTasksConfiguration.kt
```diff
@@ -29,6 +32,17 @@ import symbols.HideSkiaSymbolsTask
 import java.io.File
 import kotlin.collections.plus
 
+private val nativeSymbolSourcesOsAttribute =
```
igordmn, line 36: What is the purpose of this Gradle metadata in the artifacts?

skiko/buildSrc/src/main/kotlin/tasks/configuration/WasmTasksConfiguration.kt
```diff
@@ -143,19 +172,28 @@ fun SkikoProjectContext.declareWasmTasks() {
         )
 
         buildSuffix.set("d8")
-        jsOutputFileName.set("skikod8.mjs") // this determines the name .wasm file too
-        libOutputFileName.set("skikod8.wasm")
+        jsOutputFileName.set(if (isSideModule) "${libBaseName}d8.wasm" else "skikod8.mjs") // this determines the name .wasm file too
```
igordmn: If it is not only "js" files, we need to rename it to something like `outputFileName`

skiko/buildSrc/src/main/kotlin/tasks/configuration/WasmTasksConfiguration.kt
```diff
@@ -110,6 +132,9 @@ fun SkikoProjectContext.declareWasmTasks() {
             // skiko.mjs is referenced in karma.config.d/*/config.js
             // so symbols must be replaced right after linking
             val jsFile = outDir.asFile.get().walk().first { it.name == jsOutputFileName.get() }
+            if (jsFile.extension != "mjs") {
+                return@doLast
+            }
```
igordmn:
```suggestion
            if (jsOutputFileName.get().endsWith("mjs")) {
                return@doLast
            }
            val jsFile = outDir.asFile.get().walk().first { it.name == jsOutputFileName.get() }
```
Having check on the property and in the beginning shows the intention better

.github/workflows/tests.yml
```diff
@@ -33,7 +33,8 @@ jobs:
             -Pskiko.native.enabled=true \
```
igordmn, line 33: We need to build a Snapshot build from this branch before merging to be sure that publishing works correctly

.github/workflows/tests.yml
```diff
@@ -33,7 +33,8 @@ jobs:
...
             -Dskiko.test.onci=true \
```
igordmn, line 34: Can we write tests that check published to maven local artifacts and run them on GitHub Actions?

Some simple tests would be enough, it would ensure correct linking, initialization and bundled resources

skiko/buildSrc/src/main/kotlin/tasks/configuration/NativeTasksConfiguration.kt
hub-bla: To resolve right artifacts from other modules

On jvm, each exstension module creates consumable configurations with the skia symbols that it needs from the core. Core module then resolves them and assembles them into the full list of symbols that it need to expose

On native, core creates consumable configuration with skia libraries is using so that  extension modules can properly patch the skia symbols in skia extension libs

On wasm, each extension module creates consumable configuration with the generated wasm binary so that core can resolve them and link in the creation of core wasm. (That linking step also expose necessary symbols required by extension modules)

.github/workflows/tests.yml
hub-bla: If I understand correctly, you mean adding a small test project that uses all the published extension modules and the core module?

igordmn: Something like this, yes. Also, just a core module without extensions.

When we have a complicated build there are a lot of things that can go wrong, and such tests can help noticing this.

igordmn: We probably need to write additional steps after each "dry-run".

skiko/karma.config.d/js/config.js
```diff
@@ -18,6 +19,7 @@ debug(`karma generatedAssetsPath: ${generatedAssetsPath}`);
 config.proxies = {
     "/wasm/": wasmPath,
     "/resources": path.resolve(basePath, "kotlin"),
+    "/skiko-skottie.wasm": skottieWasmPath,
```
MatkovIvan, line 22: It seems it might be a breaking change in some situations and requires Compose Gradle DSL adoption to handle such submodules.
I suggest including it in the next "major" (0.x) skiko release only, not patch

skiko/skiko-skottie/src/webMain/kotlin/org/jetbrains/skia/skottie/SkottieLibrary.kt
```diff
@@ -0,0 +1,9 @@
+package org.jetbrains.skia.skottie
+
+internal actual object SkottieLibrary {
```
eymar, line 3: looks like this actual can be shared between js and wasm in webMain?

skiko/skiko-skottie/build.gradle.kts
```diff
@@ -0,0 +1,401 @@
...
+)
+
+repositories {
+    mavenCentral()
+    google()
+}
+
+kotlin {
+    compilerOptions {
+        languageVersion.set(KotlinVersion.KOTLIN_2_2)
```
eymar: Ideally, the languageVersion and apiVersion would be set once per entire project so it's easier to update it just in one place. 

Or there can be a constant in buildSrc and that constant would be used in all modules. WDYT?

skiko/src/webMain/resources/pre-setup.mjs
```diff
@@ -5,10 +5,59 @@ export const loadedWasm = {
...
+
+const extensionLoadPromises = new Map();
+
+export const loadSkikoExtension = (extensionPath) => {
+    if (extensionLoadPromises.has(extensionPath)) return extensionLoadPromises.get(extensionPath);
+    const loadPromise = awaitSkikoCore.then(async (module) => {
+        // Mimic the wasm loader: route this side module through
+        // locateFile so Emscripten does not fall back to a file:// scriptDirectory.
+        const originalLocateFile = module.locateFile;
+        module.locateFile = (path, prefix) => {
```
eymar: as I understand eventually there will be more than 1 side module. So 2+ side modules will call `loadSkikoExtension` and they will concurrently attempt to modify / reset `module.locateFile`.

Maybe changing it permanently is an option:

```
const originalLocateFile = module.locateFile;
module.locateFile = (path, prefix) => {
    // If path is already an absolute URL, don't prepend scriptDirectory
    if (path.startsWith("http://") || path.startsWith("https://") || path.startsWith("blob:")) {
        return path;
    }
    return originalLocateFile(path, prefix);
};
```
WDYT?

skiko/skiko-skottie/src/webMain/resources/pre-skiko-skottie.mjs
```diff
@@ -0,0 +1,21 @@
...
+const ensureSkottieLoaded = () => {
+    if (!skottieLoadPromise) {
+        skottieLoadPromise = loadSkikoExtension(skottieWasm);
+    }
+    return skottieLoadPromise;
+};
+
+registerSkikoWasmReadyCallback(() => ensureSkottieLoaded());
+
+const isSideModuleLoaded = () => skottieLoadPromise !== null;
```
eymar: `skottieLoadPromise` might be not null (initialized as expected), but at the same time it could be rejected or not resolved yet. 

Perhaps it's an option to add one more state to track the status of the Promise and in `isSideModuleLoaded` just return true if the status is ok.

skiko/skiko-skottie/src/webMain/kotlin/org/jetbrains/skia/skottie/SkottieLibrary.kt
hub-bla: Done

skiko/skiko-skottie/build.gradle.kts
hub-bla: Yeah, I agree it will be easier later to maintain. Done

skiko/src/webMain/resources/pre-setup.mjs
hub-bla: Thanks, didn't notice it at first. Changed to be updated only once

skiko/skiko-skottie/src/webMain/resources/pre-skiko-skottie.mjs
hub-bla: It shouldn't be unresolved once we check the `skottieLoadPromise`. At least this is what I intended with resolving the promises in `awaitSkiko` so that once we run the app once everything is resolved. 
But yeah, the current check will return true for rejected promise as well.

Fixed

.github/workflows/tests.yml
hub-bla: I added a new sample project for this purpose (it's based on the `SkiaMultiplatformSample`) and added test runs in ci for it.  Let me know if this logic to run sample project with auto exit is fine and if so I'll add it to the `SkiaMultiplatformSample` as well. It will cover the testing part with just the core

samples/SkikoExtensionsSample/scripts/check-web-sample.sh
```diff
@@ -0,0 +1,15 @@
...
+SAMPLE_DIR="$ROOT_DIR/samples/SkikoExtensionsSample"
+PLAYWRIGHT_BIN="${PLAYWRIGHT_BIN:-playwright}"
+
+run_web_sample() {
+  local task="$1"
+  WEB_SERVER_COMMAND="$ROOT_DIR/gradlew -p $SAMPLE_DIR --no-daemon --stacktrace $task" \
+    "$PLAYWRIGHT_BIN" test --config "$SAMPLE_DIR/scripts/playwright.config.js"
+}
+
+run_web_sample :jsBrowserDevelopmentRun
```
igordmn, line 14: Where is the logic that stops the server after some time?

skiko/buildSrc/src/main/kotlin/tasks/configuration/NativeTasksConfiguration.kt
igordmn: My question was about why we set the attributes:
```
attribute(nativeSymbolSourcesOsAttribute, os.name)
attribute(nativeSymbolSourcesArchAttribute, arch.name)
attribute(nativeSymbolSourcesUikitSimAttribute, isUikitSim.toString())
```

This will set a Gradle metadata that is published to Maven repo. But without using it somewhere else on the consumer side, it shouldn't do anything, right?

samples/SkikoExtensionsSample/scripts/check-web-sample.sh
hub-bla: Playwright is responsible for running the server as well as closing it after the check that there are no runtime errors in the sample app

```
  webServer: {
    command,
    url: baseURL,
    timeout: 180_000,
  }
```

`command` here is basically the `./gradlew -p $SAMPLE_DIR --no-daemon --stacktrace :jsBrowserDevelopmentRun/:wasmJsBrowserDevelopmentRun`

skiko/buildSrc/src/main/kotlin/tasks/configuration/NativeTasksConfiguration.kt
hub-bla: These attributes are not published to Maven. These are just for internal usage for the purpose of "communication" between the skiko modules (for how see the first reply in this thread). It replaces relying purely on other modules tasks as it's not the gradle best practice.

See: https://docs.gradle.org/current/userguide/multi_project_builds.html#sec:depending_on_output_of_another_project

igordmn (approved):

.github/workflows/tests.yml
```diff
@@ -238,6 +250,14 @@ jobs:
           if-no-files-found: error
           retention-days: 5
 
+      - uses: actions/upload-artifact@v4
+        name: 'Upload skiko-skottie linuxArm64 test binary as artifact'
+        with:
+          name: test-binary-skiko-skottie-linuxArm64
+          path: ./skiko/skiko-skottie/build/bin/linuxArm64
```
MatkovIvan: Just use
```suggestion
          path: ./skiko/**/build/bin/linuxArm64
```

In existing step, do not duplicate it per module

.github/workflows/tests.yml
```diff
@@ -201,13 +209,17 @@ jobs:
           name: test-reports-linux
           path: |
             ./skiko/build/reports/tests
+            ./skiko/skiko-skottie/build/reports/tests
             ./skiko/src/jvmTest/screenshots/*_actual.png
+            ./skiko/skiko-skottie/src/jvmTest/screenshots/*_actual.png
             ./skiko/hs_err_pid*.log
+            ./skiko/skiko-skottie/hs_err_pid*.log
```
MatkovIvan:
```suggestion
            ./skiko/**/build/reports/tests
            ./skiko/**/src/jvmTest/screenshots/*_actual.png
            ./skiko/**/hs_err_pid*.log
```

samples/SkikoExtensionsSample/gradle/wrapper/gradle-wrapper.properties
```diff
@@ -0,0 +1,7 @@
+distributionBase=GRADLE_USER_HOME
+distributionPath=wrapper/dists
+distributionUrl=https\://services.gradle.org/distributions/gradle-9.3.0-bin.zip
```
MatkovIvan: skiko uses gradle-9.5.0 in all plases now

samples/SkikoExtensionsSample/build.gradle.kts
```diff
@@ -0,0 +1,388 @@
+@file:OptIn(org.jetbrains.kotlin.gradle.ExperimentalWasmDsl::class)
```
MatkovIvan: import over FQNs

samples/SkikoExtensionsSample/build.gradle.kts
MatkovIvan, line 1: It seems there is a log of logic in this file that was just copied. It's far from obvious, so it is worth being extracted/shared.

Or as alternative variant - just add it into an existing sample project, I don't see anything wrong to reuse it in tfor this case

skiko/skiko-skottie/src/commonMain/kotlin/org/jetbrains/skia/sksg/InvalidationController.kt
```diff
@@ -42,19 +36,24 @@ class InvalidationController internal constructor(ptr: NativePointer) : Managed(
...
         Stats.onNativeCall()
 
-        InvalidationController_nReset(_ptr)
+        InvalidationController_nReset(nativePtr)
         return this
     }
 }
 
+private fun makeInvalidationController(): NativePointer {
+    SkottieLibrary.load()
```
MatkovIvan: Using it instead of `companion object { init` allows bypassing it in the case of an internal constructor. What's the reason to change the pattern?

hub-bla: I thought calls would be in a bit different order but rechecked it and yes. `companion object { init` is completly fine. I'll revert that

samples/SkikoExtensionsSample/build.gradle.kts
hub-bla: The intention behind creating it as a separate sample project was to avoid overthinking the setup and to have a simple, short-term project for testing the extension module.

I plan to remove it in the near future as part of the Ganesh extraction work. As a result, `SkiaMultiplatformSample` will adopt the extension module (ganesh), making this separate sample project obsolete and allowing it to be removed.

Given that it's only a temporary solution, does leaving it as is sounds reasonable to you or would you still prefer doing something with it?

skiko/skiko-skottie/src/commonMain/kotlin/org/jetbrains/skia/sksg/InvalidationController.kt
hub-bla: Done

samples/SkikoExtensionsSample/build.gradle.kts
hub-bla: Done

samples/SkikoExtensionsSample/gradle/wrapper/gradle-wrapper.properties
hub-bla: Done

samples/SkikoExtensionsSample/build.gradle.kts
MatkovIvan: If it's intermediate state - not blocking

.github/workflows/tests.yml
```diff
@@ -51,14 +51,18 @@ jobs:
           name: test-reports-macos
           path: |
             ./skiko/build/reports/tests
+            ./skiko/skiko-skottie/build/reports/tests
```
MatkovIvan:
```suggestion
            ./skiko/**/build/reports/tests
```

Please update all similar places

.github/workflows/tests.yml
```diff
@@ -116,7 +120,9 @@ jobs:
         if: failure()
         with:
           name: test-reports-ios
-          path: ./skiko/build/reports/tests
+          path: |
+            ./skiko/build/reports/tests
+            ./skiko/skiko-skottie/build/reports/tests
```
MatkovIvan:
```suggestion
            ./skiko/**/build/reports/tests
```

.github/workflows/tests.yml
```diff
@@ -238,6 +247,14 @@ jobs:
           if-no-files-found: error
           retention-days: 5
 
+      - uses: actions/upload-artifact@v4
+        name: 'Upload skiko-skottie linuxArm64 test binary as artifact'
+        with:
+          name: test-binary-skiko-skottie-linuxArm64
+          path: ./skiko/**/build/bin/linuxArm64
```
MatkovIvan: It should be combined with previous step. Now it duplicates it

.github/workflows/tests.yml
hub-bla: Done

.github/workflows/publish-dry-run.yml
```diff
@@ -42,14 +42,40 @@ jobs:
...
-            ./gradlew --no-daemon --stacktrace \
+            ./gradlew -p skiko --no-daemon --stacktrace \
               -Pskiko.wasm.enabled=true \
-              :skiko:publishToMavenLocal
+              publishToMavenLocal
+
+      - uses: actions/setup-node@v4
+        name: 'Setup Node.js'
+        with:
+          node-version: 22
```
MatkovIvan, line 63: Why 22? 24 is active LTS version at the moment

MatkovIvan (approved):

.github/workflows/publish-dry-run.yml
hub-bla: No particular reason. Both have LTS status so I took the older one

.github/workflows/tests.yml
hub-bla: Snapshot publishing passed: https://buildserver.labs.intellij.net/buildConfiguration/Skiko_PublishSnapshot/1002443845?buildTab=overview

igordmn (approved):

PR #1217, opened 2026-06-19
Fix memory leak in drawVertices JNI call

LionPa:
**What and why**
This PR fixes a memory leak in the `Canvas.drawVertices` JNI implementation. The `indexArr` elements were accessed via `GetShortArrayElements` but were never released, which could lead to pinned arrays or memory exhaustion over time.

**Testing**
Successfully built the native C++ libraries locally on Windows x64. Ran tests — all passed except `TextLineTest.emojiTest`, which seems to be a local Windows environment issue unrelated to Canvas bindings.

**Release Notes**
* Fixed a JNI memory leak in `Canvas.drawVertices`.

**Issue:**
https://youtrack.jetbrains.com/issue/SKIKO-1149/Memory-leak-pinned-array-in-Canvas.drawVertices-JNI-due-to-missing-ReleaseShortArrayElements


hub-bla (approved):
Hi, thanks for contributing. Good catch!

PR #1219, opened 2026-06-24
Wasm texture adoption

serjsysoev:
This chain of commits enables adoption of external WebGL textures.

"Add Wasm texture registration helper" is required because skia expects a texture id, but on WebGL the textures don't have ids. The ids are provided to skia by emscripten. [Here](https://github.com/google/skia/blob/7c071bfb59ddcf6031d8e9d88db740c5bcd9440a/modules/canvaskit/webgl.js#L206-L224) is the same hack in skia sources.

"Expose borrowed recording contexts" is required to get `DirectContext` to pass into `adoptTextureFrom`. I also fixed ownership: the context returned by canvas/surface should be borrowed, since otherwise we will accidentally dispose of a context that is still being used for rendering.

### DIFF
--- a/skiko/src/commonMain/kotlin/org/jetbrains/skia/Image.kt
+++ b/skiko/src/commonMain/kotlin/org/jetbrains/skia/Image.kt
@@ -176,6 +176,47 @@ class Image internal constructor(ptr: NativePointer) : RefCnt(ptr), IHasImageInf
             }
         }
 
+        /**
+         * Creates GPU-backed SkImage from backendTexture associated with context.
+         *
+         * Skia will assume ownership of the resource and will release it when no longer needed.
+         * A non-null Image is returned if format of backendTexture is recognized and supported.
+         * Recognized formats vary by GPU backend.
+         *
+         * @param context         GPU context
+         * @param backendTexture  texture residing on GPU
+         * @param origin          origin of backendTexture
+         * @param colorType       color type of the resulting Image
+         * @param alphaType       alpha type of the resulting Image
+         * @return                created Image
+         *
+         * @throws RuntimeException - if nullPtr is returned.
+         */
+        fun adoptTextureFrom(
+            context: DirectContext,
+            backendTexture: BackendTexture,
+            origin: SurfaceOrigin,
+            colorType: ColorType,
+            alphaType: ColorAlphaType,
+        ): Image {
+            return try {
+                Stats.onNativeCall()
+                val ptr = _nAdoptTextureFromAlphaType(
+                    getPtr(context),
+                    getPtr(backendTexture),
+                    origin.ordinal,
+                    colorType.ordinal,
+                    alphaType.ordinal,
+                )
+                if (ptr == NullPointer) throw RuntimeException("Failed to Image::makeFromTexture")
+                Image(ptr)
+            }
+            finally {
+                reachabilityBarrier(context)
+                reachabilityBarrier(backendTexture)
+            }
+        }
+
         init {
             staticLoad()
         }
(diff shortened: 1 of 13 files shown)


skiko/src/commonMain/kotlin/org/jetbrains/skia/Image.kt
```diff
@@ -490,3 +531,12 @@ external fun _nAdoptTextureFrom(
     surfaceOrigin: Int,
     colorType: Int
 ): NativePointer
+
+@ExternalSymbolName("org_jetbrains_skia_Image__1nAdoptTextureFromAlphaType")
+external fun _nAdoptTextureFromAlphaType(
```
eymar: let's keep it private like other methods. I assume `_nAdoptTextureFrom` is public my mistake.

skiko/src/wasmJsMain/kotlin/org/jetbrains/skiko/GLTexture.wasm.kt
```diff
@@ -0,0 +1,16 @@
+@file:OptIn(kotlin.js.ExperimentalWasmJsInterop::class)
+
+package org.jetbrains.skiko
+
+import kotlin.js.JsAny
+
+@JsFun(
+    """(gl, texture) => {
+        const textureHandle = gl.getNewId(gl.textures);
+        gl.textures[textureHandle] = texture;
```
eymar, line 10: Does it ever get cleaned up? Who is responsible for clearing `gl.textures`?

skiko/src/commonMain/kotlin/org/jetbrains/skia/Surface.kt
```diff
@@ -1115,7 +1117,7 @@ private external fun _nGenerationId(ptr: NativePointer): Int
 private external fun _nNotifyContentWillChange(ptr: NativePointer, mode: Int)
 
 @ExternalSymbolName("org_jetbrains_skia_Surface__1nGetRecordingContext")
-private external fun _nGetRecordingContext(ptr: NativePointer): NativePointer
+private external fun _nGetSurfaceRecordingContext(ptr: NativePointer): NativePointer
```
eymar, line 1120: Wdyt about renaming then ExternalSymbolName for consitency?

skiko/src/wasmJsMain/kotlin/org/jetbrains/skiko/GLTexture.wasm.kt
```diff
@@ -0,0 +1,16 @@
...
+        return textureHandle;
+    }"""
+)
+private external fun pushTexture(gl: GLInterface, texture: JsAny): Int
+
+fun pushTexture(texture: JsAny): Int = pushTexture(GL, texture)
```
eymar, line 36: Could you please add some KDoc for this method? and let's add `@ExperimentalSkikoApi` on it?

eymar, line 36: `texture: JsAny` - could it be `WebGLTexture` ? Or you prefer avoiding a dependency on kotlinx-browser?

serjsysoev: Skiko already has kotlinx-browser dependency, so it shouldn't add any dependencies skiko doesn't already have. We are using [JetBrains/kotlin-wrappers](https://github.com/jetbrains/kotlin-wrappers) for our JS interop though, so we have `web.gl.WebGLTexture` instead of `org.khronos.webgl.WebGLTexture`. That's why I would like to avoid specifying a type from a specific interop library here. But if you prefer to have a specific type here, I'm ok with re-casting in our code

skiko/src/wasmJsMain/kotlin/org/jetbrains/skiko/GLTexture.wasm.kt
serjsysoev: You are right, I completely forgot to introduce the unregister operation. I've added this API specifically for use with `Image.adoptTextureFrom`, it takes ownership of the texture and disposes of the texture id in the table itself, so it wasn't a problem for me :)

Added `unregisterTexture`

serjsysoev:
I have no idea why `:skiko:awtTest` crashed on CI after fixups (SIGSEGV in JRE), the only thing that may be of interest in my changes is `ExternalSymbolName` change for `_nGetSurfaceRecordingContext`... The tests run fine on my MacOS machine... Is it possible to rerun the test?

skiko/src/wasmJsMain/kotlin/org/jetbrains/skiko/GLTexture.wasm.kt
eymar: I see. Makes sense to me. Let's keep it JsAny for now.

eymar:
LGTM!

@MatkovIvan could you please have a look too? Some new API is exposed  (reasonably)

eymar (approved):

MatkovIvan (approved):

PR #1222, opened 2026-06-30
Setup runtime fat jar publication

hub-bla:
Fixes [SKIKO-1131](https://youtrack.jetbrains.com/issue/SKIKO-1131)


kropp (approved):

skiko/buildSrc/src/main/kotlin/publishing.kt
```diff
@@ -18,6 +19,34 @@ import org.jetbrains.kotlin.gradle.plugin.KotlinPlatformType
 
 private val SkikoProjectContext.publishing get() = project.extensions.getByType(PublishingExtension::class.java)
 
+/**
+ * Shared POM metadata for all Skiko publications (license, project URL, SCM, developers).
+ *
```
MatkovIvan:
```suggestion
```

skiko/ci/build.gradle.kts
```diff
@@ -109,3 +118,162 @@ fun connectToGitHub() =
     GitHubBuilder()
         .withOAuthToken(System.getenv("SKIKO_GH_RELEASE_TOKEN"))
         .build()
+
+
+configure<PublishingExtension> {
+    repositories {
+        maven {
+            name = "ComposeRepo"
+            url = uri(skiko.composeRepoUrl)
```
MatkovIvan: As far as I understand `skiko/buildSrc/src/main/kotlin/publishing.kt` should be a single source of truth for publishing.
Why do you need a copy here?

skiko/ci/build.gradle.kts
```diff
@@ -109,3 +118,162 @@ fun connectToGitHub() =
...
+                localDir = artifactsDir.resolve(
+                    "${skiko.deployVersion}/${SkikoArtifacts.DEFAULT_GROUP_ID}/$artifactId"
+                )
+            )
+        }
+    }
+
+    val downloadInputs = tasks.register(
+        "download${publicationName.replaceFirstChar { it.uppercaseChar() }}Inputs",
+        DownloadFromSpaceMavenRepoTask::class.java
```
MatkovIvan: Why do we need to download something here?
How to build it locally without intermediate publishing?

skiko/ci/build.gradle.kts
```diff
@@ -109,3 +118,162 @@ fun connectToGitHub() =
...
+    val publishTask = tasks.named(publishTaskName)
+    publishAllJvmRuntimeAllPublications.configure {
+        dependsOn(publishTask)
+    }
+}
+
+// Resolves the actual main runtime jar downloaded into `module.localDir`. For releases the canonical
+// "<artifactId>-<version>.jar" name is used. For unique (timestamped) SNAPSHOTs the real file name is
+// read from maven-metadata.xml (e.g. "<artifactId>-0.0.0-20260630.105840-3.jar").
+fun resolveDownloadedRuntimeJar(module: ModuleToUpload): File {
```
MatkovIvan: I'd move it out of the "root" `build.gradle.kts` file and place it in the current `buildSrc` structure to avoid mixing "what" we're doing with "how"

skiko/ci/build.gradle.kts
hub-bla: >Why do we need to download something here?

Because the fat jar needs artifacts from three different Os'es. We don't have the ability to build all of them on one machine currently.

>How to build it locally without intermediate publishing?

This is the problem now. I guess the best way would be to generalize this task so that on ci we will make full fat jar and locally we will produce it only with artifacts needed on current OS. Wdyt?

MatkovIvan: Yep, it sounds like we need some reduced variant to make it work locally

kropp: Missed this initially. We can add an artifact dependency in TeamCity (all these builds are part of the same chain) and download artifacts directly from other builds, if they are published there.
That way, we skip potentially error prone downloading, and it would easier reproducible locally

hub-bla: I've updated the entire flow

On TeamCity, each jvm platform adds their produced runtime jars into `awtRuntimeAllInputs/<artifact-prefix>/` directory.
These are later merged from other builds into one `awtRuntimeAllInputs/<artifact-prefix>/` that contains all necessary jars.

The artifacts rules inside teamcity config are recreating the directory structure so that it remains the same.

Locally, it simply creates fat jar from the locally built jars.

skiko/.gitignore
```diff
@@ -5,6 +5,9 @@ local.properties
 .DS_Store
 dependencies/
 
+# CI-only directory for the AWT runtime fat jar
+awtRuntimeAllInputs/
```
MatkovIvan: We already download skia binaries. Can we reuse the same folder for all downloads?

hub-bla: Sure, I'll move the `awtRuntimeAllInputs` dir into dependencies folder

hub-bla: Updated here and in teamcity config as well

MatkovIvan (approved):

PR #1224, opened 2026-07-04
Native Linux SkiaLayer: Wayland/EGL and X11/GLX backends behind a windowing seam

hi-mod:
### Summary

Implements `SkiaLayer` for the native Linux targets (`linuxX64`, `linuxArm64`), replacing the TODO stubs in `SkiaLayer.linux.kt`. Rendering goes through a small windowing seam (`LinuxWindow`, `LinuxGlContext`) with two backends: Wayland/EGL as the primary path and X11/GLX for X-only environments. Review feedback on the companion Compose PR (compose-multiplatform-core#3189) asked that new Linux windowing not assume one display server; the seam keeps everything above it compositor-agnostic.

### Windowing seam

`SkiaLayer`, the GL context handler, and the redrawer talk only to `LinuxWindow`/`LinuxGlContext`. Backend selection is the consumer's: construct a `WaylandWindow` or an `X11Window` and attach the layer.

### Wayland/EGL backend

- `WaylandWindow`: xdg-shell toplevel. HiDPI uses `fractional-scale-v1` plus `viewporter`, falling back to the `wl_output` integer scale. Server-side decorations are requested through `xdg-decoration` where the compositor advertises the manager (KWin does; GNOME does not, so the window stays borderless there until a libdecor fallback lands). Frames are compositor-paced via `wl_callback`; close arrives as `xdg_toplevel.close`.
- `WaylandEglContext`: EGL context creation and buffer swap.
- `display` and `surface` are public with a documented ownership contract, so the Compose event pump can bind `wl_seat` and friends off the same connection without owning either object.

### X11/GLX backend

`X11Window` selects a double-buffered GLX visual, handles `WM_DELETE_WINDOW`, and reads scale from `Xft.dpi`; `GlContext` drives vsync through `glXSwapIntervalEXT`. `LinuxMainDispatcher` is a mutex-guarded work queue drained from the host event loop, since Kotlin/Native has no `Dispatchers.Main` on Linux.

### Build and cinterop

Three cinterops: `x11gl`, `waylandegl`, and `xkbcommon`. The `waylandegl` def file is generated per target because its include path points at build-time protocol codegen output. Protocol client code is generated with `wayland-scanner` from the system `wayland-protocols` for: xdg-shell, viewporter, fractional-scale-v1, xdg-decoration-unstable-v1, cursor-shape-v1 (plus tablet-v2, which its generated code references), and text-input-unstable-v3. The last three, like the xkbcommon klib, have no consumer inside skiko; they are published for the Compose event pump.

Host headers are added with `-idirafter` so the Konan sysroot's per-arch glibc headers keep priority when cross-compiling to `linuxArm64`. When `/opt/arm-gnu-toolchain` is absent, the build falls back to a distro `aarch64-linux-gnu-g++`.

### CI changes that need reviewer attention

1. Linux native targets are now registered only on Linux hosts (`build.gradle.kts`).
2. The Linux publish dry run drops `-Pskiko.native.enabled` in the `linux-compat` image. Cinterop tasks cannot run there: Konan's bundled libclang needs `GLIBC_2.29` and the image ships glibc 2.26 by design. The following step in `linux-amd64` already publishes both native targets.
3. The docs job moves to `linux-amd64` for the same reason.
4. The `linux-amd64` image gains `libwayland-dev`, `libegl1-mesa-dev`, `libxkbcommon-dev` (host and arm64) and vendors wayland-protocols 1.45 from the release tarball, because focal packages 1.20 and fractional-scale-v1 first shipped in 1.31.
5. `x11gl.def` no longer links `-lfreetype`. Skia's Linux archives bundle freetype statically; with the shared library on the link line those members never entered the test binary, and lld rejected the generated version script's `local: FT_*` entries as undefined. This is what failed `linkDebugTestLinuxArm64` in the previous CI run.

### Verification

- KDE Plasma Wayland session (KWin): a raw-`SkiaLayer` demo renders compositor-paced, gets a KWin titlebar through xdg-decoration, follows a compositor-driven resize from 800x600 to 1100x750, and exits 0 on close.
- X11 backend on the same machine via Xwayland: renders and exits 0 on `WM_DELETE_WINDOW`.
- `linkDebugTestLinuxArm64` (cross-compiled from x64) links against the same ARM GNU toolchain 10.3-2021.07 and focal arm64 libraries the CI image uses.
- `compileKotlinLinuxX64` and `compileKotlinLinuxArm64` are green; both native publications build to mavenLocal.

### DIFF
--- a/.github/workflows/docs.yml
+++ b/.github/workflows/docs.yml
@@ -23,10 +23,13 @@ jobs:
       - uses: ./.github/actions/setup-prerequisites
         name: 'Setup Prerequisites'
 
+      # linux-amd64 rather than linux-compat: with skiko.native.enabled the Linux
+      # targets run cinterop tasks, which need a glibc newer than compat's 2.26
+      # (Konan's libclang requires GLIBC_2.29) plus the X11/Wayland dev headers.
       - uses: ./.github/actions/docker-skiko-run
         name: 'Build Dokka Documentation'
         with:
-          image_name: linux-compat
+          image_name: linux-amd64
           command: |
             ./gradlew --no-daemon --stacktrace \
               -Pskiko.native.enabled=true \
--- a/.github/workflows/publish-dry-run.yml
+++ b/.github/workflows/publish-dry-run.yml
@@ -61,13 +61,16 @@ jobs:
       - uses: ./.github/actions/setup-prerequisites
         name: 'Setup Prerequisites'
 
+      # No -Pskiko.native.enabled here: the Linux native targets carry cinterops,
+      # and cinterop tasks cannot run in linux-compat (its glibc 2.26 predates the
+      # GLIBC_2.29 required by Konan's bundled libclang). The native publications
+      # are produced by the next step in the linux-amd64 image instead.
       - uses: ./.github/actions/docker-skiko-run
         name: 'Publish x64 Maven Local'
         with:
           image_name: linux-compat
           command: |
             ./gradlew --no-daemon --stacktrace \
-              -Pskiko.native.enabled=true \
               -Pskiko.awt.enabled=true \
               -Pskiko.arch=x64 \
                :skiko:publishToMavenLocal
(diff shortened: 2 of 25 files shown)


hi-mod:
@hub-bla, I pushed a fix in 146ef3d0 to add Wayland support and get the CI green (the Dockerfile needed the Android SDK for Dokka).

PR #1227, opened 2026-07-06
Apply `RenderNode.alpha` during draw

MatkovIvan:
Copies HWUI approach: https://android.googlesource.com/platform/frameworks/base/+/1cdfff555f4a21f71ccc978290e2e212e2f8b168/libs/hwui/pipeline/skia/RenderNodeDrawable.cpp#200
Required to fix [CMP-10436](https://youtrack.jetbrains.com/issue/CMP-10436) [skiko/iOS] CompositingStrategy.ModulateAlpha bakes alpha into the recorded picture at record time; animating alpha leaves content invisible until next content invalidation

We have behavioral tests are only on compose side, so it's not included here 😢

### DIFF
--- a/skiko/src/commonMain/cpp/common/node/RenderNode.cpp
+++ b/skiko/src/commonMain/cpp/common/node/RenderNode.cpp
@@ -1,5 +1,6 @@
 #include <SkNWayCanvas.h>
 #include <SkPathBuilder.h>
+#include <SkPaintFilterCanvas.h>
 #include <SkPicture.h>
 #include <SkShadowUtils.h>
 #include "node/RenderNode.h"
@@ -54,6 +55,25 @@ protected:
     void onDrawDrawable(SkDrawable* drawable, const SkMatrix* matrix) override {
         drawable->draw(this, matrix);
     }
+};
+
+class AlphaModulateFilterCanvas : public SkPaintFilterCanvas {
+public:
+    AlphaModulateFilterCanvas(SkCanvas* canvas, float alpha)
+        : SkPaintFilterCanvas(canvas), alpha(alpha) {}
+
+protected:
+    bool onFilter(SkPaint& paint) const override {
+        paint.setAlphaf(paint.getAlphaf() * alpha);
+        return true;
+    }
+
+    void onDrawDrawable(SkDrawable* drawable, const SkMatrix* matrix) override {
+        drawable->draw(this, matrix);
+    }
+
+private:
+    float alpha;
 };
 
 RenderNode::RenderNode(const sk_sp<RenderNodeContext>& context)
@@ -252,7 +272,12 @@ void RenderNode::onDraw(SkCanvas* canvas) {
         canvas->save();
     }
 
-    canvas->drawDrawable(this->contentCache.get());
+    if (this->alpha < 1.0f && !this->layerPaint) {
+        AlphaModulateFilterCanvas alphaCanvas(canvas, this->alpha);
+        alphaCanvas.drawDrawable(this->contentCache.get());
+    } else {
+        canvas->drawDrawable(this->contentCache.get());
+    }
 }
 
 SkRect RenderNode::onGetBounds() {
--- a/skiko/src/commonMain/kotlin/org/jetbrains/skiko/node/RenderNode.kt
+++ b/skiko/src/commonMain/kotlin/org/jetbrains/skiko/node/RenderNode.kt
@@ -28,6 +28,14 @@ class RenderNode internal constructor(ptr: NativePointer, managed: Boolean = tru
         Stats.onNativeCall()
     }
 
+    /**
+     * Optional paint used when compositing this RenderNode into an offscreen layer.
+     *
+     * When set, the paint alpha is applied to the composited layer content and therefore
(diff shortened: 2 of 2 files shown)


hub-bla (approved):

skiko/src/commonMain/cpp/common/node/RenderNode.cpp
```diff
@@ -252,7 +272,12 @@ void RenderNode::onDraw(SkCanvas* canvas) {
         canvas->save();
     }
 
-    canvas->drawDrawable(this->contentCache.get());
+    if (this->alpha < 1.0f && !this->layerPaint) {
+        AlphaModulateFilterCanvas alphaCanvas(canvas, this->alpha);
```
igordmn, line 276: Do we apply `alpha` in Compose at the moment? If so, we need to not apply it there, and let's create a PR there before merging this PR.

skiko/src/commonMain/cpp/common/node/RenderNode.cpp
```diff
@@ -252,7 +272,12 @@ void RenderNode::onDraw(SkCanvas* canvas) {
         canvas->save();
     }
 
-    canvas->drawDrawable(this->contentCache.get());
+    if (this->alpha < 1.0f && !this->layerPaint) {
```
igordmn, line 275: In which cases `layerPaint` isn't set?

skiko/src/commonMain/cpp/common/node/RenderNode.cpp
MatkovIvan: Yes, Compose adoption is needed.
Added TODOs in https://github.com/JetBrains/compose-multiplatform-core/pull/3191, will open adoption PR after merging it.

skiko/src/commonMain/cpp/common/node/RenderNode.cpp
MatkovIvan: When we don't need offscreen layer.
Compose logic to determine it is [quite complex](https://github.com/JetBrains/compose-multiplatform-core/blob/95549cb7a90c54d5abd0bb14c60090239ad10d16/compose/ui/ui-graphics/src/skikoMain/kotlin/androidx/compose/ui/graphics/layer/SkiaGraphicsLayer.skiko.kt#L486-L494)

skiko/src/commonMain/cpp/common/node/RenderNode.cpp
igordmn: > will open adoption PR after merging it.

We need to prepare and test it before merging this PR, to be sure that we can implement it correctly

igordmn: If it is also a breaking change, we have to do this in the Skiko update PR, but we need to prepare it beforehand

skiko/src/commonMain/cpp/common/node/RenderNode.cpp
igordmn: I suppose, even if `layerPaint` is set, we need to apply `this->alpha`. This looks correct from general usage point of view, regardless of how we use it in Compose

skiko/src/commonMain/cpp/common/node/RenderNode.cpp
MatkovIvan: https://github.com/JetBrains/compose-multiplatform-core/pull/3195

skiko/src/commonMain/cpp/common/node/RenderNode.cpp
MatkovIvan: `layerPaint` contains alpha already. The only existing gap is when it's missing.
It marches HWUI implementation (see the link in the description) and is valid regardless of how we use it in Compose.

So:
- If `layerPaint` already carries alpha, alpha is applied twice.
- Even if it doesn’t, we would still be changing semantics from "alpha on the final composited layer" to "alpha on each draw op before they blend with each other".

igordmn: If you think it is better not to, let's add a comment to KDoc `setAlpha`, that it is

- ignored for layer itself if `layerPaint` is set
- still applied to the shadow

igordmn: > layerPaint contains alpha already. The only existing gap is when it's missing.

That is my main concern, API that works under certain conditions is worse than API that works unconditionally

MatkovIvan: Well, from this angle it's better not to expose `layerPaint` as public API and expose all its properties instead.
It was chosen to do so because it significantly reduces the public API surface here, but yes, it's a "sharp knife".

MatkovIvan: Added KDocs

igordmn: It also unconditionally affect shadow - we should write about that too

igordmn (approved):

PR #1234, opened 2026-07-10
AWT: collapse the per-backend render stack

MatkovIvan:
Every AWT backend was implemented **twice**, in two parallel inheritance trees: a `*Redrawer` (native device + present + its own frame loop) and a `*ContextHandler` (Skia `DirectContext` + surface). Each `*Redrawer` owned the matching `*ContextHandler` (same prefix), and each ran its own copy of the schedule → draw → present loop — where the frame-scheduling bugs lived.

### Before — two parallel hierarchies

```mermaid
classDiagram
    direction LR
    Redrawer <|-- AWTRedrawer
    AWTRedrawer <|-- MetalRedrawer
    AWTRedrawer <|-- SoftwareRedrawer
    AWTRedrawer <|-- AngleRedrawer
    AWTRedrawer <|-- Direct3DRedrawer
    AWTRedrawer <|-- WindowsOpenGLRedrawer
    AWTRedrawer <|-- LinuxOpenGLRedrawer
    AWTRedrawer <|-- AbstractDirectSoftwareRedrawer
    AbstractDirectSoftwareRedrawer <|-- WindowsSoftwareRedrawer
    AbstractDirectSoftwareRedrawer <|-- LinuxSoftwareRedrawer
    class AWTRedrawer { <<abstract>> }

    ContextHandler <|-- JvmContextHandler
    JvmContextHandler <|-- ContextBasedContextHandler
    JvmContextHandler <|-- ContextFreeContextHandler
    ContextBasedContextHandler <|-- MetalContextHandler
    ContextBasedContextHandler <|-- AngleContextHandler
    ContextBasedContextHandler <|-- Direct3DContextHandler
    ContextBasedContextHandler <|-- OpenGLContextHandler
    ContextFreeContextHandler <|-- SoftwareContextHandler
    ContextFreeContextHandler <|-- DirectSoftwareContextHandler

    MetalRedrawer ..> MetalContextHandler : owns
    SoftwareRedrawer ..> SoftwareContextHandler : owns
    AngleRedrawer ..> AngleContextHandler : owns
    Direct3DRedrawer ..> Direct3DContextHandler : owns
    WindowsOpenGLRedrawer ..> OpenGLContextHandler : owns
    LinuxOpenGLRedrawer ..> OpenGLContextHandler : owns
    AbstractDirectSoftwareRedrawer ..> DirectSoftwareContextHandler : owns
```

### After — one class per backend, one loop

```mermaid
classDiagram
    direction LR
    Redrawer <|.. OnScreenRedrawer
    OnScreenRedrawer ..> AWTRedrawer : drives
    AWTRedrawer <|-- MetalRedrawer
    AWTRedrawer <|-- SoftwareRedrawer
    AWTRedrawer <|-- AngleRedrawer
    AWTRedrawer <|-- Direct3DRedrawer
    AWTRedrawer <|-- AbstractOpenGLRedrawer
    AbstractOpenGLRedrawer <|-- WindowsOpenGLRedrawer
    AbstractOpenGLRedrawer <|-- LinuxOpenGLRedrawer
    AWTRedrawer <|-- AbstractDirectSoftwareRedrawer
    AbstractDirectSoftwareRedrawer <|-- WindowsSoftwareRedrawer
    AbstractDirectSoftwareRedrawer <|-- LinuxSoftwareRedrawer
    class AWTRedrawer {
        <<abstract>>
    }
    class AbstractOpenGLRedrawer {
        <<abstract>>
        the former OpenGLContextHandler
    }
```

The right-hand tree is gone. Each backend is now **one** render-context class (owns device + context + surface + present) implementing the small `AWTRedrawer` hook interface, and the single generic `OnScreenRedrawer` drives all of them.

**Removed** — the entire `org.jetbrains.skiko.context` package, 9 classes: `JvmContextHandler`, `ContextBasedContextHandler`, `ContextFreeContextHandler`, and the per-API `Metal / Angle / Direct3D / OpenGL / Software / DirectSoftware` `*ContextHandler`s. `AWTRedrawer` drops from base class to interface.

**Added** — `OnScreenRedrawer`, the one shared frame loop.

No public API change — everything here is internal.


m-sasha:
Very welcome change, but let's hold off on merging it until I'm done with my ongoing skiko changes.

skiko/src/awtMain/kotlin/org/jetbrains/skiko/redrawer/MetalRedrawer.kt
```diff
@@ -185,43 +253,8 @@ internal class MetalRedrawer(
...
-
-        fun cancel() {
-            updateDispatcher.cancel()
-            frameDispatcher.cancel()
-        }
-    }
+    // Native GPU-surface entry points; implemented in MetalRedrawerSurface.mm.
+    private external fun makeMetalContext(device: Long): Long
+    private external fun makeMetalRenderTarget(device: Long, width: Int, height: Int): Long
+    private external fun finishFrame(device: Long)
```
hub-bla: Can we decouple Ganesh from these methods? Instead of creating the Ganesh context/render target internally we could expose bindings for retrieving the device and texture. That would be a better fit and would make our rendering glue GPU backend-independent

MatkovIvan: It's just the first part of my refactor, where I tried to reduce the diff while keeping it meaningful.
I suggest keeping this change scoped and making target fixes/changes after landing this.
So, yes, but let's make it separately (it's already partially addressed in the next commits)

skiko/src/awtMain/objectiveC/macos/MetalRedrawerSurface.mm
```diff
@@ -86,8 +86,8 @@ static void finishFrame(jlong devicePtr, void (^present)(MetalDevice *device, id
 /// Presents the current drawable asynchronously — skiko's default, used for every frame outside a live
 /// resize. Called off the AppKit main thread (from the background frame loop) so present work doesn't
 /// destabilize FPS on the main thread.
-JNIEXPORT void JNICALL Java_org_jetbrains_skiko_context_MetalContextHandler_finishFrame(
-    JNIEnv *env, jobject contextHandler, jlong devicePtr)
+JNIEXPORT void JNICALL Java_org_jetbrains_skiko_redrawer_MetalRedrawer_finishFrame(
+    JNIEnv *env, jobject redrawer, jlong devicePtr)
```
hub-bla, line 90: Only the argument is changing. I guess it is obsolete then?

skiko/src/awtMain/kotlin/org/jetbrains/skiko/redrawer/OnScreenRedrawer.kt
```diff
@@ -0,0 +1,150 @@
...
+
+/**
+ * The single generic on-screen frame loop for AWT [SkiaLayer]s: it owns the coalescing frame dispatcher and
+ * drives any [AWTRedrawer] through its pacing/present hooks, so the scheduling logic exists once here rather
+ * than per backend. Each backend places its pacing by implementing [AWTRedrawer.paceBeforeFrame] and/or
+ * [AWTRedrawer.paceAfterFrame]. Metal additionally opts into a separate update ticker
+ * ([AWTRedrawer.separatesUpdateAndDraw]) so a non-throttled [needRender] can run [SkiaLayer.update] without
+ * waiting for the vsync-paced frame.
+ */
+internal class OnScreenRedrawer(
```
hub-bla, line 14: There are couple of properties that are annotated with `ExperimentalSkikoApi`, so we should OptIn them here

skiko/src/awtMain/kotlin/org/jetbrains/skiko/redrawer/OnScreenRedrawer.kt
```diff
@@ -0,0 +1,150 @@
...
+        updateRequested.set(true)
+        if (updateDispatcher != null && !throttledToVsync) {
+            updateDispatcher.scheduleFrame()
+        }
+        frameDispatcher.scheduleFrame()
+    }
+
+    override fun renderImmediately() {
+        check(!isDisposed) { "OnScreenRedrawer is disposed" }
+        layer.update(System.nanoTime())
```
hub-bla: Was normalizing all AWT backends to `System.nanoTime()` intentional? Previously `Metal`, `Direct3D`, software and every `renderImmediately` used `renderTime()`

skiko/src/awtMain/kotlin/org/jetbrains/skiko/redrawer/LinuxOpenGLRedrawer.kt
```diff
@@ -1,44 +1,91 @@
...
 import org.jetbrains.skiko.*
-import org.jetbrains.skiko.context.OpenGLContextHandler
 
+/**
+ * The single per-window Linux (GLX) OpenGL on-screen render context ([AWTRedrawer]): it owns the native
+ * GLX context lifecycle (created from the window's X11 display), the Skia [DirectContext] and on-screen GPU
+ * surface for the current frame, and the present/swap. The frame loop itself lives in the generic
+ * [OnScreenRedrawer].
+ *
+ * Pacing is per window: every GLX call (create / make-current / swap / swap-interval / destroy) runs inside
```
hub-bla: It seems like changing to this behavior is causing some problems. 
I think 

Tested on my linux machine

Previous:
https://github.com/user-attachments/assets/17dc67f4-5f66-4989-80ac-e039b6d1dbea

Current:
https://github.com/user-attachments/assets/bdcd779f-d5f0-4376-affc-ea7f2a81214b

hub-bla (changes requested):

skiko/src/awtMain/objectiveC/macos/MetalRedrawerSurface.mm
MatkovIvan: This function is a member of the class that was renamed. The second argument that is changed is just `this` reference

skiko/src/awtMain/kotlin/org/jetbrains/skiko/redrawer/LinuxOpenGLRedrawer.kt
MatkovIvan: This one requires a proper separate fix, but for now I've just reverted behavior/threading like it was on `master` to reduce the scope

skiko/src/awtMain/kotlin/org/jetbrains/skiko/redrawer/OnScreenRedrawer.kt
MatkovIvan: Reverted

skiko/src/awtMain/kotlin/org/jetbrains/skiko/redrawer/OnScreenRedrawer.kt
MatkovIvan: Added.
In general, I think it's better to opt it in module wide, but let's do it later.

hub-bla (approved):

skiko/src/awtMain/cpp/windows/direct3DContext.cc
```diff
@@ -11,7 +11,7 @@
```
igordmn, line 11: Let's deprecate `ContextHandler`, move it to `macosMain`, and add TODO to remove it

skiko/src/awtMain/kotlin/org/jetbrains/skiko/redrawer/AbstractDirectSoftwareRedrawer.kt
```diff
@@ -1,66 +1,135 @@
...
-    analytics: SkiaLayerAnalytics,
+    layer: SkiaLayer,
     private val properties: SkiaLayerProperties
-) : AWTRedrawer(layer, analytics, GraphicsApi.SOFTWARE_FAST) {
-    private val contextHandler = DirectSoftwareContextHandler(layer)
-    override val renderInfo: String get() = contextHandler.rendererInfo()
+) : AWTRedrawer(layer) {
+
+    /** Frame loop and [releaseResources] both run on the EDT here, so this only makes the exclusion explicit. */
+    private val drawLock = Any()
```
igordmn: If the redrawer single-threaded, better to remove the lock (we didn't have it). It is unnecessary and makes the code more difficult.

skiko/src/awtMain/kotlin/org/jetbrains/skiko/redrawer/SoftwareRedrawer.kt
```diff
@@ -1,63 +1,127 @@
...
+    layer: SkiaLayer,
     properties: SkiaLayerProperties
-) : AWTRedrawer(layer, analytics, GraphicsApi.SOFTWARE_FAST) {
-    init {
-        onDeviceChosen("Software")
-    }
+) : AWTRedrawer(layer) {
+
+    /** Frame loop and [releaseResources] both run on the EDT here, so this only makes the exclusion explicit. */
+    private val drawLock = Any()
```
igordmn: If the redrawer single-threaded, better to remove the lock

skiko/src/awtMain/kotlin/org/jetbrains/skiko/redrawer/AWTRedrawer.kt
```diff
@@ -1,87 +1,84 @@
...
-    private val analytics: SkiaLayerAnalytics,
-    private val graphicsApi: GraphicsApi,
-) : Redrawer {
-    private var isFirstFrameRendered = false
-
-    private val rendererAnalytics = analytics.renderer(Version.skiko, hostOs, graphicsApi)
-    private var deviceAnalytics: DeviceAnalytics? = null
-    protected var isDisposed = false
-        private set
+internal abstract class AWTRedrawer(protected val layer: SkiaLayer) : AutoCloseable {
```
igordmn: Previously, `Redrawer` was selected because it drives the redrawing. Now this is moved to `OnScreenRedrawer`. Because of that, I suggest:
- remove `Redrawer` as an interface
- use `OnScreenRedrawer` inside `SkiaLayer` directly
- rename `AWTRedrawer` to `Renderer`

P.S. We can keep the file names the same if git diff doesn't understand that it is a rename (and rename it separately)

skiko/src/awtMain/kotlin/org/jetbrains/skiko/redrawer/OnScreenRedrawer.kt
```diff
@@ -0,0 +1,126 @@
...
+/**
+ * The single generic on-screen frame loop for AWT [SkiaLayer]s: it owns the coalescing frame dispatcher and
+ * drives any [AWTRedrawer], so the scheduling logic exists once here rather than per backend. A backend wraps
+ * the frame in [AWTRedrawer.runFrame] to place its own pacing around it, and hears about each request through
+ * [AWTRedrawer.onFrameRequested]; what it does with either is its own business.
+ */
+@OptIn(ExperimentalSkikoApi::class)
+internal class OnScreenRedrawer(
+    private val layer: SkiaLayer,
+    val ctx: AWTRedrawer,
```
igordmn:
```suggestion
    private val renderer: Renderer,
```

- exposing it only for tests isn't worth it, eventually it become used somewhere else
- the assert `require(redrawer.ctx is MetalRedrawer)` seems isn't needed in that test
- the name `ctx` is vague.

igordmn (commented):
(this is the first part of the comments, still reviewing it)

skiko/src/awtMain/kotlin/org/jetbrains/skiko/redrawer/WindowsOpenGLRedrawer.kt
```diff
@@ -31,105 +34,47 @@ internal class WindowsOpenGLRedrawer(
...
-
-    companion object {
-        private val toRedraw = mutableSetOf<WindowsOpenGLRedrawer>()
-        private val toRedrawCopy = mutableSetOf<WindowsOpenGLRedrawer>()
-        private val toRedrawVisible = toRedrawCopy
-            .asSequence()
-            .filterNot(WindowsOpenGLRedrawer::isDisposed)
-            .filter { it.layer.isShowing }
-
-        private val frameDispatcher = FrameDispatcher(MainUIDispatcher) {
```
igordmn, line 95: The common `frameDispatcher` was added on purpose because of non-stable FPS in multiple windows. How does it work now?

igordmn: This issue should be caught by `SkiaLayerPerformanceTest`, we need to compare it with old code on Windows/Linux OpenGL

skiko/src/awtMain/kotlin/org/jetbrains/skiko/redrawer/AWTRedrawer.kt
```diff
@@ -1,87 +1,84 @@
...
+    open fun attachFrameHost(host: FrameHost) {}
+
+    @Volatile
+    protected var isDisposed = false
+        private set
+
+    /**
+     * Releases every native and Skia resource this backend owns. [isDisposed] is already `true` when it runs,
+     * so an in-flight frame that re-checks it under the backend's own lock stands down. Called once: the
+     * check-and-set in [close] is not atomic, and relies on disposal being confined to the EDT.
```
igordmn:
```suggestion
     * Releases every native and Skia resource this backend owns. [isDisposed] is already `true` when it runs
```
This is difficult to read and associate with something without the context. I recently noticed in another PRs and own usages that some AI uses convoluted language unless you ask it to be simple.

skiko/src/awtMain/kotlin/org/jetbrains/skiko/redrawer/AWTRedrawer.kt
```diff
@@ -1,87 +1,84 @@
...
+
+    open fun isTransparentBackgroundSupported(): Boolean = defaultIsTransparentBackgroundSupported(layer)
 
     /**
-     * Should be called when the device name is known as early, as possible.
+     * Renders and presents exactly one frame at [scope]'s size. Owns the backend's threading, drawing-surface
+     * lock, autorelease pool and size guard; [immediate] selects the synchronous-redraw variant.
+     *
+     * Throwing [org.jetbrains.skiko.RenderException] means the frame failed, and makes
+     * [SkiaLayer.inDrawScope] fall back to the next render API.
```
igordmn:
```suggestion
     * Renders and presents exactly one frame at [scope]'s size. [immediate] selects the synchronous-redraw variant.
     *
     * Throwing [org.jetbrains.skiko.RenderException] means the frame failed, and makes
     * [SkiaLayer.inDrawScope] fall back to the next render API.
```

- I don't understand what `owns` means here. `Owns` usually should be referred to something created it, not a function using it
- `lock` is implementation specific, outside it is single threaded.
- `autorelease pool` - mac os specific
- `size guard` - isn't obvious without the context

skiko/src/awtMain/kotlin/org/jetbrains/skiko/redrawer/LinuxOpenGLRedrawer.kt
```diff
@@ -1,44 +1,47 @@
...
-import org.jetbrains.skiko.context.OpenGLContextHandler
 
+/**
+ * Every GLX call runs inside [org.jetbrains.skiko.lockLinuxDrawingSurface], and the whole per-frame body runs
+ * on the EDT: a GLX context stays current on the thread that bound it, and the native `makeCurrent` here
+ * never releases it, so binding one from two threads is an error. The vblank wait cannot move off the EDT
+ * either — GLX fuses it into `glXSwapBuffers`, with no `dwmFlush` equivalent — so only one window per process
+ * waits per frame; see [swapIntervalForFrame]. A software frame limiter runs in [runFrame], because
+ * some Linuxes don't honour vsync.
+ */
```
igordmn:
```suggestion
/**
 * Every GLX call runs inside [org.jetbrains.skiko.lockLinuxDrawingSurface], and the whole per-frame body runs
 * on the EDT. 
 * 
 * A software frame limiter runs in [runFrame], because some Linuxes don't honour vsync.
 */
```

- "a GLX context stays current on the thread that bound it, and the native `makeCurrent`" - this is OpenGL basics and in the doc of `makeCurrent`
- a note about `glXSwapBuffers` belongs where it is called, here it is out of the context.
- `dwmFlush` is unrelated to Linux

skiko/src/awtMain/kotlin/org/jetbrains/skiko/redrawer/WindowsOpenGLRedrawer.kt
igordmn: I tested it on my Windows machine, it became worse.

Was:
```
Deltas [5.6, 6.5, 6.4, 5.9, 5.8, 5.8, 6.1, 5.5, 6.5, 5.5, 7.3, 6.0, 6.0, 6.3, 5.7, 5.7, 6.0, 6.0, 6.8, 6.0, 6.3, 6.6, 10.3, 6.0, 6.4, 6.2, 5.8, 7.1, 5.9, 5.8, 6.0, 5.9, 6.3, 5.4, 7.8, 10.6, 6.5, 6.8, 5.4, 5.3, 5.9, 6.0, 7.1, 6.3, 6.1, 5.8, 5.9, 6.1, 5.6, 6.2, 6.3, 6.1, 6.5, 5.8, 5.8, 5.8, 5.8, 6.0, 6.8, 6.4, 6.4, 5.6, 5.7, 6.0, 5.9, 6.5, 5.9, 7.1, 11.1, 6.4, 5.7, 5.6, 6.7, 5.5, 6.4, 6.4, 5.9, 6.1, 6.2, 6.3, 6.0, 5.7, 5.7, 6.2, 6.7, 6.3, 5.9, 5.9, 5.9, 5.6, 6.3, 5.8, 6.4, 6.9, 5.6, 6.2, 6.2, 5.9, 5.5, 5.4, 6.6, 6.6, 6.4, 5.4, 6.0, 6.4, 5.0, 6.5, 5.6, 7.1, 6.8, 5.2, 5.8, 6.4, 5.6, 5.7, 6.6, 6.9, 5.4, 6.0, 6.0, 5.9, 6.2, 5.2, 6.5, 6.4, 6.8, 5.6, 5.9, 6.6, 5.3, 5.7, 6.0, 7.9, 10.9, 6.0, 6.5, 5.2, 6.4, 5.5, 6.0, 7.2, 5.8, 5.9, 5.8, 6.3, 6.6, 4.9, 6.1, 7.8, 10.8, 6.1, 5.8, 6.2, 5.6, 5.9, 5.8, 6.9, 6.2, 5.7, 6.8, 6.2, 5.8, 5.4, 6.1, 6.5, 6.1, 5.8, 6.5, 5.7, 5.7, 5.6, 6.2, 6.4, 7.1, 5.6, 5.8, 5.8, 6.7, 5.7, 5.1, 7.8, 5.4, 6.1, 5.9, 6.0, 5.8, 5.5, 6.0, 6.3, 6.6, 6.2, 6.2, 6.0, 5.7, 6.1, 6.2, 5.7, 6.4, 6.1, 6.2, 6.3, 5.6, 6.1, 5.3, 6.3, 6.2, 6.5, 6.9, 5.7, 5.5, 5.9, 6.4, 5.5, 5.7, 6.4, 7.3, 6.6, 5.3, 5.4, 5.8, 5.9, 5.8, 5.8, 6.9, 6.3, 6.7, 5.9, 6.1, 5.3, 5.7, 5.9, 7.2, 6.0, 5.5, 6.1, 6.0, 6.4, 5.1, 6.4, 5.6, 5.8, 6.2, 7.1, 6.2, 6.3, 5.6, 6.3, 5.7, 5.6, 7.0, 6.1, 5.7, 6.1, 6.1, 5.6, 5.7, 6.5, 5.5, 6.2, 6.0, 6.2, 7.0, 5.8, 6.1, 5.7, 6.2, 5.9, 5.7, 6.4, 7.8, 5.9, 5.1, 6.1, 5.7, 6.2, 5.4, 5.9, 6.5, 6.4, 6.6, 5.8, 6.4, 5.5, 6.0, 6.6, 5.5, 5.7, 7.0, 6.0, 5.9, 6.1, 6.6, 10.4, 6.2, 6.3, 6.0, 6.4, 7.5]
Average 6.19
Standard deviation 0.82
Expected frame time: 6 ms.
[5.6, 6.5, 6.4, 5.5, 6.5, 5.5, 5.7, 5.7, 6.8, 6.6, 6.4, 5.4, 6.5, 6.8, 5.4, 5.3, 5.6, 6.5, 5.8, 5.8, 6.8, 6.4, 5.6, 5.7, 6.5, 5.7, 5.6, 6.7, 5.5, 6.4, 6.4, 5.7, 5.7, 6.7, 5.6, 6.9, 5.6, 5.5, 5.4, 6.6, 6.6, 6.4, 5.4, 6.4, 6.5, 5.6, 6.8, 6.4, 5.6, 5.7, 6.6, 6.9, 5.4, 6.5, 6.4, 6.8, 5.6, 6.6, 5.3, 5.7, 6.5, 5.5, 6.6, 5.6, 6.9, 5.7, 6.8, 5.4, 6.5, 6.5, 5.7, 5.7, 5.6, 6.4, 5.6, 5.8, 6.7, 5.7, 5.4, 5.5, 6.6, 5.7, 5.7, 6.4, 5.6, 5.3, 6.5, 6.9, 5.7, 5.5, 6.4, 5.5, 5.7, 6.4, 6.6, 5.3, 5.4, 5.8, 6.9, 6.7, 5.3, 5.7, 5.5, 6.4, 5.6, 5.8, 5.6, 5.7, 5.6, 5.7, 5.6, 5.7, 6.5, 5.5, 5.8, 5.7, 5.7, 6.4, 5.7, 5.4, 6.5, 6.4, 6.6, 6.4, 5.5, 6.6, 5.5, 5.7, 6.6, 6.4] deviate by 5%
[7.3, 7.1, 7.8, 7.1, 7.1, 5.0, 7.1, 5.2, 5.2, 5.2, 7.2, 4.9, 7.8, 7.1, 5.1, 7.8, 7.3, 7.2, 5.1, 7.1, 7.0, 7.0, 7.8, 5.1, 7.0, 7.5] deviate by 15%
[7.9] deviate by 30%
[10.3, 10.6, 11.1, 10.9, 10.8, 10.4] deviate by 50%
```

Now:
```
Deltas [7.6, 5.5, 4.4, 6.4, 5.2, 7.2, 8.3, 6.1, 3.4, 7.0, 8.2, 4.1, 5.2, 8.4, 5.7, 4.4, 7.2, 6.1, 6.7, 4.0, 7.4, 4.8, 7.0, 6.8, 5.7, 6.2, 4.9, 5.5, 5.8, 6.7, 6.8, 5.5, 7.8, 4.2, 6.2, 7.1, 6.4, 4.2, 5.4, 7.0, 8.2, 5.6, 6.4, 4.8, 5.1, 6.1, 7.2, 6.5, 7.3, 5.0, 4.6, 8.1, 5.9, 3.8, 7.3, 6.4, 5.7, 5.7, 5.9, 8.4, 5.9, 9.0, 7.2, 5.6, 4.9, 7.1, 5.7, 7.2, 8.2, 3.9, 6.0, 6.2, 6.4, 5.1, 5.6, 7.2, 6.1, 5.3, 7.2, 4.1, 6.3, 7.4, 6.6, 6.0, 5.3, 6.8, 4.2, 7.1, 6.3, 3.8, 7.8, 7.9, 5.0, 6.8, 4.1, 6.6, 7.1, 4.6, 7.6, 6.3, 4.2, 7.7, 6.4, 5.4, 4.0, 6.2, 6.3, 6.3, 7.7, 4.3, 9.2, 4.1, 7.1, 5.8, 6.1, 5.2, 7.0, 7.0, 5.5, 3.3, 6.4, 7.0, 4.4, 6.2, 5.5, 6.4, 10.1, 8.3, 7.7, 6.3, 5.2, 6.3, 5.6, 6.8, 5.2, 5.8, 6.8, 6.0, 5.3, 5.4, 7.8, 6.0, 5.7, 5.6, 8.6, 4.8, 6.5, 5.8, 5.2, 5.3, 9.0, 5.9, 4.8, 6.4, 5.1, 4.6, 7.9, 5.5, 8.5, 10.9, 4.8, 6.5, 5.2, 5.1, 6.0, 8.9, 6.2, 4.9, 5.5, 5.3, 6.3, 7.5, 5.3, 7.0, 7.2, 4.2, 8.2, 4.1, 6.0, 5.8, 5.3, 8.0, 5.3, 6.8, 4.7, 7.1, 5.1, 6.1, 5.8, 5.7, 5.4, 8.2, 7.2, 5.1, 8.2, 10.5, 3.2, 7.3, 7.3, 4.4, 6.0, 8.0, 5.9, 4.3, 6.8, 5.2, 6.6, 5.3, 9.0, 5.8, 5.2, 6.4, 4.9, 7.2, 5.7, 5.9, 6.2, 6.0, 7.8, 5.0, 5.3, 6.5, 5.1, 7.5, 4.8, 6.1, 6.0, 8.1, 6.0, 4.4, 5.3, 6.6, 5.5, 5.9, 6.5, 5.9, 6.0, 8.0, 5.3, 6.3, 4.3, 7.4, 4.9, 6.0, 6.6, 7.7, 5.6, 6.7, 5.4, 5.9, 5.1, 4.6, 6.7, 6.4, 6.6, 5.7, 7.8, 6.3, 3.0, 6.4, 5.5, 8.5, 6.3, 5.2, 8.3, 5.7, 3.3, 5.7, 6.1, 6.3, 5.8, 7.6, 6.7, 6.5, 6.9, 9.4, 5.8, 7.8, 5.4, 5.3, 4.8, 6.3, 7.6, 5.9, 9.0, 9.0, 7.3, 4.4, 6.2, 6.1, 7.3, 4.7, 6.1, 6.3, 8.1, 4.3, 5.9, 5.8, 5.4]
Average 6.18
Standard deviation 1.31
Expected frame time: 6 ms.
[5.5, 6.4, 5.7, 6.7, 6.8, 5.7, 5.5, 5.8, 6.7, 6.8, 5.5, 5.4, 7.0, 5.6, 6.4, 6.5, 6.4, 5.7, 5.7, 5.6, 5.7, 6.4, 5.6, 6.6, 5.3, 6.8, 6.8, 6.6, 6.4, 5.4, 5.8, 5.5, 6.4, 5.5, 6.4, 5.6, 6.8, 6.8, 5.4, 5.7, 5.6, 6.5, 5.3, 6.4, 5.5, 6.5, 5.5, 5.3, 5.3, 5.3, 5.3, 6.8, 5.7, 5.4, 6.8, 6.6, 5.3, 5.7, 5.3, 6.5, 5.3, 6.6, 5.5, 6.5, 5.3, 6.6, 5.6, 6.7, 5.4, 6.7, 6.4, 6.6, 5.7, 6.4, 5.5, 5.7, 5.7, 5.8, 6.7, 6.5, 6.9, 5.4, 5.3, 5.4] deviate by 5%
[7.6, 5.2, 7.2, 7.0, 5.2, 7.2, 7.4, 4.8, 7.0, 4.9, 7.8, 7.1, 4.8, 5.1, 7.2, 7.3, 5.0, 7.3, 7.2, 4.9, 7.1, 7.2, 5.1, 7.2, 5.3, 7.2, 7.4, 7.1, 7.8, 5.0, 7.1, 7.6, 7.7, 7.7, 7.1, 5.2, 7.0, 7.0, 7.0, 7.7, 5.2, 5.2, 5.3, 7.8, 4.8, 5.2, 4.8, 5.1, 4.8, 5.2, 5.1, 4.9, 7.5, 7.0, 7.2, 4.7, 7.1, 5.1, 7.2, 5.1, 7.3, 7.3, 5.2, 5.2, 4.9, 7.2, 7.8, 5.0, 5.1, 7.5, 4.8, 7.4, 4.9, 7.7, 5.1, 7.8, 5.2, 7.6, 7.8, 4.8, 7.6, 7.3, 7.3] deviate by 15%
[4.4, 8.3, 8.2, 4.1, 8.4, 4.4, 4.2, 4.2, 8.2, 4.6, 8.1, 8.4, 9.0, 8.2, 4.1, 4.2, 7.9, 4.1, 4.6, 4.2, 4.3, 4.1, 4.4, 8.3, 8.6, 9.0, 4.6, 7.9, 8.5, 8.9, 4.2, 8.2, 4.1, 8.0, 8.2, 8.2, 4.4, 8.0, 4.3, 9.0, 8.1, 4.4, 8.0, 4.3, 4.6, 8.5, 8.3, 9.0, 9.0, 4.4, 4.7, 8.1, 4.3] deviate by 30%
[3.4, 4.0, 3.8, 3.9, 3.8, 4.0, 9.2, 3.3, 10.1, 10.9, 10.5, 3.2, 3.0, 3.3, 9.4] deviate by 50%
```

0.82 -> 1.31 deviation

With empty windows.

Was:
```
Deltas [5.7, 6.3, 7.0, 5.5, 6.2, 5.7, 6.0, 6.2, 5.4, 6.3, 5.9, 6.9, 5.9, 6.0, 6.4, 5.5, 6.3, 5.5, 7.2, 6.3, 5.3, 6.3, 5.8, 6.2, 5.2, 6.4, 5.8, 7.7, 5.8, 5.4, 5.8, 6.8, 5.0, 5.7, 6.8, 6.6, 6.0, 5.4, 6.4, 5.8, 5.8, 5.7, 6.0, 6.0, 6.2, 7.0, 6.4, 5.9, 5.7, 6.2, 5.8, 5.7, 6.9, 6.1, 5.9, 5.9, 6.1, 5.5, 5.8, 6.1, 5.9, 6.1, 6.4, 7.3, 6.2, 5.2, 6.2, 6.3, 6.0, 5.8, 6.8, 6.3, 5.9, 5.9, 5.9, 5.8, 5.7, 5.7, 5.5, 6.5, 6.3, 6.3, 6.4, 5.9, 6.1, 5.7, 6.1, 5.8, 6.6, 6.1, 6.5, 5.7, 6.6, 4.9, 5.7, 6.4, 5.6, 6.5, 7.2, 5.3, 5.9, 5.9, 6.5, 5.7, 5.7, 7.2, 5.9, 6.0, 6.2, 5.5, 5.6, 6.0, 6.1, 5.5, 6.0, 7.2, 6.2, 5.9, 6.9, 5.3, 5.9, 5.5, 6.6, 7.1, 5.7, 6.4, 5.2, 5.9, 6.2, 5.4, 6.1, 6.1, 7.7, 5.5, 6.1, 5.5, 6.8, 5.8, 5.7, 6.1, 7.0, 5.4, 6.5, 5.4, 5.8, 5.7, 6.3, 5.9, 6.0, 5.8, 6.4, 6.6, 6.2, 6.1, 5.7, 6.0, 6.1, 5.4, 7.1, 6.5, 5.6, 5.9, 6.0, 5.6, 5.7, 6.4, 5.9, 5.7, 6.5, 7.2, 5.3, 7.2, 5.2, 6.6, 5.2, 6.0, 6.4, 7.0, 5.4, 5.8, 6.2, 5.7, 5.8, 5.8, 6.0, 6.4, 6.0, 7.2, 5.6, 6.5, 5.6, 5.7, 5.9, 5.7, 6.8, 6.4, 6.0, 5.9, 5.8, 6.6, 5.0, 5.9, 6.0, 6.2, 6.8, 6.3, 6.1, 5.8, 6.6, 6.0, 5.5, 6.3, 6.2, 5.9, 6.2, 5.6, 6.3, 5.8, 5.9, 5.6, 6.5, 7.3, 5.3, 5.9, 6.2, 5.9, 5.7, 6.2, 5.6, 6.7, 6.1, 6.3, 6.0, 6.0, 5.7, 5.7, 5.8, 6.0, 6.5, 6.0, 7.0, 6.0, 6.4, 5.4, 5.9, 5.9, 5.8, 6.3, 6.9, 6.2, 5.6, 5.8, 6.2, 6.0, 5.6, 6.3, 5.5, 6.5, 5.9, 7.7, 5.4, 6.1, 6.3, 5.8, 5.7, 5.9, 6.9, 5.4, 6.3, 5.7, 6.2, 6.3, 5.4, 5.9, 5.9, 6.4, 6.6, 6.2, 7.1, 4.8, 6.4, 6.0, 5.6, 5.8, 5.9, 6.6, 5.9, 6.6, 6.3, 5.8, 5.6, 5.3, 6.0, 6.5, 6.0, 6.4, 6.5, 6.3, 5.4]
Average 6.06
Standard deviation 0.50
Expected frame time: 6 ms.
[5.7, 7.0, 5.5, 5.7, 5.4, 6.9, 5.5, 5.5, 5.3, 6.4, 5.8, 5.4, 6.8, 5.7, 6.8, 6.6, 5.4, 5.8, 5.7, 5.7, 5.7, 6.9, 5.5, 5.8, 6.4, 6.8, 5.8, 5.7, 5.7, 5.5, 6.5, 6.4, 5.7, 6.6, 6.5, 5.7, 6.6, 5.7, 6.4, 5.6, 6.5, 5.3, 6.5, 5.7, 5.7, 5.5, 5.6, 5.5, 6.9, 5.3, 5.5, 6.6, 5.7, 6.4, 5.4, 5.5, 5.5, 6.8, 5.7, 5.4, 6.5, 5.4, 5.7, 6.4, 6.6, 5.7, 5.4, 6.5, 5.6, 5.6, 5.7, 6.4, 5.7, 6.5, 5.3, 6.6, 6.4, 5.4, 5.7, 5.8, 6.4, 5.6, 6.5, 5.6, 5.7, 5.7, 6.8, 6.4, 6.6, 6.8, 5.8, 6.6, 5.5, 5.6, 5.6, 6.5, 5.3, 5.7, 5.6, 6.7, 5.7, 5.7, 6.5, 6.4, 5.4, 6.9, 5.6, 5.6, 5.5, 6.5, 5.4, 5.7, 6.9, 5.4, 5.7, 5.4, 6.4, 6.6, 6.4, 5.6, 6.6, 6.6, 5.6, 5.3, 6.5, 6.4, 6.5, 5.4] deviate by 5%
[7.2, 5.2, 7.7, 5.0, 7.0, 7.3, 5.2, 4.9, 7.2, 7.2, 7.2, 7.1, 5.2, 7.7, 7.0, 7.1, 7.2, 7.2, 5.2, 5.2, 7.0, 7.2, 5.0, 7.3, 7.0, 7.7, 7.1, 4.8] deviate by 15%
[] deviate by 30%
[] deviate by 50%
```

Now:
```
Deltas [5.7, 5.2, 5.9, 6.2, 7.6, 7.6, 4.8, 6.4, 5.6, 5.5, 5.4, 6.4, 6.6, 6.3, 7.2, 5.2, 6.7, 5.9, 5.0, 6.5, 5.3, 5.4, 8.0, 6.0, 5.5, 5.5, 5.5, 6.1, 5.6, 7.2, 7.6, 4.9, 5.0, 6.2, 6.3, 5.5, 6.3, 5.2, 7.3, 6.5, 7.3, 4.8, 6.3, 6.5, 5.1, 6.0, 6.1, 7.5, 6.7, 4.3, 6.1, 5.7, 5.9, 5.4, 5.7, 6.6, 5.9, 7.5, 5.4, 6.2, 5.6, 7.3, 5.9, 5.2, 6.9, 6.7, 5.1, 6.9, 5.7, 6.6, 4.5, 5.7, 6.6, 5.8, 6.3, 7.2, 4.3, 6.7, 7.0, 4.9, 5.9, 5.1, 9.0, 5.9, 4.6, 6.5, 6.3, 4.7, 7.2, 6.1, 5.8, 5.1, 8.0, 5.0, 7.4, 6.3, 5.4, 4.9, 6.4, 5.6, 6.7, 6.4, 6.0, 5.6, 6.2, 5.6, 6.4, 5.7, 6.2, 5.9, 7.5, 6.7, 5.2, 6.5, 5.5, 4.9, 6.4, 6.5, 6.8, 5.5, 6.2, 4.9, 5.5, 5.7, 6.5, 6.6, 6.2, 7.0, 5.3, 6.2, 5.9, 6.3, 5.9, 5.2, 7.2, 6.1, 7.1, 5.6, 6.3, 4.6, 5.9, 6.9, 5.3, 6.1, 5.3, 6.9, 5.9, 6.8, 5.2, 5.9, 6.4, 5.9, 7.3, 6.2, 5.4, 6.6, 5.4, 5.6, 5.2, 6.8, 6.4, 6.0, 7.8, 4.8, 5.8, 6.9, 6.1, 5.7, 5.6, 5.1, 7.8, 5.4, 6.7, 4.7, 6.9, 6.2, 5.7, 5.6, 5.9, 6.5, 6.5, 7.0, 4.9, 5.7, 6.1, 6.4, 5.6, 6.1, 6.5, 5.7, 7.2, 6.2, 5.2, 6.5, 6.4, 5.4, 5.9, 6.0, 5.2, 7.8, 6.9, 4.9, 5.8, 6.0, 5.7, 6.3, 5.5, 7.6, 5.1, 6.6, 5.3, 5.9, 6.7, 4.7, 5.7, 6.3, 5.7, 7.8, 7.6, 4.1, 6.8, 6.0, 5.1, 5.2, 7.9, 5.9, 6.6, 5.6, 6.8, 4.8, 5.2, 6.5, 7.0, 4.6, 5.8, 6.6, 8.6, 5.2, 5.9, 6.1, 5.7, 5.2, 6.0, 5.4, 7.5, 5.1, 7.6, 5.2, 5.2, 6.0, 6.6, 6.2, 6.1, 5.6, 5.9, 7.1, 5.6, 6.3, 5.6, 6.6, 6.5, 5.7, 6.0, 7.1, 5.7, 5.7, 5.7, 5.3, 6.9, 5.2, 6.5, 5.6, 5.6, 7.3, 6.8, 6.0, 5.8, 5.1, 6.2, 5.6, 6.1, 7.1, 6.7, 6.2, 5.7, 5.2, 6.2, 5.5, 6.9, 5.7, 5.7, 6.5, 6.5, 6.9, 4.8, 5.7, 5.9, 5.8, 5.8]
Average 6.06
Standard deviation 0.81
Expected frame time: 6 ms.
[5.7, 6.4, 5.6, 5.5, 5.4, 6.4, 6.6, 6.7, 6.5, 5.4, 5.5, 5.5, 5.5, 5.6, 5.5, 6.5, 6.5, 6.7, 5.7, 5.4, 5.7, 6.6, 5.4, 5.6, 6.9, 6.7, 6.9, 5.7, 6.6, 5.7, 6.6, 6.7, 6.5, 5.4, 6.4, 5.6, 6.7, 6.4, 5.6, 5.6, 6.4, 5.7, 6.7, 6.5, 5.5, 6.4, 6.5, 6.8, 5.5, 5.5, 5.7, 6.5, 6.6, 7.0, 5.3, 5.6, 6.9, 5.3, 6.9, 6.8, 6.4, 5.4, 6.6, 5.4, 5.6, 6.8, 6.4, 6.9, 5.7, 5.6, 5.4, 6.7, 6.9, 5.7, 5.6, 6.5, 6.5, 7.0, 5.7, 6.4, 5.6, 6.5, 5.7, 6.5, 6.4, 5.4, 6.9, 5.7, 5.5, 6.6, 5.3, 6.7, 5.7, 5.7, 6.8, 6.6, 5.6, 6.8, 6.5, 6.6, 5.7, 5.4, 6.6, 5.6, 5.6, 5.6, 6.6, 6.5, 5.7, 5.7, 5.7, 5.7, 5.3, 6.9, 6.5, 5.6, 5.6, 6.8, 5.8, 5.6, 6.7, 5.7, 5.5, 6.9, 5.7, 5.7, 6.5, 6.5, 6.9, 5.7] deviate by 5%
[5.2, 7.6, 7.6, 4.8, 7.2, 5.2, 5.0, 5.3, 7.2, 7.6, 4.9, 5.0, 5.2, 7.3, 7.3, 4.8, 5.1, 7.5, 7.5, 7.3, 5.2, 5.1, 7.2, 7.0, 4.9, 5.1, 7.2, 5.1, 5.0, 7.4, 4.9, 7.5, 5.2, 4.9, 4.9, 5.2, 7.2, 7.1, 5.3, 5.2, 7.3, 5.2, 7.8, 4.8, 5.1, 7.8, 4.7, 4.9, 7.2, 5.2, 5.2, 7.8, 4.9, 7.6, 5.1, 4.7, 7.8, 7.6, 5.1, 5.2, 4.8, 5.2, 7.0, 5.2, 5.2, 7.5, 5.1, 7.6, 5.2, 5.2, 7.1, 7.1, 5.2, 7.3, 5.1, 7.1, 5.2, 4.8] deviate by 15%
[8.0, 4.3, 4.5, 4.3, 9.0, 4.6, 4.7, 8.0, 4.6, 4.1, 7.9, 4.6, 8.6] deviate by 30%
[] deviate by 50%
```

0.50 -> 0.81 deviation.

Not significant, but still big. I would restore the old way somehow. We also need to check Linux

igordmn:
As an option, we can merge the first commit as a separate PR quicker (but apply necessary fixes from the second commit, like `canvas?.close`)

skiko/src/awtMain/kotlin/org/jetbrains/skiko/redrawer/AWTRedrawer.kt
MatkovIvan: I didn't rename it to reduce the diff. I have a few more changes in the queue that follows - minimal diff approach. Let's do renaming pass separately after landing code changes

igordmn: Okay, does the suggestion sound good for you?

MatkovIvan: Not quite. I have a plan to decouple it from `SkiaLayer` entirely. The final naming in my refactor is `*RenderContext`

igordmn: > Not quite. I have a plan to decouple it from SkiaLayer entirely.

This won't be good to do in one PR. I currently see issues in the current naming, and we should fix it ASAP after merging, not keeping them in the code.

As for another changes - we should review them separately, they can be controversial.

skiko/src/awtMain/kotlin/org/jetbrains/skiko/redrawer/WindowsOpenGLRedrawer.kt
MatkovIvan: Reverted for now. Had to introduce `schedulesOwnFrames` for OpenGL case

igordmn (approved):

skiko/src/awtMain/kotlin/org/jetbrains/skiko/redrawer/OnScreenRedrawer.kt
```diff
@@ -0,0 +1,144 @@
...
+import org.jetbrains.skiko.FrameDispatcher
+import org.jetbrains.skiko.MainUIDispatcher
+import org.jetbrains.skiko.SkiaLayer
+import org.jetbrains.skiko.SkiaLayerAnalytics.DeviceAnalytics
+import org.jetbrains.skiko.LayerDrawScope
+import java.awt.Dimension
+import java.util.concurrent.atomic.AtomicBoolean
+
+@OptIn(ExperimentalSkikoApi::class)
+internal class OnScreenRedrawer(
```
m-sasha, line 14: What makes it "OnScreen"Redrawer? Are there other, non-on-screen redrawers?

skiko/src/awtMain/kotlin/org/jetbrains/skiko/redrawer/OnScreenRedrawer.kt
```diff
@@ -0,0 +1,144 @@
...
+    private val layer: SkiaLayer,
+    private val renderer: AWTRedrawer,
```
m-sasha, line 16: A "Redrawer" (`OnScreenRedrawer`) takes another "Redrawer" as an argument (in a property called `renderer`?). But `AWTRedrawer` is not actually a `Redrawer` subclass.

Confusing...

We need to rename at least one of these hierarchies (they can't both be `Redrawer`s), and explain the relationship between them.

But given the tight coupling (see comment about the `init` below), I'm not sure this separation is the best way to structure the code.

skiko/src/awtMain/kotlin/org/jetbrains/skiko/redrawer/OnScreenRedrawer.kt
```diff
@@ -0,0 +1,144 @@
...
+                        runBlocking { renderer.renderFrame(scope, immediate = true) }
+                    }
+                }
+            }
+            return
+        }
+
+        needRender(throttledToVsync = false)
+    }
+    override fun setVisible(isVisible: Boolean) = renderer.setVisible(isVisible)
```
m-sasha, line 136: Missing newline

skiko/src/awtMain/kotlin/org/jetbrains/skiko/redrawer/OnScreenRedrawer.kt
```diff
@@ -0,0 +1,144 @@
...
+            if (isDisposed) return // layer may be disposed in user code during `update`
+            withFrameAnalytics {
+                layer.inDrawScope {
+                    val scope = this
+                    if (!renderer.renderBeforeShown(scope)) {
+                        runBlocking { renderer.renderFrame(scope, immediate = true) }
+                    }
+                }
+            }
+            return
```
m-sasha, line 131: It looks like we can change this to something similar to the way it was before and avoid mostly duplicate code.
Something like:
```
if (!renderer.presentBeforeShown())
    return
```
and
```
protected open fun renderBeforeShown(): Boolean {
    renderImmediately()
    return true
}
```

There was actually a bug in the previous code; I forgot to the check the return value of `renderBeforeShown`.

skiko/src/awtMain/kotlin/org/jetbrains/skiko/redrawer/OnScreenRedrawer.kt
```diff
@@ -0,0 +1,144 @@
...
+) : Redrawer, FrameHost {
+    private val deviceAnalytics: DeviceAnalytics? get() = renderer.deviceAnalytics
+    private var isFirstFrameRendered = false
+
+    var isDisposed = false
+        private set
+
+    init {
+        renderer.attachFrameHost(this)
+    }
```
m-sasha, line 26: Such bi-directional dependency is typically a sign that these two components are not separated correctly (or don't need to be separated).

skiko/src/awtMain/kotlin/org/jetbrains/skiko/redrawer/AWTRedrawer.kt
```diff
@@ -51,58 +48,100 @@ internal abstract class AWTRedrawer(
...
+     *
+     * Returns `true` if this backend presented that frame itself; `false` presents it the ordinary way.
+     */
+    open fun renderBeforeShown(scope: LayerDrawScope): Boolean = false
+
+    /**
+     * Hands over the [FrameHost] this backend records its own frames through. Called once, before the first
+     * frame.
+     */
+    open fun attachFrameHost(host: FrameHost) {}
```
m-sasha, line 84: It looks like all implementations of this just store the `host` in a property.
Why not make that the default implementation?

skiko/src/awtMain/kotlin/org/jetbrains/skiko/redrawer/AWTRedrawer.kt
```diff
@@ -51,58 +48,100 @@ internal abstract class AWTRedrawer(
...
+    open val presentsBeforeShown: Boolean get() = true
 
     /**
      * Renders and presents a frame when the layer is already displayable but not yet showing.
      * This is needed so we have a frame ready when the window is first shown, to prevent the window background
      * flashing.
+     *
+     * Returns `true` if this backend presented that frame itself; `false` presents it the ordinary way.
+     */
+    open fun renderBeforeShown(scope: LayerDrawScope): Boolean = false
```
m-sasha, line 78: It would be cleaner if there was just one decision point on "what should happen on draw-before-visible".

skiko/src/awtMain/kotlin/org/jetbrains/skiko/redrawer/OnScreenRedrawer.kt
```diff
@@ -0,0 +1,144 @@
...
+
+    private inline fun withFrameAnalytics(body: () -> Unit) {
+        val isFirstFrame = !isFirstFrameRendered
+        isFirstFrameRendered = true
+        if (isFirstFrame) deviceAnalytics?.beforeFirstFrameRender()
+        deviceAnalytics?.beforeFrameRender()
+        body()
+        if (isFirstFrame && !isDisposed) deviceAnalytics?.afterFirstFrameRender()
+        deviceAnalytics?.afterFrameRender()
```
m-sasha, line 88: Shouldn't `isDisposed` be checked for both of (or none of) `afterFirstFrameRender` and `afterFrameRender`?

skiko/src/awtMain/kotlin/org/jetbrains/skiko/redrawer/MetalRedrawer.kt
```diff
@@ -71,11 +68,19 @@ internal class MetalRedrawer(
     /**
      * Whether this redrawer is currently driving an interactive live-resize itself (only ever true when
      * [SkikoProperties.metalSynchronousLiveResize] is enabled and this layer fills the window). Set for the
-     * duration of a drag; it quiesces the async EDT renders (frameDispatcher loop, onPlatformComponentResized) so
+     * duration of a drag; it pauses the loop's frames ([runFrame]) and the bounds sync ([syncBounds]) so
      * the synchronous AppKit-main-thread render is the only thing painting during a drag.
      */
     @Volatile
-    private var isHandlingLiveResizeNow: Boolean = false
+    final override var isHandlingLiveResizeNow: Boolean = false
```
m-sasha, line 75: `final` is redundant here

skiko/src/awtMain/kotlin/org/jetbrains/skiko/redrawer/AWTRedrawer.kt
```diff
@@ -51,58 +48,100 @@ internal abstract class AWTRedrawer(
...
+    /**
+     * Whether [SkiaLayer] presents a frame synchronously while it lays out, instead of scheduling one.
+     */
+    open val presentsOnLayout: Boolean get() = false
+
+    /**
+     * Whether the platform is driving the current resize. While `true` the loop leaves AWT resize events alone,
+     * since the platform reports the size and presents the frames itself.
+     */
+    open val isHandlingLiveResizeNow: Boolean get() = false
```
m-sasha, line 117: This shouldn't be in `AWTRedrawer`.

The fact that both metal and d3d redrawers use this flag is an implementation detail.

skiko/src/awtMain/kotlin/org/jetbrains/skiko/SkiaLayer.awt.kt
```diff
@@ -431,9 +431,9 @@ actual open class SkiaLayer internal constructor(
         // scaled in the other direction from the window size), but this seems to
         // happen less often.
         //
-        // Calling redraw during layout might break software renderers,
-        // so apply this fix only for the Direct3D case.
-        if (isShowing && (redrawer as? Direct3DRedrawer)?.isHandlingLiveResizeNow == false) {
+        // Calling redraw during layout might break software renderers, so only backends that ask for it
+        // present here.
+        if (isShowing && (redrawer as? OnScreenRedrawer)?.presentsOnLayout == true) {
```
m-sasha, line 436: Can we just remove this? Does it still improve the situation in D3D after the smooth-resize changes?

skiko/src/awtMain/kotlin/org/jetbrains/skiko/redrawer/AWTRedrawer.kt
```diff
@@ -51,58 +48,100 @@ internal abstract class AWTRedrawer(
...
+    /**
+     * `true` suppresses the loop's per-window frame dispatcher; the backend then drives every frame itself
+     * through its [FrameHost].
+     */
+    // TODO: remove along with the cross-window batch, once one frame clock serves every window on a display.
+    open val schedulesOwnFrames: Boolean get() = false
+
+    open fun setVisible(isVisible: Boolean) {}
+
+    open fun syncBounds() {}
```
m-sasha, line 106: I liked `syncBoundsFromPlatformComponent` - it made it clearer what we're syncing the bounds with.

skiko/src/awtMain/kotlin/org/jetbrains/skiko/redrawer/AWTRedrawer.kt
igordmn: In my understanding this property belongs to a class that drives the frames. I.e. we should move it somehow to `OnScreenRedrawer`, and `AWTRedrawer` should provide callbacks to subscribe to resize (or move these callbacks out of renderer too, we can do that per-OS)

skiko/src/awtMain/kotlin/org/jetbrains/skiko/redrawer/OnScreenRedrawer.kt
igordmn: This is because extraction of driving frames is only partial at the moment.

@MatkovIvan, I suggest to finish extraction soon. I.e. extract live-resize-driving and extract global-window-driving in OpenGL. This way `FrameHost` can be removed.

skiko/src/awtMain/kotlin/org/jetbrains/skiko/redrawer/OnScreenRedrawer.kt
igordmn: I suppose there will be `OffScreenRedrawer` that draws into a texture. It is needed for benchmarks

skiko/src/awtMain/kotlin/org/jetbrains/skiko/redrawer/OnScreenRedrawer.kt
igordmn: The name is indeed confusing and should be fixed soon. I suggested names https://github.com/JetBrains/skiko/pull/1234#discussion_r3730089832

skiko/src/awtMain/kotlin/org/jetbrains/skiko/redrawer/AWTRedrawer.kt
m-sasha: If it's in a generic, graphics-api-independent class/interface, it can't be called `isHandlingLiveResizeNow`, because "live-resize" is a specific behavior. It has to have a more generic name and meaning.

skiko/src/awtMain/kotlin/org/jetbrains/skiko/redrawer/OnScreenRedrawer.kt
m-sasha: Until we have such a class, I wouldn't call this `OnScreenRedrawer`. This class looks more to be like `AWTRedrawer` (and the current `AWTRedrawer` should be something else).

MatkovIvan: > Are there other, non-on-screen redrawers?

not yet. Renamed into `FrameDriver` in #1273

skiko/src/awtMain/kotlin/org/jetbrains/skiko/redrawer/OnScreenRedrawer.kt
MatkovIvan: Renamed into `AwtRenderer` in #1273

skiko/src/awtMain/kotlin/org/jetbrains/skiko/redrawer/OnScreenRedrawer.kt
MatkovIvan: Fixed in #1273

skiko/src/awtMain/kotlin/org/jetbrains/skiko/redrawer/OnScreenRedrawer.kt
MatkovIvan: Added `BeforeShownFrame` enum in #1273

skiko/src/awtMain/kotlin/org/jetbrains/skiko/redrawer/OnScreenRedrawer.kt
MatkovIvan: `this` here is just an interface reference, not the bi-directional link. Replaced by `LiveResizeListener` in #1273

skiko/src/awtMain/kotlin/org/jetbrains/skiko/redrawer/AWTRedrawer.kt
MatkovIvan: Fixed in #1273

skiko/src/awtMain/kotlin/org/jetbrains/skiko/redrawer/AWTRedrawer.kt
MatkovIvan: Added `BeforeShownFrame` enum in #1273

skiko/src/awtMain/kotlin/org/jetbrains/skiko/redrawer/OnScreenRedrawer.kt
MatkovIvan: It wasn't changed here - https://github.com/JetBrains/skiko/blob/v0.151.0/skiko/src/awtMain/kotlin/org/jetbrains/skiko/redrawer/AWTRedrawer.kt#L75-L78

cc @igordmn who has more context of analytics requirements here

skiko/src/awtMain/kotlin/org/jetbrains/skiko/redrawer/MetalRedrawer.kt
MatkovIvan: Fixed in #1273

skiko/src/awtMain/kotlin/org/jetbrains/skiko/SkiaLayer.awt.kt
MatkovIvan: Not sure about it. Out of the scope of this PR

skiko/src/awtMain/kotlin/org/jetbrains/skiko/redrawer/AWTRedrawer.kt
MatkovIvan: Restored in #1273

skiko/src/awtMain/kotlin/org/jetbrains/skiko/redrawer/OnScreenRedrawer.kt
igordmn: > who has more context of analytics requirements here

The main purpose of this is to measure native failures. And this means `isDisposed` shouldn't be checked here in the first place, as it captures non-failure `dispose`'s too. Fell free to remove or just keep it.

skiko/src/awtMain/kotlin/org/jetbrains/skiko/redrawer/OnScreenRedrawer.kt
igordmn: > this here is just an interface reference, not the bi-directional link. 

The issue was in semantics, a child renderer explicitly requests some actions of its owner (host), and owner requests actions from this renderer.

In the new code, it is different - a child renderer doesn't request something from the owner, it reports its internal state, and now the owner requests reporting it.

PR #1237, opened 2026-07-17
Eliminate background flash on first window show

m-sasha:
Fix by presenting the first frame when the window is displayable but not yet visible.

Fixes https://youtrack.jetbrains.com/issue/CMP-10484/Metal-Eliminate-window-background-flash-on-first-show
Fixes https://youtrack.jetbrains.com/issue/CMP-10485/Direct3D-Eliminate-window-background-flash-on-first-show

### DIFF
--- a/skiko/src/awtMain/kotlin/org/jetbrains/skiko/SkiaLayer.awt.kt
+++ b/skiko/src/awtMain/kotlin/org/jetbrains/skiko/SkiaLayer.awt.kt
@@ -7,8 +7,8 @@ import kotlinx.coroutines.launch
 import org.jetbrains.skia.*
 import org.jetbrains.skia.Canvas
 import org.jetbrains.skiko.internal.fastForEach
-import org.jetbrains.skiko.redrawer.AWTRedrawer
 import org.jetbrains.skiko.redrawer.Direct3DRedrawer
+import org.jetbrains.skiko.redrawer.Redrawer
 import org.jetbrains.skiko.redrawer.RedrawerManager
 import java.awt.*
 import java.awt.Color
@@ -32,7 +32,6 @@ import kotlin.math.floor
 actual open class SkiaLayer internal constructor(
     accessibleContextProvider: ((Component) -> AccessibleContext)? = null,
     val properties: SkiaLayerProperties,
-    /** Must produce [AWTRedrawer]s */
     private val renderFactory: RenderFactory = RenderFactory.Default,
     private val analytics: SkiaLayerAnalytics = SkiaLayerAnalytics.Empty,
     actual val pixelGeometry: PixelGeometry = PixelGeometry.UNKNOWN,
@@ -354,20 +353,20 @@ actual open class SkiaLayer internal constructor(
     @Volatile
     private var isDisposed = false
 
-    private val redrawerManager = RedrawerManager<AWTRedrawer>(
+    private val redrawerManager = RedrawerManager<Redrawer>(
         defaultRenderApi = properties.renderApi,
         redrawerFactory = { renderApi, oldRedrawer ->
             oldRedrawer?.dispose()
             renderFactory.createRedrawer(this, renderApi, analytics, properties).also {
                 it.syncBoundsFromPlatformComponent()
-            } as AWTRedrawer
+            }
         },
         onRenderApiChanged = {
             notifyChange(PropertyKind.Renderer)
         }
     )
 
-    internal val redrawer: AWTRedrawer? by redrawerManager::redrawer
+    internal val redrawer: Redrawer? by redrawerManager::redrawer
 
     actual var renderApi: GraphicsApi by redrawerManager::renderApi
 
--- a/skiko/src/awtMain/kotlin/org/jetbrains/skiko/SkiaLayer.awt.kt
+++ b/skiko/src/awtMain/kotlin/org/jetbrains/skiko/SkiaLayer.awt.kt
@@ -7,8 +7,8 @@ import kotlinx.coroutines.launch
 import org.jetbrains.skia.*
 import org.jetbrains.skia.Canvas
 import org.jetbrains.skiko.internal.fastForEach
+import org.jetbrains.skiko.redrawer.AWTRedrawer
 import org.jetbrains.skiko.redrawer.Direct3DRedrawer
-import org.jetbrains.skiko.redrawer.Redrawer
 import org.jetbrains.skiko.redrawer.RedrawerManager
 import java.awt.*
 import java.awt.Color
@@ -32,14 +32,16 @@ import kotlin.math.floor
 actual open class SkiaLayer internal constructor(
     accessibleContextProvider: ((Component) -> AccessibleContext)? = null,
     val properties: SkiaLayerProperties,
+    /** Must produce [AWTRedrawer]s */
     private val renderFactory: RenderFactory = RenderFactory.Default,
     private val analytics: SkiaLayerAnalytics = SkiaLayerAnalytics.Empty,
     actual val pixelGeometry: PixelGeometry = PixelGeometry.UNKNOWN,
     /**
-     * Whether this layer fills its entire host window. When true, platform-specific window-level
-     * optimizations may be used (e.g., on macOS, driving the interactive live-resize redraw from the
-     * window). Set to false when the layer is embedded as a Swing component somewhere in the window's
-     * hierarchy rather than covering the whole window, so those window-level paths are disabled.
+     * Whether this layer fills its entire host window.
+     *
+     * When true, platform-specific window-level optimizations may be used (e.g., on macOS, driving the interactive
+     * live-resize redraw from the window). Set to false when the layer is embedded as a Swing component somewhere in
+     * the window's hierarchy rather than covering the whole window, so those window-level paths are disabled.
      */
     internal val fillsWindow: Boolean = false,
 ) : JComponent(), Accessible {
@@ -124,7 +126,7 @@ actual open class SkiaLayer internal constructor(
                 @Suppress("DEPRECATION")
                 super.reshape(x, y, width, height)
 
-                redrawer?.onPlatformComponentResized()
+                redrawer?.onLayerComponentResized()
             }
 
             override fun getInputMethodRequests(): InputMethodRequests? {
@@ -352,20 +354,20 @@ actual open class SkiaLayer internal constructor(
     @Volatile
     private var isDisposed = false
 
-    private val redrawerManager = RedrawerManager<Redrawer>(
+    private val redrawerManager = RedrawerManager<AWTRedrawer>(
         defaultRenderApi = properties.renderApi,
         redrawerFactory = { renderApi, oldRedrawer ->
             oldRedrawer?.dispose()
             renderFactory.createRedrawer(this, renderApi, analytics, properties).also {
                 it.syncBoundsFromPlatformComponent()
-            }
+            } as AWTRedrawer
         },
         onRenderApiChanged = {
             notifyChange(PropertyKind.Renderer)
         }
     )
 
-    internal val redrawer: Redrawer? by redrawerManager::redrawer
+    internal val redrawer: AWTRedrawer? by redrawerManager::redrawer
 
     actual var renderApi: GraphicsApi by redrawerManager::renderApi
 
--- a/skiko/src/awtMain/kotlin/org/jetbrains/skiko/SkiaLayer.awt.kt
+++ b/skiko/src/awtMain/kotlin/org/jetbrains/skiko/SkiaLayer.awt.kt
@@ -9,10 +9,15 @@ import org.jetbrains.skiko.internal.fastForEach
 import org.jetbrains.skiko.redrawer.Direct3DRedrawer
 import org.jetbrains.skiko.redrawer.Redrawer
 import org.jetbrains.skiko.redrawer.RedrawerManager
+import org.jetbrains.skiko.redrawer.renderTime
+import org.jetbrains.skiko.swing.SoftwareSwingPainter
+import org.jetbrains.skiko.swing.SwingLayerProperties
 import java.awt.Color
 import java.awt.Component
 import java.awt.Dimension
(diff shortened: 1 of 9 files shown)


skiko/src/awtTest/kotlin/org/jetbrains/skiko/SkiaLayerTest.kt
```diff
@@ -1120,6 +1126,115 @@ class SkiaLayerTest {
         }
     }
 
+    @OptIn(ExperimentalAtomicApi::class)
+    @Test
```
igordmn:
```suggestion
    @Test(timeout = 120000)
```

skiko/src/awtMain/objectiveC/macos/MetalContextHandler.mm
```diff
@@ -148,5 +148,26 @@ JNIEXPORT void JNICALL Java_org_jetbrains_skiko_context_MetalContextHandler_fini
     });
 }
 
+/// Presents the current drawable for the warm-up frame rendered just before the window is first shown, so
+/// the window's first on-screen frame draws our content instead of briefly flashing the Canvas's background color.
+///
+/// A plain asynchronous present is enough: once a warm-up frame has been presented, the window server
+/// composites the window's first on-screen frame only after that drawable's GPU work completes, so the
+/// background is never shown.
+JNIEXPORT void JNICALL Java_org_jetbrains_skiko_context_MetalContextHandler_finishFrameBeforeShown(
```
igordmn: Looks identical to `finishFrameInLiveResize`, and the difference is only in when it is used, which usually shouldn't be a factor for keeping it separate.

Let's use a single function and rename it to something like `finishFrameAndWait` (or `finishFrameAndPresent`, not sure)

igordmn: The comments - it is better to move them to the calling side and reduce, they look out of place for me here

skiko/src/awtMain/kotlin/org/jetbrains/skiko/redrawer/MetalRedrawer.kt
```diff
@@ -143,6 +143,24 @@ internal class MetalRedrawer(
...
+        checkDisposed()
+        update()
+        inDrawScope {
+            if (!isDisposed) { // Redrawer may be disposed in user code, during `update`
+                performDraw(finishFrame = false)
+            }
+        }
+        synchronized(drawLock) {
+            if (!isDisposed) {
+                contextHandler.finishFrameBeforeShown()
```
igordmn: This needs to be wrapped into `autoreleasepool {`, because it is called from the AWT event loop, not the AppKit event  loop.

I suggest to extract a function to reuse:
```
    private fun performNativeDrawAction(action: () -> Unit) = synchronized(drawLock) {
        if (!isDisposed) {
            autoreleasepool {
                action()
            }
        }
    }
```

skiko/src/awtMain/kotlin/org/jetbrains/skiko/redrawer/AWTRedrawer.kt
```diff
@@ -83,5 +83,33 @@ internal abstract class AWTRedrawer(
...
+     */
+    protected open val supportsRenderingBeforeShown: Boolean get() = false
+
+    /**
+     * Renders and presents a frame while the layer is displayable but not showing.
+     *
+     * Only called when [supportsRenderingBeforeShown] is `true`. Implementations must present the frame in a
+     * way that the window's first on-screen composite reflects it rather than the background.
+     */
+    protected open fun renderBeforeShown() {
```
igordmn: We also perform a force render inside Compose. How is it coordinated with this change? Do you plan to remove it, call conditionally, or just keep the double call?

skiko/src/awtMain/kotlin/org/jetbrains/skiko/redrawer/MetalRedrawer.kt
```diff
@@ -143,6 +143,24 @@ internal class MetalRedrawer(
         }
     }
 
+    override val supportsRenderingBeforeShown: Boolean
+        get() = true
+
+    override fun renderBeforeShown() {
```
igordmn:
```suggestion
    override fun renderImmediatellyAndWait() {
```

This is easier for me to read, because I see the difference between different render variations.

On the other hand mentioning the use case in the name doesn't tell much and it is difficult to implement the other redrawers based on it. I see it will be more simpler if we narrow the semantics.

In the future, this also helps combining it with `renderImmediatelly`, `drawWhileInLiveResize`, the draw in the scheduler (in the end they are doing the same thing)

skiko/src/awtMain/objectiveC/macos/MetalContextHandler.mm
m-sasha: Right. It was different at first and only became the same in the last refactoring.

skiko/src/awtMain/kotlin/org/jetbrains/skiko/redrawer/AWTRedrawer.kt
m-sasha: I plan to remove the workaround from Compose.

skiko/src/awtMain/kotlin/org/jetbrains/skiko/redrawer/MetalRedrawer.kt
m-sasha: This is meant as a specific use-case function, not a generic one. The difference is in how it evolves in the future.

A generic function does a fixed thing. When what needs to be done changes, you use a different function. The generic function doesn't change.

A use-case function does what needs to be done for its use case. When what needs to be done changes, the function changes.

skiko/src/awtMain/kotlin/org/jetbrains/skiko/redrawer/AWTRedrawer.kt
igordmn: But it should be removed when we implement all the renderers, right?

skiko/src/awtMain/kotlin/org/jetbrains/skiko/redrawer/MetalRedrawer.kt
igordmn: I understand the difference, and argue that the use case function doesn't suit here.

The caller `onPlatformComponentResized` already knows what it needs to do - render a buffer and wait until it is rendered. There is no need to provide ability to customize this semantics, and we don't know yet what can be customized. If we would need that, we can convert it to an use-case function or make a broader semantics later. Until then I see this as preliminary abstraction (from the caller perspective it is more abstract). It complicates maintenance and readability.

But it is minor. If you don't agree renaming it, I suggest another fix - convert it to an event-based function `fun beforeShown() = needRender(throttledToVsync = false)` and remove `supportsRenderingBeforeShown`

skiko/src/awtMain/kotlin/org/jetbrains/skiko/redrawer/AWTRedrawer.kt
m-sasha: I'm not sure how Metal-specific this PR is, really. It's possible that `supportsRenderingBeforeShown` is not needed and we can, as a default, call `renderImmediately` (which is what we do from Compose anyway). If this turns out to be true, then we're just moving the current workaround from Compose into Skiko while allowing some renderers to do it even better, by writing their own renderer-specific code.

I will know better when I implement it for Direct3D.

skiko/src/awtMain/kotlin/org/jetbrains/skiko/redrawer/MetalRedrawer.kt
m-sasha: > I suggest another fix - convert it to an event-based function fun beforeShown() = needRender(throttledToVsync = false) and remove supportsRenderingBeforeShown

Then we'll be duplicating the logic (the `needRender(throttledToVsync = false)` call).

We could also remove both `supportsRenderingBeforeShown` and `renderBeforeShown` from `AWTRedrawer`, but then the subclasses would need to duplicate the 
```
!layer.isShowing && layer.isDisplayable && layer.width > 0 && layer.height > 0
```
check.

I think the difference between all these permutations of the same code are insignificant. I'll have a better idea how to arrange this code when there's at least another implementation, so I suggest we revisit it in the Direct3D PR.

igordmn: > Then we'll be duplicating the logic (the needRender(throttledToVsync = false) call).

It would be a duplication if it is an imperative based function. If we create an event-based function (instead of my main suggestion `renderImmediatellyAndWait`), then it is not a duplication, it is a default implementation with ability to define a different behavior

igordmn: > I think the difference between all these permutations of the same code are insignificant.

Yes, thus, choose whatever is better now from your perspective if you don't agree with me.

igordmn (approved):

skiko/src/awtMain/kotlin/org/jetbrains/skiko/SkiaLayer.awt.kt
```diff
@@ -352,20 +354,20 @@ actual open class SkiaLayer internal constructor(
...
 
-    private val redrawerManager = RedrawerManager<Redrawer>(
+    private val redrawerManager = RedrawerManager<AWTRedrawer>(
         defaultRenderApi = properties.renderApi,
         redrawerFactory = { renderApi, oldRedrawer ->
             oldRedrawer?.dispose()
             renderFactory.createRedrawer(this, renderApi, analytics, properties).also {
                 it.syncBoundsFromPlatformComponent()
-            }
+            } as AWTRedrawer
```
igordmn: It is better to move the function `onLayerComponentResized` to the interface rather than doing casting. The only purpose of `Redrawer` is to be used in `SkiaLayer`

igordmn (commented):
Please coordinate with @MatkovIvan what should be merged first and what should be adapted:
https://github.com/JetBrains/skiko/pull/1234

skiko/src/awtMain/kotlin/org/jetbrains/skiko/SkiaLayer.awt.kt
m-sasha: `onLayerComponentResized` was in `Redrawer` before the latest changes. I moved it into `AWTRedrawer` on purpose because it's not needed and not being called in `Redrawer`, and it's better to not have unneeded and unused API.

The cast can be removed if we make `Redrawer` an expect interface, but I didn't want to make that change since I know Ivan is making changes about that anyway.

igordmn: > it's not needed and not being called in Redrawer

If you use it in `SkiaLayer`, it is needed and called.

There is indeed an issue that it is created in the other compile targets that don't call this method, but it should not stop you adding it here. There is no contract to call it.

Generally, I would not make `Redrawer` common, exactly because there is no need, and we mistaken its purpose - it is not common, it exists for platform-specific single class `SkiaLayer`

skiko/src/awtMain/kotlin/org/jetbrains/skiko/redrawer/AWTRedrawer.kt
igordmn: Please, before merging this, create a PR in Compose to remove the render in Compose, to be sure that you can implement it.

skiko/src/awtMain/kotlin/org/jetbrains/skiko/SkiaLayer.awt.kt
m-sasha: But right now, `Redrawer` is in common, and `AWTRedrawer` is the type specific to the source-set. If `Redrawer` wasn't in common then there would not be a problem for me to put `onLayerComponentResized` there.

I don't want to pile on a mistake on top of an existing mistake. I thought that's what you preferred.

> If you use it in SkiaLayer, it is needed and called.

What we talk about `SkiaLayer` in common, we necessarily refer to the common `SkiaLayer`, which doesn't need or call `onLayerComponentResized`.

m-sasha: Never mind. Apparently we use non-`AWTRedrawer`s in AWT tests, so I can't cast it.

igordmn: I see no big issues adding methods even to common `Redrawer`. If it is needed by one client - it is not pretty, but justified, and it is not the issue that should be considered or fixed in your PR.

On the other hand, the approach with casting and using `AWTRedrawer` I consider wrong because it leads of overusing `AWTRedrawer` inside `SkiaLayer`, and dropping `Redrawer` that is designed to be the main interface for `SkiaLayer`.

If we don't use what was designed to be used, we should deprecate it and remove extending from it. But actually this will be done in the Ivan PR. In your PR I would just stick to the current semantics if you merge it before Ivan.

igordmn: Please, also revert the changes in SkiaLayer (`private val redrawerManager = RedrawerManager<AWTRedrawer>(` and others)

m-sasha: Done

igordmn (approved):

skiko/src/awtMain/kotlin/org/jetbrains/skiko/redrawer/AWTRedrawer.kt
m-sasha: Not a PR because skiko needs to be published first, but here's the branch:
https://github.com/JetBrains/compose-multiplatform-core/tree/m-sasha/integrate-skiko-fixes

PR #1242, opened 2026-07-23
Replace emcc compiler with clang++ for wasm target

nopjar:
This PR replaces the emcc compiler with the clang++ compiler.
Target of this PR is to reduce size of generated files by not including all provided helper functions of emcc anymore.

Needs https://github.com/JetBrains/skia/pull/28


skiko/src/webMain/resources/pre-setup.mjs
```diff
@@ -66,8 +246,49 @@ export const awaitSkiko = awaitSkikoCore.then(async (module) => {
...
+                return 0;
+            };
+        }
+        if (prop === 'makeContextCurrent') {
+            return (contextId) => {
+                // Delegate to Emscripten's makeContextCurrent which sets
+                // GL.currentContext, GLctx, etc.
+                var result = _emscriptenGL.makeContextCurrent(contextId);
+                if (result) {
+                    console.log("Setting WebGL context:", contextId);
```
Schahen: Please, don't forget to remove only debug logging

skiko/src/webMain/resources/pre-setup.mjs
```diff
@@ -66,8 +246,49 @@ export const awaitSkiko = awaitSkikoCore.then(async (module) => {
 });
 
 export const GL = new Proxy({}, {
-    get(object, propName) {
-        return skikoGl[propName];
+    get(target, prop) {
+        if (prop === 'createContext') {
+            return (canvas, attr) => {
+                const webGLCtx = canvas.getContext('webgl2', attr) || canvas.getContext('webgl', attr);
```
Schahen: If memory serves me well, prior to this PR we were creating WebGL2 context, why are you trying to fallback?

skiko/buildSrc/src/main/kotlin/properties.kt
```diff
@@ -51,11 +51,11 @@ fun compilerForTarget(os: OS, arch: Arch): String =
         OS.Android -> "clang++"
         OS.Windows -> "clang-cl.exe"
         OS.MacOS, OS.IOS, OS.TVOS -> "clang++"
-        OS.Wasm -> if (Os.isFamily(Os.FAMILY_WINDOWS)) "emcc.bat" else "emcc"
+        OS.Wasm -> project.findProperty("wasi.sdk")?.toString()?.let { "$it/bin/clang++" } ?: "/opt/wasi-sdk-33.0-arm64-macos/bin/clang++"
```
Schahen: Do we have other places in code where we are configuring something via property and then resolving to subpath? This is something we need to investigate and design properly.

Absolute paths, however, absolutely for sure do not belong here. 
Even if we'll won't manage to setup everything automatically, the actual path to wasi-sdk should be configure on the host system,

skiko/src/webMain/resources/emscripten-compat.js
```diff
@@ -0,0 +1,211 @@
+// Emscripten build-time utility stubs for browser compatibility.
+// These functions are used by libwebgl.preprocessed.js and libwebgl2.preprocessed.js
+// at the tail end of each file. They are normally provided by Emscripten's Node.js
```
Schahen, line 3: It looks like if we are supporting only WebGl2, we don't need this

skiko/build.gradle.kts
```diff
@@ -164,16 +164,18 @@ val coreDependencies: SkikoDependencyScope.() -> Unit = {
...
-                "-s", "AUTOLOAD_DYLIBS=0",
-                "-l", "GL",
-                "-s", "MAX_WEBGL_VERSION=2",
-                "-s", "MIN_WEBGL_VERSION=2",
-                "-s", "MODULARIZE=1",
-                "-s", "EXPORT_ES6=1",
-                "-s", "EXPORT_NAME=loadSkikoWASM",
-                "-s", "EXPORTED_RUNTIME_METHODS=\"[GL, wasmExports, loadDynamicLibrary, LDSO, HEAPU8]\"",
-                "--bind",
+                "--sysroot=${findProperty("wasi.sdk")?.toString() ?: "/opt/wasi-sdk-33.0-arm64-macos"}/share/wasi-sysroot",
```
Schahen: let's not have absolute path in config

skiko/src/commonMain/cpp/common/FontMgrDefaultFactory.cc
```diff
@@ -9,7 +9,7 @@
 #include "ports/SkTypeface_win.h"
 #endif
 
-#if (defined(SK_BUILD_FOR_UNIX) || defined(SK_BUILD_FOR_LINUX)) && !defined(SKIKO_WASM)
+#if defined(SK_BUILD_FOR_UNIX) || defined(SK_BUILD_FOR_LINUX)
```
Schahen: This is wrong this defined stood there for a reason - otherwise system-related definitions will leak to skiko build

skiko/buildSrc/src/main/kotlin/tasks/configuration/WasmTasksConfiguration.kt
```diff
@@ -215,6 +248,31 @@ fun SkikoProjectContext.declareWasmTasks() {
             println("Wasm and JS at: ${archiveFile.get().asFile.absolutePath}")
         }
     }
+
+    val optimizeWasm by project.tasks.registering(Exec::class) {
+        dependsOn(linkWasm)
+        val wasmFileProvider = linkWasm.flatMap { it.outDir.file(it.libOutputFileName) }
+
+        executable = "wasm-opt"
```
Schahen: The path to wasm-opt should be configurable

skiko/import-generator/src/jvmMain/kotlin/org/jetbrains/skiko/ImportGeneratorExtension.kt
```diff
@@ -50,4 +54,29 @@ internal class ImportGeneratorExtension(
             }
         }
     }
+
+    private fun inlineLocalImports(content: String, baseDir: File): String {
+        val namedImportRegex = Regex("""import\s+\{([^}]+)}\s+from\s+"\.\/([^"]+)";""")
```
Schahen, line 59: Why we need this? Which exactly files are processed?

skiko/build.gradle.kts
```diff
@@ -164,16 +164,18 @@ val coreDependencies: SkikoDependencyScope.() -> Unit = {
...
+                "-lsetjmp",
```
Schahen, line 166: it would be nice to have comments for each flag actually used

skiko/buildSrc/src/main/kotlin/tasks/configuration/WasmTasksConfiguration.kt
```diff
@@ -215,6 +248,31 @@ fun SkikoProjectContext.declareWasmTasks() {
...
+        argumentProviders.add(CommandLineArgumentProvider {
+            val wasmFile = wasmFileProvider.get().asFile
+            listOf("-Oz", "--strip-debug", "--converge", "--strip-producers", wasmFile.absolutePath, "-o", wasmFile.absolutePath + ".opt")
+        })
+
+        doLast {
+            val wasmFile = wasmFileProvider.get().asFile
+            val optimizedFile = File(wasmFile.absolutePath + ".opt")
+            if (optimizedFile.exists()) {
+                wasmFile.delete()
```
Schahen: it's better not to delete original file but rather filter it out so it won't end in the final build.
if the goal is to rename the optimized file we can achieve it similarly - when building the artifact - but ideally it's better the preoptimized file be named diffently.

skiko/src/webMain/resources/emscripten-compat.js
nopjar: What exactly do you mean? This file is needed to provide functionality that is normally provided by emscripten. The mentioned filed need those functionalities.

If you mean `libwebgl.preprocessed.js`: We still need that one, even if we only support webgl2, as it serves as a base layer as `libwebgl2.preprocessed.js` does not contain everything needed.

skiko/import-generator/src/jvmMain/kotlin/org/jetbrains/skiko/ImportGeneratorExtension.kt
nopjar: This inlines any import statement in the `pre-setup.mjs`. It is needed to include the emscripten preprocessed files into the resulting `skiko.mjs`. But the regex should consider that imports can be commented out I guess? That makes it of course a bit more difficult.

I would adapt the regex to require the `import` being at the beginning of the line. But that of course doesn't exclude block comments..

skiko/buildSrc/src/main/kotlin/tasks/configuration/WasmTasksConfiguration.kt
nopjar: Currently the gradle pipeline is like this: compile wasm -> link wasm -> optimize wasm

I don't know if we can use the optimization tool without having the output of the link wasm already written on disk.

And the goal is not really to rename, but rather to replace the original wasm with the optimized wasm.

But I am open for ideas! :)

skiko/src/webMain/resources/pre-setup.mjs
nopjar: I didn't know that we do not support webgl at all. I removed the fallback :)

skiko/buildSrc/src/main/kotlin/properties.kt
nopjar: I've removed the absolute path.

But I couldn't find anywhere else that uses that approach. Should we just pass the compiler directly instead of going through wasi-sdk?

skiko/buildSrc/src/main/kotlin/tasks/configuration/WasmTasksConfiguration.kt
nopjar: Should we make it configurable via properties of the project or do you wan't it another way?

skiko/buildSrc/src/main/kotlin/tasks/configuration/WasmTasksConfiguration.kt
Schahen: The actual goal is not to replace original wasm artefact  but rather be sure that to the final build contains skiko.wasm 

So, idea is to
1) compile to something like skiko.unoptimized.wasm
2) optimize it in whatever build pipeline for it we have and name that artifact skiko.wasm
3) make sure the it end up in final artifact (and skiko.optimized does not)

nopjar:
The current failure in the CI is because macos headers are leaking from skia as `__EMSCRIPTEN__` is not defined anymore. To fix this, https://github.com/JetBrains/skia/pull/28 needs to be merged.

EDIT: This is fixed with a temporary commit that picks skia from my private fork

skiko/import-generator/src/jvmMain/kotlin/org/jetbrains/skiko/ImportGeneratorExtension.kt
nopjar: Replaced it with a new gradle task

skiko/buildSrc/src/main/kotlin/tasks/configuration/WasmTasksConfiguration.kt
nopjar: done in a different PR

PR #1226, opened 2026-07-05
Make window resizing smooth on macOS with Metal

m-sasha:
The main insight here is that when resizing the window we need to present the drawable with the ambient resize transaction (on the AppKit main thread).

To do so, we:
1. During live resize, set `presentsWithTransaction = YES`.
2. Override `CAMetalLayer.setBounds` and synchronously call rendering , ending with a call to `finishFrame` which presents the drawable.

During live resize, the normal redrawing path (on the AWT EDT thread) is disabled to avoid race conditions and deadlocks. Instead, `needRender` schedules drawing the frame on the AppKit main thread, through the same mechanism as the resize frame.

One use-case this doesn't fix is programmatic window resize. When resizing programmatically, the resizing will probably be as before (content lags after window size, showing the infamous white bars).

Tested with
```
package org.jetbrains.skiko

import kotlinx.coroutines.awaitCancellation
import kotlinx.coroutines.delay
import kotlinx.coroutines.launch
import kotlinx.coroutines.runBlocking
import org.jetbrains.skia.Canvas
import org.jetbrains.skia.Paint
import org.jetbrains.skia.Rect
import org.jetbrains.skia.Typeface
import org.jetbrains.skia.paragraph.FontCollection
import org.jetbrains.skia.paragraph.ParagraphBuilder
import org.jetbrains.skia.paragraph.ParagraphStyle
import org.jetbrains.skia.paragraph.TextStyle
import org.jetbrains.skia.paragraph.TypefaceFontProvider
import org.jetbrains.skia.tests.makeFromResource
import org.jetbrains.skiko.util.uiTest
import org.junit.Test
import java.awt.BorderLayout
import java.awt.Color
import java.awt.Dimension
import java.awt.Toolkit
import javax.swing.JFrame
import kotlin.math.absoluteValue
import kotlin.test.BeforeTest
import kotlin.time.Duration.Companion.milliseconds
import kotlin.time.Duration.Companion.nanoseconds
import kotlin.time.Duration.Companion.seconds

class WindowResizeTest {

    @BeforeTest
    fun setup() {
        System.setProperty("skiko.test.ui.enabled", "true")
        System.setProperty("skiko.test.ui.renderApi", "metal")
        System.setProperty("skiko.rendering.macos.metalSynchronousLiveResize", "true")
    }

    @Test(timeout = 1_000_000)
    fun `white borders on resize`() = uiTest {
        val window = UiTestWindow {
[...]

PR #1243, opened 2026-07-23
Make window resizing smooth on Windows with Direct3D

m-sasha:
The mechanism here is similar to the [Metal fix](https://github.com/JetBrains/skiko/pull/1226).

During live resize draw the frame synchronously when receiving the native event. There's no transaction, but with some careful timing and vsync, it works well too.

The normal (async) rendering path is disabled during the live-resize.

The functionality is enabled by the `skiko.rendering.windows.direct3DSynchronousLiveResize` system property (disabled by default for now)

https://github.com/user-attachments/assets/f3b363ef-f5c2-4c2c-8d7d-4cba45ca2593

Fixes https://youtrack.jetbrains.com/issue/CMP-10423/DirectX-Smooth-window-resize

Test code:
```
import kotlinx.coroutines.runBlocking
import kotlinx.coroutines.suspendCancellableCoroutine
import org.jetbrains.skia.*
import org.jetbrains.skia.paragraph.*
import org.jetbrains.skia.tests.makeFromResource
import org.jetbrains.skiko.util.uiTest
import org.junit.Test
import java.awt.BorderLayout
import java.awt.Color
import java.awt.Dimension
import java.awt.Window
import java.awt.event.WindowAdapter
import java.awt.event.WindowEvent
import javax.swing.JFrame
import javax.swing.WindowConstants.DISPOSE_ON_CLOSE
import kotlin.coroutines.resume
import kotlin.math.absoluteValue
import kotlin.test.BeforeTest
import kotlin.time.Duration.Companion.nanoseconds

class WindowResizeTest {

    @BeforeTest
    fun setup() {
        System.setProperty("skiko.fps.enabled", "true")
        System.setProperty("skiko.rendering.windows.direct3DSynchronousLiveResize", "true")
        System.setProperty("skiko.test.ui.enabled", "true")
        System.setProperty("skiko.test.ui.renderApi", "DIRECT3D")
    }

    @Test(timeout = 1_000_000)
    fun `white borders on resize`() = uiTest {
        println(System.getProperty("java.vendor"))
        println(System.getProperty("java.version"))
        val window = UiTestWindow {
            layer.renderDelegate = SolidColorRenderer(layer, Color.BLACK)
//            layer.renderDelegate = HeavyRenderRenderer()
            contentPane.add(layer, BorderLayout.CENTER)

            preferredSize = Dimension(800, 800)
            defaultCloseOperation = DISPOSE_ON_CLOSE
            pack()
            isVisible = true
        }

        suspendCancellableCoroutine { continuation ->
            window.addWindowListener(object : WindowAdapter() {
                override fun windowClosed(e: WindowEvent?) {
                    continuation.resume(Unit)
                }
            })
        }
    }
}

private class SolidColorRenderer(
    private val layer: SkiaLayer,
    color: Color,
) : SkikoRenderDelegate {

    val paint1 = Paint().also { it.color = color.rgb }
    val red = Paint().also { it.color = Color.RED.rgb }

    val style = ParagraphStyle().apply {
        height = 40.0f
        maxLinesCount = 8
        textStyle = TextStyle().apply {
            fontFamilies = arrayOf("Inter")
            fontSize = 42.0f
            this.color = Color.WHITE.rgb
            this.baselineShift = baselineShift
        }
    }

    val fontCollection = runBlocking {
        FontCollection().apply {
            setDefaultFontManager(TypefaceFontProvider().apply {
                val inter = Typeface.makeFromResource("./fonts/Inter-Hinted-Regular.ttf", 0)
                registerTypeface(inter, "Inter")
            })
        }
    }

    var window: Window? = null
    var windowShowTime = 0L

    override fun onRender(canvas: Canvas, width: Int, height: Int, nanoTime: Long) {
        Logger.debug { "Rendering at size $width x $height" }
        canvas.drawRect(Rect(0f, 0f, width.toFloat(), height.toFloat()), paint1)
        canvas.drawRect(Rect(0f, 0f, width.toFloat(), 2f), red)
        canvas.drawRect(Rect(width - 2f, 0f, width.toFloat(), height.toFloat()), red)
        canvas.drawRect(Rect(0f, height-2f, width.toFloat(), height.toFloat()), red)
        canvas.drawRect(Rect(0f, 0f, 2f, height.toFloat()), red)

        val milliTime = nanoTime.nanoseconds.inWholeMilliseconds
        val progress = milliTime.mod(4000)
        val rectPaint = Paint().also {
            val hue = progress / 4000f
            it.color = Color.HSBtoRGB(hue, 1f, 1f)
        }
        val hueRectCenterX = (width / 2f) - ((progress - 2000f).absoluteValue / 1000f - 1f) * width/6
        canvas.drawRect(Rect(hueRectCenterX - width/6f, height/3f, hueRectCenterX + width/6f, 2*height/3f), rectPaint)

        val builder = ParagraphBuilder(style, fontCollection)
        builder.addText("Hello, Skiko")
        val paragraph = builder.build().layout(1600f)

        paragraph.paint(canvas, width/2f - 150, 100f)
        paragraph.paint(canvas, width/2f - 150, height - 100f)
        paragraph.close()

        layer.needRender(throttledToVsync = false)

//         Test calling an AWT method which calls into AppKit
        if ((window != null) && (milliTime - 5000 > windowShowTime)) {
            window!!.dispose()
            window = null
        }
        if ((milliTime + 5000).mod(10_000) in 0..50) {
            if (window == null) {
//                window = JDialog(null as Frame?, true)  // Test modal dialog
                window = JFrame("Another frame")
                window!!.size = Dimension(400, 400)
                windowShowTime = milliTime
                window!!.isVisible = true
            }
        }
    }
}

// Renders a very GPU-heavy picture
private class HeavyRenderRenderer : SkikoRenderDelegate {
    private val paint = Paint().also { it.color = Color.RED.rgb }
    private val loadFill = Paint().also { it.color = Color.GRAY.rgb }

    override fun onRender(canvas: Canvas, width: Int, height: Int, nanoTime: Long) {
        val rect = Rect(0f, 0f, width.toFloat(), height.toFloat())
        repeat(1000) {
            val layerPaint = Paint().also {
                it.imageFilter = ImageFilter.makeBlur(30f, 30f, FilterTileMode.CLAMP)
            }
            canvas.saveLayer(rect, layerPaint)
            canvas.drawRect(rect, loadFill)
            canvas.restore()
        }
        canvas.drawRect(rect, paint)
    }
}
```


skiko/src/awtMain/kotlin/org/jetbrains/skiko/redrawer/EdtInvoker.kt
```diff
@@ -0,0 +1,49 @@
...
+/**
+ * Runs a block on the AWT event dispatch thread (EDT) from another thread — e.g. the Windows toolkit thread
+ * that drives a live resize — and blocks the caller until the block completes, pumping the calling thread's
+ * sent and posted messages meanwhile so the EDT's own cross-thread window ops (marshaled back to the calling thread)
+ * complete instead of deadlocking against the wait.
+ *
+ * The Windows analog of macOS `LWCToolkit.invokeAndWait`, which has no JDK equivalent. The wait event and the
+ * message pump are handled entirely natively; see `edtInvoker.cc`.
+ */
+internal object EdtInvoker {
```
igordmn:
```suggestion
internal object WindowsEdtInvoker {
```

skiko/src/awtMain/kotlin/org/jetbrains/skiko/SkiaLayer.awt.kt
```diff
@@ -606,6 +607,10 @@ actual open class SkiaLayer internal constructor(
         check(isEventDispatchThread()) { "Method should be called from AWT event dispatch thread" }
         check(!isDisposed) { "SkiaLayer is disposed" }
 
+        // Protects against re-entrant calls.
+        // This can actually happen if, for example, renderDelegate.onRender shows a modal dialog during a live-resize.
+        if (isRendering) return
```
igordmn, line 612:
```suggestion
        if (isRendering) {
            needRedraw()
            return
        }
```
If someone requested `redrawImmediatelly` after the state is changed, but `update` didn't read it, we should not skip, but schedule a new frame. Otherwise the new state will be not drawn

skiko/src/awtMain/cpp/windows/edtInvoker.cc
```diff
@@ -0,0 +1,146 @@
...
+    // stutter instead of a hard hang. The abandoned task finishes harmlessly later (see the CloseHandle note below).
+    //
+    // Fully self-contained: creates its own per-call auto-reset event (so independent invocations never
+    // cross-signal), wraps `runnable` to signal that event on completion, posts it (EventQueue.invokeLater),
+    // pump-waits, and closes the event. The only Java involved is the EdtInvocationTask shim, because JNI
+    // cannot fabricate a java.lang.Runnable to post.
+    JNIEXPORT void JNICALL Java_org_jetbrains_skiko_redrawer_EdtInvoker_invokeAndWaitWhilePumping(
+        JNIEnv *env, jobject invoker, jobject runnable)
+    {
+        HANDLE done = CreateEventW(nullptr, /*manualReset*/ FALSE, /*initialState*/ FALSE, nullptr);
```
igordmn:
```suggestion
        HANDLE doneEvent = CreateEventW(nullptr, /*manualReset*/ FALSE, /*initialState*/ FALSE, nullptr);
```

I thought it is a boolean at first glance

skiko/src/awtMain/cpp/windows/edtInvoker.cc
```diff
@@ -0,0 +1,146 @@
...
+            cls = (jclass)env->NewGlobalRef(local);
+            env->DeleteLocalRef(local);
+            mid = env->GetStaticMethodID(cls, "invokeLater", "(Ljava/lang/Runnable;)V");
+        }
+        if (mid) env->CallStaticVoidMethod(cls, mid, runnable);
+        if (env->ExceptionCheck()) { env->ExceptionDescribe(); env->ExceptionClear(); }
+    }
+
+    // Constructs a new EdtInvocationTask(runnable, doneEvent) — the small Java shim that runs `runnable` on the EDT
+    // and then signals `doneEvent`. Class + ctor cached on first use; clears any pending exception (returns null on
```
igordmn: There are a lot of over-explaining comments everywhere. I often skip them because code is easier to read.

I assume they are AI generated, do you know if there is anything bulky that is not essential and we can remove it?

skiko/src/awtMain/kotlin/org/jetbrains/skiko/redrawer/Direct3DRedrawer.kt
```diff
@@ -126,8 +171,49 @@ internal class Direct3DRedrawer(
...
+    @Suppress("unused")
+    private fun onLiveResizeStarted() {
+        isHandlingLiveResizeNow = true
+    }
+
+    /** Called (on the toolkit thread) when the live-resize session ends. */
+    @Suppress("unused")
+    private fun onLiveResizeEnded() {
+        EdtInvoker.invokeAndWaitWhilePumping {
+            javax.swing.SwingUtilities.getWindowAncestor(layer)?.let {
```
igordmn, line 191:
```suggestion
            if (isDisposed) return@invokeAndWaitWhilePumping
            javax.swing.SwingUtilities.getWindowAncestor(layer)?.let {
```

skiko/src/awtMain/cpp/windows/directXRedrawer.cc
```diff
@@ -5,6 +5,7 @@
 #include "jni_helpers.h"
```
igordmn, line 6: I have occasional 2-seconds freezes during quick resizes of `SkiaAwtSample`. Do you have them?

A clue: if I stop mouse during enlarging and freezing, I see a black bar. It disappears when I release the mouse (probably because we call `renderImmediately`)

skiko/src/awtMain/cpp/windows/directXRedrawer.cc
```diff
@@ -120,12 +129,268 @@ class DirectXDevice
...
+// pacing on the actual vblank with a non-blocking Present(0) holds a clean refresh rate.
+static void waitForPrimaryVBlank()
+{
+    if (!g_monitorOutput.get())
+    {
+        gr_cp<IDXGIFactory1> factory;
+        if (SUCCEEDED(CreateDXGIFactory1(IID_PPV_ARGS(&factory))))
+        {
+            gr_cp<IDXGIAdapter1> adapter;
+            if (SUCCEEDED(factory->EnumAdapters1(0, &adapter))) adapter->EnumOutputs(0, &g_monitorOutput);
```
igordmn: I suspect that we shouldn't use this global adapter, but reuse the adapter that is is passed to `createDirectXDevice`.

There can be a memory leak and incorrect behavior in multi-display setup with the current code.

skiko/src/awtMain/cpp/windows/directXRedrawer.cc
```diff
@@ -120,12 +129,268 @@ class DirectXDevice
...
+}
+
+// The live-resize WndProc. Keep the window's own on-screen swapchain as the content and, on each resize step,
+// synchronously render+present it at the NEW client size BEFORE letting the geometry commit — "delay the resize
+// until content is drawn" — so DWM never exposes an unrendered (white) region at the edges. One instance is
+// installed per window; per-window state is recovered from the frame HWND (liveResizeStateFor).
+static LRESULT CALLBACK LiveResizeWndProc(HWND hWnd, UINT msg, WPARAM wParam, LPARAM lParam)
+{
+    LiveResizeState *s = liveResizeStateFor(hWnd);
+    if (!s) return DefWindowProcW(hWnd, msg, wParam, lParam); // hooked but state already torn down; nothing to chain to
```
igordmn: If `uninstallLiveResizeHook` is called in the same thread, when this can happen?

skiko/src/awtMain/cpp/windows/directXRedrawer.cc
```diff
@@ -120,12 +129,268 @@ class DirectXDevice
...
+                // its edge. The active drag never drifts because every NCCALCSIZE re-asserts it; the hold must too.
+                growContentChildTo(s, s->lastClientWidth, s->lastClientHeight);
+                javaDrawFrameWhileLiveResizing(s);
+                // Pace the stationary hold at the refresh rate — one present per vblank (a pure monitor wait on
+                // this toolkit thread; no render/Skia state, so it needn't run on the EDT). Beyond capping idle
+                // FPS, this is what keeps ORIGIN-MOVE (top/left) resize clean: real drags are full of micro-pauses
+                // that hit this WM_PAINT path, and an unpaced hold floods DWM's present queue, adding latency that
+                // desyncs content from the moving window origin. A shallow queue keeps top/left welded to the edge
+                // (right/bottom don't move the origin, so they tolerate the latency).
+                waitForPrimaryVBlank();
```
igordmn: Why don't we use `drawAndSwap(withVsync = true)` inside `javaDrawFrameWhileLiveResizing`?

skiko/src/awtMain/kotlin/org/jetbrains/skiko/SkiaLayer.awt.kt
m-sasha: You mean if someone called `renderImmediately` from inside `renderDelegate.onRender`?
Is that even valid?

In any case, it will cause an infinite "loop" (it will keep re-requesting `needRedraw`) in the case of a modal dialog shown inside ``renderDelegate.onRender`, which is the use-case this is meant to protect against.

skiko/src/awtMain/cpp/windows/edtInvoker.cc
m-sasha: I cleaned up some of the comments that looked obvious (the 2nd sentence in this one is too; removed it), but in general I prefer to leave the long explanations because it's absolutely not obvious to me what's going on here, and the comments are useful both to a future human reader and to a future AI reading the code to understand the subtleties.

igordmn: > I prefer to leave the long explanations

If you are sure that all that is written is useful to you, let's keep it. Did you read all it, does it explain well?

For me personally, there was a lot of unnecessary info that I can't wrap my mind around, so I just read the code, and it explained what happens better.

skiko/src/awtMain/cpp/windows/directXRedrawer.cc
m-sasha: Here's what the AI says:

> Good catch to question it, but no — switching to the adapter from `createDirectXDevice` would be a step in the wrong direction. Here's the reasoning:
>
> **What `WaitForVBlank` actually needs is an *output* (monitor), not a render adapter.** An `IDXGIOutput` is a monitor, and `EnumOutputs` only returns the monitors *physically wired to that adapter*.  
>
> **On the multi-GPU case this bites the other way.** `chooseAdapter` uses `EnumAdapterByGpuPreference(..., adapterPriority)` — with a high-performance preference on a laptop that returns the **discrete GPU**, which very commonly owns **zero outputs** (the dGPU renders and hands off to the iGPU, which actually scans out to the panel). So `device_adapter->EnumOutputs(0, ...)` would just fail, `g_monitorOutput` stays null, and we'd get **no vblank pacing at all** — strictly worse than today's `EnumAdapters1(0)`, which at least tends to land on the adapter driving the primary monitor. 
>
> So the device adapter is the wrong key. What the pacing actually wants is *the monitor the window is currently displayed on*. The current code has two real (pre-existing) shortcomings there, both independent of which adapter renders: 
>
> 1. It hardcodes `EnumAdapters1(0)` → `EnumOutputs(0)`, i.e. always the **first output**, not the one the window is on. Wrong monitor on a multi-monitor setup with differing refresh rates.
> 2. `g_monitorOutput` is a process-global cached **once** and never refreshed, so moving the window to another monitor keeps pacing to the original. 
>
> If we want to make this correct rather than device-adapter-based, the right source is the window's monitor:
> - `MonitorFromWindow(hWnd, MONITOR_DEFAULTTONEAREST)`, then scan adapters/outputs for the one whose `DXGI_OUTPUT_DESC.Monitor` matches — and re-resolve when the window changes monitor; **or**
> - `swapChain->GetContainingOutput()`, which returns exactly the output the swapchain overlaps most (needs the swapchain reachable from here, which it isn't in this static helper today). 
>
> `hWnd` is in scope at the `waitForPrimaryVBlank()` call site (WM_PAINT in `LiveResizeWndProc`), so the `MonitorFromWindow` version is straightforward to wire up. 
>
> That said, this is a pacing nicety with a non-blocking `Present(0)` behind it — for a single-monitor or uniform-refresh-rate setup the current code already produces the right cadence, and the failure mode is a slightly-off hold rate, not a visual glitch. 
>
> Want me to replace it with the `MonitorFromWindow`-based lookup (passing `hWnd` through, and re-resolving on monitor change)? That fixes the genuine multi-monitor bug and is the "correct" version — whereas the device-adapter route would regress dGPU laptops.

(I told it yes)

skiko/src/awtMain/cpp/windows/directXRedrawer.cc
m-sasha: I also had this thought, but unfortunately no.
The "wait" here is somehow different from what `swap(isVsyncEnabled=true)` does.
If I call `drawAndSwap(withVsync = true)` there's jumpiness when resizing by the top or left window edges.

> Good question — but no, and the code actually documents why. Let me confirm the `withVsync → Present` mapping to be precise.
>
> Confirmed: `drawAndSwap(withVsync = true)` → `swap(device, true)` → `Present(1, 0)`. That's precisely the mechanism the code deliberately avoids, for two independent reasons:
>
>**1. It would pace the active drag too — which must stay unpaced.** `drawFrameWhileLiveResizing` is the *single* present path shared by **both** WndProc callers:
> - WM_NCCALCSIZE (active drag) — [directXRedrawer.cc:336](skiko/src/awtMain/cpp/windows/directXRedrawer.cc), commented "active drag: unpaced"
> - WM_PAINT (stationary hold) — [directXRedrawer.cc:355](skiko/src/awtMain/cpp/windows/directXRedrawer.cc), followed by the vblank wait
>
> Flipping the `withVsync` flag there applies `Present(1)` to the active drag as well. The active drag is meant to be driven at mouse-move cadence via WM_NCCALCSIZE; blocking each of those on vblank throttles the drag against the display instead of the input, which is exactly what the design keeps separate ("an active drag stays unpaced, driven by mouse-move WM_NCCALCSIZE"). You'd need to thread a per-call flag through to vsync *only* the hold path — at which point you haven't really removed the separate-pacing concept, just moved it.
>
> **2. `Present(1)` is the wrong pacer here anyway — that's the whole reason `WaitForVBlank` + `Present(0)` exists.** Per the comment at [directXRedrawer.cc:272](skiko/src/awtMain/cpp/windows/directXRedrawer.cc): on a windowed, DWM-composited flip swapchain, `Present(1)`'s sync interval beats against DWM's own composition cadence and lands the loop at **refresh × 2/3 (~40 fps on a 60 Hz screen)**. Pacing on the *actual* vblank and presenting non-blocking (`Present(0)`) is what holds a clean refresh rate. So even restricted to the hold path, the vsync approach gives measurably worse cadence than what's there now.
> 
> There's also a lesser point: the present runs on the **EDT** (via `invokeAndWaitWhilePumping`), while `waitForWindowMonitorVBlank` runs on the **toolkit thread** *after* the round-trip returns — "Native code handles the vsync to avoid blocking the EDT" ([Direct3DRedrawer.kt:211](skiko/src/awtMain/kotlin/org/jetbrains/skiko/redrawer/Direct3DRedrawer.kt)). Present(1) would hold the EDT for up to a frame while it waits.
> 
> So the dedicated wait isn't redundant with the vsync flag — it exists specifically to (a) pace only the hold, not the drag, (b) get a clean refresh-rate cadence that `Present(1)` can't deliver on a composited window, and (c) keep the wait off the EDT.

~But I've removed my misleading comment after the `drawAndSwap(withVsync = false)`~

m-sasha: Also, it would block the EDT thread (here we're blocking the native thread)

m-sasha: Hmm, actually forced it to try (but it needs to pass a `vsync` flag into `javaDrawFrameWhileLiveResizing`) and it works ok. The only downside is that we'd be blocking the EDT thread.

skiko/src/awtMain/cpp/windows/directXRedrawer.cc
igordmn: > needs the swapchain reachable from here

You can use `d3dDevice->swapChain` and pass it in `LiveResizeState`

m-sasha: I'll just get rid of it as we'll be doing the vsync via `swap`

skiko/src/awtMain/cpp/windows/directXRedrawer.cc
m-sasha: I'll investigate it.

skiko/src/awtMain/cpp/windows/edtInvoker.cc
m-sasha: I don't have enough knowledge to say whether it's useful or obvious for someone who understands the Windows/Direct3D APIs. But it's the result of an investigation and many failed attempts, so it at least documents why things are the way they are.

skiko/src/awtMain/cpp/windows/directXRedrawer.cc
m-sasha: It doesn't look like it's always called in the same thread. `Direct3DRedrawer.dispose` is typically called on the AWT EDT thread. `LiveResizeWndProc` is called on the native Windows toolkit thread.

The AI says there's another reason for this guard:

> But that's not even the main reason the guard exists.
>
> **The guard's real purpose: uninstall doesn't always unhook our proc — but it always removes the prop.** Look at [uninstallLiveResizeHook](skiko/src/awtMain/cpp/windows/directXRedrawer.cc):

```cpp
WNDPROC current = (WNDPROC)GetWindowLongPtrW(state->frameHwnd, GWLP_WNDPROC);
if (current == LiveResizeWndProc && state->originalProc) {          // only when we're STILL the top proc
    SetWindowLongPtrW(state->frameHwnd, GWLP_WNDPROC, (LONG_PTR)state->originalProc);
}
RemovePropW(state->frameHwnd, kLiveResizeStateProp);                // ...but the prop goes regardless
...
delete state;
```

> If some other subclass has layered over us since install (`current != LiveResizeWndProc`), restoring the original proc blindly would clobber that foreign subclass — so uninstall deliberately **leaves `LiveResizeWndProc` wired into the chain** and only removes the prop.

skiko/src/awtMain/cpp/windows/directXRedrawer.cc
igordmn, line 2: What worries me with new live-resize code that there is no tests on it. Is it possible to use `Robot` for resizing?

skiko/src/awtMain/cpp/windows/directXRedrawer.cc
igordmn: > It would pace the active drag too

This looks like wrong reasoning. Having different drag/paint behaviors can be just solved by calling different functions.

> 2/3 (~40 fps on a 60 Hz screen)

If that is the case, we would have the same on usual  rendering where we use `Present(1)`.

The main issue here is that we use different vsync logic for usual animations and animations during resize - it complicates maintenance and can bring bugs. If this logic is wrong, we can replace it to `waitForVBlank` in the future.

> There's also a lesser point: the present runs on the EDT

This is valid, but the solution is the same as in Metal - call `swap(withVsync)` after the AWT callback (with drawLock)

skiko/src/awtMain/cpp/windows/edtInvoker.cc
igordmn: > I don't have enough knowledge to say whether it's useful or obvious for someone who understands the Windows/Direct3D APIs. But it's the result of an investigation and many failed attempts, so it at least documents why things are the way they are.

I can't agree with this. If we add new code/comments, we should understand it. If we don't understand, we should remove it. AI often misleads, wrong, tends to overexplain things, and adds "in-between-conversation" corrections that has no place in the comment. Wrong info is much worse than no info.

igordmn: > I don't have enough knowledge to say whether it's useful or obvious for someone who understands the Windows/Direct3D APIs

A correction - if you understand, and just don't know if it is obvious or not, this is enough to keep the comment.

igordmn: My question is - do you understand everything that is written and is it helpful to you?

skiko/src/awtMain/cpp/windows/directXRedrawer.cc
igordmn: Right, we install/uninstall in the AWT thread. 

> If some other subclass has layered over us since install

This looks like a correct reason, I would add it as a comment here.

But AI wording is difficult here, I would rewrite (`layered`, `wired into the chain` - means calling `originalProc` the same way, `clobber` - probably a rare word)

skiko/src/awtMain/cpp/windows/directXRedrawer.cc
m-sasha: > This is valid, but the solution is the same as in Metal - call swap(withVsync) after the AWT callback (with drawLock)

Tried it, but it causes crashes. Not sure why (neither is the AI).

skiko/src/awtMain/cpp/windows/edtInvoker.cc
m-sasha: In my theory about comments is there are two kinds. They answer:

## What does this code do?
This kind of comment is generally considered not useful because you can just read the code and see what it does. However it is useful if the reader has a poor understanding of the code and underlying APIs.

So in the case of AI-generated code in an area I'm not an expert in I think they're moderately useful. But I did try to limit them to the parts that are related to native APIs.

## Why does this code do what it does?
These are considered good and useful, and we should definitely keep them.

skiko/src/awtMain/cpp/windows/directXRedrawer.cc
m-sasha: Fixed

skiko/src/awtMain/cpp/windows/edtInvoker.cc
igordmn: I am not talking about types of the comments, but rather correctness, complexity, applicability, over-explanation and misleading.

For me, personally I see these comments difficult to read comparing to the code, and have a guess that they are not entirely correct and misleading. I read them partially, and we can discuss the approach if you want.

The only criteria of keeping them I see is that there is some real person, for example you, that understands them and thinks that they are helpful for them (not only for AI).

m-sasha: Are you seeing any comments that are incorrect? Non applicable? Misleading? Let's fix that if you do.

Complex and over-explaining are relative and need domain knowledge to fix. Since I don't have it, I can't fix it. I can ask the AI to be more terse if you'd like, but I don't know if it will result in better comments. If you can fix it, I'm all for it.

skiko/src/awtMain/cpp/windows/directXRedrawer.cc
m-sasha: It's not hard to use Robot to resize a window, but it'll be much harder to catch the visual glitches.

See the "no background flash on window show" test and how complicated that is, and it's much easier because you know where to look for the white pixels, and the white pixels are the only symptom (here you can have white bars, but also jumpiness, hiccups/micro-freezes).

igordmn: > It's not hard to use Robot to resize a window, but it'll be much harder to catch the visual glitches.

We can just check that there are no crashes. The new code can crash or interfere with the previous code. We can also assert the final size of the canvas or/and the final screenshot

skiko/src/awtMain/cpp/windows/directXRedrawer.cc
```diff
@@ -120,12 +127,185 @@ class DirectXDevice
...
+    int childW = (cr.right - cr.left) > w ? (cr.right - cr.left) : w;
+    int childH = (cr.bottom - cr.top) > h ? (cr.bottom - cr.top) : h;
+    SetWindowPos(s->contentHwnd, nullptr, 0, 0, childW, childH,
+                 SWP_NOMOVE | SWP_NOZORDER | SWP_NOACTIVATE | SWP_NOOWNERZORDER);
+}
+
+static LRESULT CALLBACK LiveResizeWndProc(HWND hWnd, UINT msg, WPARAM wParam, LPARAM lParam)
+{
+    LiveResizeState *s = liveResizeStateFor(hWnd);
+    if (!s) return DefWindowProcW(hWnd, msg, wParam, lParam);
```
igordmn:
```suggestion
    // This can happen when:
    // - something else registered another procedure via `SetWindowLongPtrW` that delegates to the old one (this one)
    // - `uninstallLiveResizeHook` is called - it nullified `liveResizeStateFor` without resetting `SetWindowLongPtrW`
    if (!s) return DefWindowProcW(hWnd, msg, wParam, lParam);
```

skiko/src/awtMain/cpp/windows/directXRedrawer.cc
```diff
@@ -120,12 +127,185 @@ class DirectXDevice
...
         return swapChainFactory4->CreateSwapChainForHwnd(queue.get(), hWnd, &swapChainDesc, nullptr, nullptr, swapChain1);
     }
 };
 
+// ===================== Direct3D synchronous live-resize =====================
+// Renders and presents the content synchronously inside WM_NCCALCSIZE, before the new geometry commits, so DWM never
+// composites a frame whose edges we haven't painted yet.
+//
+// Not SetWindowSubclass, which would chain more cleanly: it must be called from the thread that OWNS the window, and
+// we install from the EDT while the AWT frame HWND belongs to the toolkit thread. Cross-thread it just returns FALSE.
```
igordmn:
```suggestion
// we install from the EDT while the AWT frame HWND belongs to the toolkit thread.
```

I can't understand this sentence.

skiko/src/awtMain/cpp/windows/directXRedrawer.cc
m-sasha: Did it differently, as there are now two procs

skiko/src/awtMain/cpp/windows/directXRedrawer.cc
m-sasha: Added a test

skiko/src/awtMain/cpp/windows/directXRedrawer.cc
```diff
@@ -120,12 +128,223 @@ class DirectXDevice
...
+
+struct LiveResizeState {
+    WNDPROC originalProc = nullptr;
+    WNDPROC originalContentProc = nullptr;
+    HWND frameHwnd = nullptr;
+    HWND contentHwnd = nullptr;
+    jobject redrawer = nullptr;     // global ref
```
igordmn:
```suggestion
    jobject redrawer = nullptr;
```

It looks local for `LiveResizeState` and we set it per-redrawer

skiko/src/awtMain/cpp/windows/winApiEdtInvoker.cc
```diff
@@ -0,0 +1,113 @@
...
+            jclass local = env->FindClass("java/awt/EventQueue");
+            cls = (jclass)env->NewGlobalRef(local);
+            env->DeleteLocalRef(local);
+            mid = env->GetStaticMethodID(cls, "invokeLater", "(Ljava/lang/Runnable;)V");
+        }
+        if (mid) env->CallStaticVoidMethod(cls, mid, runnable);
+        if (env->ExceptionCheck()) { env->ExceptionDescribe(); env->ExceptionClear(); }
+    }
+
+    // A Java shim is unavoidable here: JNI cannot fabricate a java.lang.Runnable to post.
```
igordmn:
```suggestion
    // java.lang.Runnable can't be created from native code, so delegate its creation to Java
```

skiko/src/awtMain/cpp/windows/winApiEdtInvoker.cc
```diff
@@ -0,0 +1,113 @@
...
+
+    // The Win32 counterpart of the nested CFRunLoop macOS LWCToolkit.invokeAndWait spins. Which message classes it
+    // pumps is the whole design:
+    //
+    // POSTED as well as SENT, because EDT work makes synchronous cross-thread calls back to this thread that travel
+    // as posted messages — showing a window from a Compose recomposition posts focus/IME setup here and blocks the
+    // EDT on the reply. A sent-only pump starves those and the EDT deadlocks against us.
+    //
+    // But never input: we are nested inside the modal move/size loop, which owns the drag's input, and draining it
+    // stops the window following the cursor.
```
igordmn:
```suggestion
    // While waiting, handle SENT and POSTED messages. EDT can send POSTED messages (focus/IME) back here and wait for them.
    //
    // Do not handle input messages. They are handled by the parent event loop. Handling
    // them here stops the window following the cursor.
```

This comment is difficult to read.

igordmn (approved):

skiko/src/awtMain/cpp/windows/winApiEdtInvoker.cc
m-sasha: I simplified the WinEdtInvoker instead.

igordmn (approved):

PR #1247, opened 2026-07-28
[web] Extract wasm-opt phase to a separate step

nopjar:
This PR introduces a separate optimization step for wasm target.
See https://youtrack.jetbrains.com/issue/SKIKO-1159

It is a cleaner approach than the one included in #1242.

Emscripten basically does the same, but with `-O2` instead of `-Oz`, which results in a size reduction of ~200KB (9 MB -> 8.8 MB).

Ticket can be found [here](https://youtrack.jetbrains.com/issue/SKIKO-1159/Introduce-wasm-opt-as-a-separate-build-step).


skiko/buildSrc/src/main/kotlin/tasks/configuration/WasmTasksConfiguration.kt
```diff
@@ -187,6 +187,29 @@ fun SkikoProjectContext.declareWasmTasks() {
...
+        buildVariant.set(buildType)
+
+        optimizer.set(project.findProperty("wasm.opt.path")?.toString() ?: "wasm-opt")
+        val wasmFileProvider = linkWasm.flatMap { it.outDir.file(it.libOutputFileName) }
+        inputFile.set(wasmFileProvider.get().asFile.absolutePath)
+        libOutputFileName.set("$libBaseName.wasm")
+
+        flags.addAll(
+            listOf(
+                "-Oz",
```
Schahen: can you please leave comments for each of the flags so that in future it will be easier to read this code without using any reference

skiko/buildSrc/src/main/kotlin/tasks/configuration/WasmTasksConfiguration.kt
```diff
@@ -187,6 +187,29 @@ fun SkikoProjectContext.declareWasmTasks() {
         configureCommon(prefixPath)
     }
 
+    val optimizeWasm by project.tasks.registering(OptimizeSkikoWasmTask::class) {
+        dependsOn(linkWasm)
+        buildSuffix.set("es6")
```
eymar: is this argument (es6) relevant to wasm-opt?

nopjar: Actually not. I just added it as `linkWasm` also has it. I can remove it, if you want.

Schahen: @nopjar yes please, I mean we should have only flags that do configure the actual task

skiko/buildSrc/src/main/kotlin/AbstractSkikoNativeToolTask.kt
```diff
@@ -102,10 +103,11 @@ abstract class AbstractSkikoNativeToolTask : DefaultTask() {
 
     protected open fun beforeRun() {}
     protected open fun afterRun() {}
-    protected fun logArgs(prefix: String, args: ArgBuilder, argFile: File) {
-        if (logger.isInfoEnabled) {
+    protected fun logArgs(prefix: String, args: ArgBuilder, argFile: File? = null) {
+        if (logger.isInfoEnabled || argFile == null) {
             val argsString = args.toArray().joinToString(", ", prefix = "[", postfix = "]")
-            logger.info("$prefix: $argsString")
+            val logLevel = if (logger.isInfoEnabled) LogLevel.INFO else LogLevel.WARN
```
Schahen, line 109: What's the purpose of this changes?

Schahen (changes requested):
Apart from the comments left, there's one fundamental thing this PR should take care of so that we won't harm workflow for our fellow skiko developers.

the EMSDK_DIR is a good direction - let's discuss whether it should be gradle property rather then environment variable (@MatkovIvan what is your though on this) - however ideologically its the right thing to do. 

However what will be broken is that for people who don't have this param set but already been relying on the existence of emsdk/upstream (to which the `which emcc` is resolved). 

Can you please, add following logic:
1) if EMSDK_DIR is present (again, later on may be we'll switch to a property) then both emcc and wasm-opt are resolved relying on it

2) if it is not present, then we fallback to emcc  and if that emcc is actually in the upstream folder of the emsdk then we resolve wasm-opt accordingly

3) otherwise we fall back to wasm-opt available globally

Schahen:
Another thing that make sense to do in the framework of this PR is to make sure that emcc step is not invoking wasm-opt by itself.

skiko/buildSrc/src/main/kotlin/AbstractSkikoNativeToolTask.kt
nopjar: I've oriented myself on the `LinkSkikoTask.kt` and the `CompileSkikoCppTask.kt`. Both store their arguments for the `emcc` command in a text file and call it with `@path/to/args.txt`. That syntax is not support by wasm-opt. So we need to pass the arguments directly. But that method requires an arg-file. So I updated the method to also accept no arg-file while the result is unchanged and still respects log-levels like before.

nopjar:
> Another thing that make sense to do in the framework of this PR is to make sure that emcc step is not invoking wasm-opt by itself.

I just did a few experiments and I couldn't find a way to just disable the wasm-opt optimization without having to sacrifice other optimizations that emcc currently does.

skiko/buildSrc/src/main/kotlin/tasks/configuration/WasmTasksConfiguration.kt
nopjar: Done :)

Schahen:
@nopjar are you comparing with the +wasm-opt scenario?
in theory switching off the emcc-side wasm-opt invocation should not affect negatively artefact size if we still have ours.

can you elaborate on that?

nopjar:
Yes, passing `-O1` instead of `-O2` will result in a optimized wasm which is roughly 100 KB bigger. The 'unoptimized' is even 700 KB bigger.

If I understand it correctly, it passes the same arg to clang and wasm-opt. We can't therefore only omit one.
https://github.com/WebAssembly/binaryen/discussions/5598

Schahen (approved):
I don't have any strong subjections anymore

skiko/buildSrc/src/main/kotlin/tasks/configuration/WasmTasksConfiguration.kt
```diff
@@ -187,6 +189,68 @@ fun SkikoProjectContext.declareWasmTasks() {
...
+        }
+
+        optimizer.set(wasmOptPath)
+        inputDir.set(linkTask.flatMap { it.outDir })
+        libOutputFileName.set("$libBaseName$nameSuffix")
+
+        flags.addAll(
+            listOf(
+                "-Oz", // set optimization level to compress (highest size reduction)
+                "--strip-debug", // strip debug info (including the names section)
```
eymar: We have an additional mode to build skiko.wasm for profiling needs:
```
if (skiko.isWasmBuildWithProfiling) add("--profiling")
```

Such a skiko.wasm should preserve the names so we can see them in a profiler. 

But with `--strip-debug` removes the names, which breaks the expectations from the "profiling" skiko.wasm. I think that OptimizeSkikoWasmTask should be aware of the `skiko.isWasmBuildWithProfiling`.

Could you please ensure that the names are stripped conditionally?

skiko/buildSrc/src/main/kotlin/tasks/configuration/WasmTasksConfiguration.kt
```diff
@@ -187,6 +189,68 @@ fun SkikoProjectContext.declareWasmTasks() {
...
+            }
+        }
+
+        optimizer.set(wasmOptPath)
+        inputDir.set(linkTask.flatMap { it.outDir })
+        libOutputFileName.set("$libBaseName$nameSuffix")
+
+        flags.addAll(
+            listOf(
+                "-Oz", // set optimization level to compress (highest size reduction)
```
eymar: Since `-Oz` optimizes size, it might affect the performance (comparing with `-O2`).
We have Compose benchmarks - https://github.com/JetBrains/compose-multiplatform/tree/master/benchmarks/multiplatform
Have a look at different modes in README - https://github.com/JetBrains/compose-multiplatform/tree/master/benchmarks/multiplatform#benchmark-modes
Checking with SIMPLE and REAL should be enough. 

Before leaving it `-Oz` it's worth to verify `O2` vs `Oz` on Compose benchmarks. Could you please check it?

Alternatively, to move forward with the PR sooner than later, we can set O2 here for now and then compare O2 vs Oz on the benchmarks + change to Oz in a new PR. 

What do you prefer?

nopjar: After running some benchmarks with different cases, the conclusion is that with `-Oz` we do not have a regression overall. Neither compared to the current master nor with just passing `-O2` again.

This table illustrates it to some extend. Also one needs to keep in mind that of course some variation in the timings is possible as various external factors influence the execution. Therefore, there were also runs where `-Oz` performed better and worse than `master` in almost all categories. But it has been always just a few percent. So I'd say that there is almost no impact (if at all) to performance.

| Benchmark | -Oz (ms) | -O2 (ms) | master (ms) |
|--------|--------|--------|--------|
| AnimatedVisibility | 0.090 | 0.091 | 0.090 |
| CanvasDrawing | 9.524 | 9.621 | 9.552 |
| HeavyShader | 1.612 | 1.626 | 1.597 |
| LazyGrid | 4.695 | 4.635 | 4.642 |
| LazyGrid-ItemLaunchedEffect | 4.660 | 4.602 | 4.689 |
| LazyGrid-SmoothScroll | 0.448 | 0.443 | 0.457 |
| LazyGrid-SmoothScroll-ItemLaunchedEffect | 0.463 | 0.463 | 0.467 |
| LazyList | 0.124 | 0.124 | 0.124 |
| MultipleComponents | 0.179 | 0.179 | 0.182 |
| MultipleComponents-NoVectorGraphics | 0.146 | 0.145 | 0.147 |
| TextLayout | 41.874 | 41.532 | 41.423 |
| VisualEffects | 1.034 | 1.024 | 1.011 |

So if nothing from your side speak against it, I would go with the `-Oz` step as it also reduces size.

eymar (approved):

PR #1248, opened 2026-07-29
Run iOS simulator arm64 tests with Metal in CI

hub-bla:


.github/workflows/tests.yml
```diff
@@ -100,17 +100,19 @@ jobs:
...
-            -Dskiko.test.onci=true \
-            -Dskiko.test.performance.enabled=false \
-            iosSimulatorArm64Test
-
-      # TODO run iOS specific tests on iPhone simulator
+      - uses: nick-fields/retry@v2
+        name: 'Run iOS (arm64) tests'
+        with:
+          max_attempts: 10
+          timeout_minutes: 60
```
MatkovIvan: Isn't it too high?

Can we lower it to 3 retries and 30 min timeout? I believe that is enough

hub-bla: No blockers here, I'll do it for the existing x64 as well then

hub-bla: Done

MatkovIvan (approved):

PR #1250, opened 2026-07-29
Fix RenderNode static assertion constexpr compatibility

hub-bla:
With LLVM 22.1.3 the following error occurs:
```
skiko/skiko/src/commonMain/cpp/common/node/RenderNode.cpp:38:15: error: static assertion expression is not an integral constant expression
  >    38 | static_assert(static_cast<uint32_t>(kRenderNode_Type) > static_cast<uint32_t>(SkFlattenable::kSkShader_Type),
```
This PR fixes the issue.

### DIFF
--- a/skiko/src/commonMain/cpp/common/node/RenderNode.cpp
+++ b/skiko/src/commonMain/cpp/common/node/RenderNode.cpp
@@ -32,11 +32,9 @@ static const float NON_ZERO_EPSILON = 0.001f;
 
 // Since "kSkDrawable_Type" isn't really used anywhere outside of serialization, we can use it
 // to identify RenderNode objects without RTTI, like SkRuntimeEffect does (even without adding to original enum).
-static const SkFlattenable::Type kRenderNode_Type = static_cast<SkFlattenable::Type>(0x2d2595b6);
 // The value has to stay outside the range Skia assigns, or a node is mistaken for a
 // flattenable of that type. kSkShader_Type is the largest Skia declares.
-static_assert(static_cast<uint32_t>(kRenderNode_Type) > static_cast<uint32_t>(SkFlattenable::kSkShader_Type),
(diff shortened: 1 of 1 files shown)


skiko/src/commonMain/cpp/common/node/RenderNode.cpp
```diff
@@ -32,10 +32,11 @@ static const float NON_ZERO_EPSILON = 0.001f;
 
 // Since "kSkDrawable_Type" isn't really used anywhere outside of serialization, we can use it
 // to identify RenderNode objects without RTTI, like SkRuntimeEffect does (even without adding to original enum).
-static const SkFlattenable::Type kRenderNode_Type = static_cast<SkFlattenable::Type>(0x2d2595b6);
+static constexpr uint32_t kRenderNode_TypeValue = 0x2d2595b6;
+static const SkFlattenable::Type kRenderNode_Type = static_cast<SkFlattenable::Type>(kRenderNode_TypeValue);
 // The value has to stay outside the range Skia assigns, or a node is mistaken for a
 // flattenable of that type. kSkShader_Type is the largest Skia declares.
-static_assert(static_cast<uint32_t>(kRenderNode_Type) > static_cast<uint32_t>(SkFlattenable::kSkShader_Type),
+static_assert(kRenderNode_TypeValue > static_cast<uint32_t>(SkFlattenable::kSkShader_Type),
```
MatkovIvan: Well, if it doesn't work without such a split, I'd suggest just removing this `static_assert` entirely.
It was a pre-caution of the future skia changes, but we already choose so large a value that the original enum won't realistically reach any time

hub-bla: Removed

MatkovIvan (approved):

PR #1251, opened 2026-07-30
feat: add gradle task to automatically download emscripten

nopjar:
Currently, developers and build pipeline are expected to have a working emcc installation in their path.

This PR introduces a new gradle task that will pull and install emscriptens emcc locally in the gradle user home.


skiko/buildSrc/src/main/kotlin/tasks/configuration/WasmTasksConfiguration.kt
```diff
@@ -50,16 +51,33 @@ private val wasmSideModuleLinkTaskAttribute =
 private const val WASM_SIDE_MODULE_USAGE = "skiko-wasm-side-module"
 private const val WASM_TEST_RESOURCES_USAGE = "skiko-wasm-test-resources"
 
+private const val DEFAULT_EMSDK_VERSION = "4.0.7"
```
hub-bla: Maybe it's worth putting it in `gradle.properties`. wdyt?

skiko/buildSrc/src/main/kotlin/EnsureEmscriptenTask.kt
```diff
@@ -0,0 +1,123 @@
...
+        if (result.exitValue != 0) {
+            throw GradleException(
+                "Command '${args.joinToString(" ")}' failed with exit code ${result.exitValue}.\n" +
+                "stdout: $stdoutStr\n" +
+                "stderr: $stderrStr"
+            )
+        }
+    }
+
+    private fun isWindows(): Boolean =
```
hub-bla: We already have such parsing logic defined.
See: https://github.com/JetBrains/skiko/blob/be47d484fbd560974098a5e30bb9ed3c65de5cb1/skiko/buildSrc/src/main/kotlin/properties.kt#L103

skiko/buildSrc/src/main/kotlin/EnsureEmscriptenTask.kt
```diff
@@ -0,0 +1,123 @@
...
+        emsdkDir.get().asFile.resolve("upstream/emscripten/${if (isWindows()) "emcc.bat" else "emcc"}")
+
+    private fun installEmsdk() {
+        val version = emsdkVersion.get()
+        val sdkDir = emsdkDir.get().asFile
+
+        // Clone or update emsdk
+        if (sdkDir.resolve(".git").isDirectory) {
+            logger.lifecycle("Updating existing emsdk clone...")
+            exec("git", "pull", workingDir = sdkDir)
```
hub-bla: Maybe we could pin the `emsdk` revision instead of pulling. Seems to be more reproducible. wdyt?

skiko/buildSrc/src/main/kotlin/tasks/configuration/WasmTasksConfiguration.kt
```diff
@@ -50,16 +51,33 @@ private val wasmSideModuleLinkTaskAttribute =
...
+
 fun SkikoProjectContext.declareWasmTasks() {
     if (!project.supportWeb) {
         return
     }
     val isSideModule = kind == SkikoModuleKind.EXTENSION
 
+    val emsdkVersion = project.findProperty("skiko.emsdk.version")?.toString() ?: DEFAULT_EMSDK_VERSION
+    val ensureEmscripten by project.tasks.registering(EnsureEmscriptenTask::class) {
```
hub-bla: Since there are modules that depend on core, perhaps we only need to register this task there? I wonder if we currently install `emsdk` multiple times, e.g. for both `skiko-skottie` and "core" skiko

skiko/buildSrc/src/main/kotlin/tasks/configuration/WasmTasksConfiguration.kt
nopjar: Having it in a more global settings-like place sounds good to me.

skiko/buildSrc/src/main/kotlin/EnsureEmscriptenTask.kt
nopjar: Sounds good! Should we rather reuse the `emsdkVersion` that we already define and use it to clone/checkout via git-tag or do you think having a separate property that allows to checkout at a specific sha is better?

Reusing the version for the git-tag of course requires that the tags actually correspond to the emsdk-version which I don't know.

skiko/buildSrc/src/main/kotlin/tasks/configuration/WasmTasksConfiguration.kt
nopjar: If there are multiple execution, then `isEmccAvailable()` would prevent multiple installations.

From what I see in the `build.gradle.kts` of the `:skiko` module, the installation task would not really fit in there. It mainly just wires things. The `WasmTasksConfiguration` suits the "heavy-work" theme of the installation task a bit more imho. But I may have overlooked something so please correct me if I am wrong or misunderstood something.

@Schahen What do you think?

.github/workflows/publish-dry-run.yml
```diff
@@ -55,6 +55,7 @@ jobs:
           command: |
             ./gradlew -p skiko --no-daemon --stacktrace \
               -Pskiko.wasm.enabled=true \
+              -Pskiko.emsdk.dir="\$EMSDK_DIR" \
               publishToMavenLocal
```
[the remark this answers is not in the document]
nopjar: We want that env to not be replaced in the github action, but on the ci runner. So escaping it is intended.

Schahen:
This branch in its current state (HEAD f28cff222) does not built for me with following error:
```
[ shabunc skiko (feat/auto-download-emscripten)]$ ./gradlew   :skiko:publishToMavenLocal -Pskiko.wasm.enabled=true -Pskiko.awt.enabled=false -Pskiko.native.enabled=false  -Pdeploy.version=0.0.1-SHAGEN   -Pdeploy.release=true

FAILURE: Build failed with an exception.

* Where:
Build file '/Users/shagen/code/skiko/skiko/build.gradle.kts' line: 249

* What went wrong:
skiko.emsdk.version property is not set


```

skiko/buildSrc/src/main/kotlin/EnsureEmscriptenTask.kt
hub-bla: AFAIR, Skia itself pins the `emsdk` version using a specific SHA. We could try just reusing that one.

If not, then I think using the latest stable version is fine. The main point is just to ensure that the dependency is pinned to a fixed state so it doesn't change unless we explicitly want it to

nopjar: Good call, I've found the sha in skia and used the exact same as for this PR. Also the checkout is now done at the specified commit and the commit is also used to make sure the cache-directory is only used for that exact commit.

skiko/buildSrc/src/main/kotlin/tasks/configuration/WasmTasksConfiguration.kt
hub-bla: I agree that having it in `WasmTasksConfiguration` is better place for it. 

What I meant was to keep its implementation in `WasmTasksConfiguration` (just as it is currently) and simply use the configuration created only in the `build.gradle.kts` of the `:skiko` module

But if `isEmccAvailable()` prevents multiple downloads, then that’s fine with me.

skiko/buildSrc/gradle.properties
```diff
@@ -1 +1 @@
-kotlin.version=2.3.20
+kotlin.version=2.3.20
```
Schahen: Please avoid changes that does not supposed to be part of PR (even if the y do nothing)

skiko/gradle.properties
```diff
@@ -6,6 +6,8 @@ kotlin.mpp.enableCInteropCommonization=true
 kotlin.js.yarn.lock.file=false
 skiko.kotlin.language.version=2.2
 skiko.kotlin.api.version=2.2
+skiko.emsdk.version=4.0.7
+skiko.emsdk.commit=c69d433d8509c5c64564c2f0d054bf102a5cf67e
```
Schahen: What is important is to have one param instead of two because they actually supposed to have one-to-one correspondence and because having less params is always better to have one. 

So, it's better to leave just skiko.emsdk.version=4.0.7

Also, I'd rather download zip archive if it is present rather than cloning git repository - but this is more matter of personal taster rather then crucial aspect of the build - it's just that take a look, may be there's alreadt something in project that can be repurposed for downloading zipped artifacts

Schahen (changes requested):

Schahen (approved):
It would be nice to name ensureEmscripten setupEmscripten - just a tad more intuitive name for actions of such nature - but that's just nitpicking already, apart from that have nothing to add to this review.

hub-bla (approved):
LGTM!

MatkovIvan:
Does it work correctly inside pre-installed `emcc` in docker (`skiko/docker/linux-emscripten-amd64/Dockerfile`) ?
Do we even need a separate docker image for it now?

nopjar:
> Does it work correctly inside pre-installed `emcc` in docker (`skiko/docker/linux-emscripten-amd64/Dockerfile`) ? Do we even need a separate docker image for it now?

It should make images with a pre-installed emscripten binary redundant, yes.

Unfortunately, the installation script of emscripten requires a higher version of python than our base linux image provides.
More details are in this commit where I had to undo/reintroduce the emscripten image: https://github.com/JetBrains/skiko/pull/1251/commits/76197318aab6d953a84492d4041fcfcb0f6e373a

But it *should* be not needed anymore once the base linux image gets a bump in the python version.

hub-bla:
From what I understand the `linux-emscripten` Docker image is there because the TeamCity agent does not have Emscripten installed. Docker allows us to change the Emscripten version without modifying the TeamCity agent configuration so it might be possible to remove it if the TeamCity agent has the required python version.

If not then, I recently updated `linux-emscripten` to a slightly newer version of Ubuntu exactly because of the Python version. I think it should be possible to drop the current installation of Emscripten from the Docker image and download it through this new Gradle task

PR #1252, opened 2026-07-30
Fix graphite publishing

hub-bla:
Previously, Graphite’s iOS and tvOS target artifacts could be published, but its root `skiko-graphite` module was missing. This prevented normal KMP dependency resolution through:                                                  
```
implementation("org.jetbrains.skiko:skiko-graphite:<version>")                                                                                                                                                           
```                                                                                                                                                                                                                                     
The common publication task (in TeamCity) ran only on Linux CI, while `skiko-graphite` was included only on macOS. Exposing the project on Linux closes that CI coverage gap without publishing a JVM variant.

### DIFF
--- a/skiko/gradle.properties
+++ b/skiko/gradle.properties
@@ -4,9 +4,6 @@ kotlin.code.style=official
 kotlin.mpp.applyDefaultHierarchyTemplate=false
 kotlin.mpp.enableCInteropCommonization=true
 kotlin.js.yarn.lock.file=false
-# skiko-graphite is included on Linux so CI can publish its common KMP module metadata.
-# With the default properties below, Graphite has no enabled targets on Linux because its JVM target is macOS-only.
-kotlin.internal.suppressGradlePluginErrors=NoKotlinTargetsDeclared
 skiko.kotlin.language.version=2.2
 skiko.kotlin.api.version=2.2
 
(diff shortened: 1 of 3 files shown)


skiko/gradle.properties
```diff
@@ -4,6 +4,9 @@ kotlin.code.style=official
 kotlin.mpp.applyDefaultHierarchyTemplate=false
 kotlin.mpp.enableCInteropCommonization=true
 kotlin.js.yarn.lock.file=false
+# skiko-graphite is included on Linux so CI can publish its common KMP module metadata.
+# With the default properties below, Graphite has no enabled targets on Linux because its JVM target is macOS-only.
+kotlin.internal.suppressGradlePluginErrors=NoKotlinTargetsDeclared
```
MatkovIvan: Should we add it only to submodule instead of project-wide?

hub-bla: We can, but I requires a more of a workaround: 
1. Do not apply the kmp plugin at the start
2.  Then apply it like:
```kotlin
extra["kotlin.internal.suppressGradlePluginErrors"] = "NoKotlinTargetsDeclared"
apply(plugin = "org.jetbrains.kotlin.multiplatform")
val kotlin = extensions.getByType<KotlinMultiplatformExtension>()
```
Found adding it globally more appealing. 

BTW, this error itself do not make our CI fail at all, so it does not require a suppression

MatkovIvan: Isn't a separate `gradle.properties` in a subfolder with this one line enough?

hub-bla: It's not a build root so I'm assuming there cannot be a separate `gradle.properties`

hub-bla: Moved to Graphite only using the above workaround

MatkovIvan (approved):

PR #1261, opened 2026-08-11
Add Graphite Vulkan

hub-bla:
This PR introduces the Skia Native Graphite Vulkan API necessary to implement the renderers.

Additional changes that are still required:
- Release Skia with Vulkan build for Windows and fix the `SkOpts::RGBA_to_BGRA` function-pointer issue caused by Graphite being a separate module (Will create it soon)
- Add the Vulkan SDK to TeamCity on Windows (Or try to get rid of Vulkan SDK)

Fixes - [SKIKO-1167](https://youtrack.jetbrains.com/issue/SKIKO-1167)


ohlunasa:
Hi! Is there any possibility of Vulkan being available on the macOS platform, or is that a limitation posed by Skia?

.github/actions/setup-prerequisites/action.yml
```diff
@@ -19,6 +19,18 @@ runs:
       shell: bash
       run: brew install llvm
 
+    - name: Install Vulkan SDK
```
igordmn, line 35: Worth to cache it via `actions/cache@v4` as it takes 1 minute and involves network.

skiko/buildSrc/src/main/kotlin/SkikoProjectContext.kt
```diff
@@ -104,7 +105,7 @@ fun SkikoProjectContext.declareSkiaTasks() {
 
             val skiaReleaseTag = project.skiaVersion(target)
 
-            val skiaBaseUrl = "https://github.com/JetBrains/skia/releases/download/$skiaReleaseTag"
+            val skiaBaseUrl = "https://github.com/hub-bla/skia/releases/download/$skiaReleaseTag"
```
igordmn: (blocker) Should be changed to JetBrains before merging

skiko/skiko-graphite/src/commonMain/kotlin/org/jetbrains/skia/gpu/graphite/BackendTexture.kt
```diff
@@ -34,6 +36,38 @@ class BackendTexture internal constructor(ptr: NativePointer) : Managed(ptr, _Fi
...
+        /**
+         * Creates a Graphite backend texture that wraps an existing `VkImage`.
+         *
+         * The image memory is assumed to be managed by the client (or by the driver, as is the case
+         * for swapchain images), so no allocation info is passed to Skia.
+         */
+        fun makeVulkan(
+            width: Int,
+            height: Int,
+            imagePtr: NativePointer,
```
igordmn: Let's sort args the same way as in [the Skia API](https://github.com/JetBrains/skiko/blob/621bf661b095fe2dd276c1d7a3acfa03b849293c/skiko/skiko-graphite/src/jvmMain/cpp/common/BackendTexture.cc#L46)

skiko/skiko-graphite/src/commonMain/kotlin/org/jetbrains/skia/gpu/graphite/VulkanTypes.kt
```diff
@@ -0,0 +1,144 @@
...
+ * @param imageUsageFlags Vulkan image usage flags (`VkImageUsageFlags`).
+ * @param sampleCount Number of samples per pixel (1 for non-MSAA).
+ * @param mipmapped `true` if mipmaps are present/allocated for this texture.
+ * @param flags Vulkan image creation flags (`VkImageCreateFlags`).
+ * @param imageTiling Vulkan image tiling layout (`VkImageTiling`).
+ * @param sharingMode Vulkan queue sharing mode (`VkSharingMode`).
+ * @param aspectMask Vulkan image aspect mask (`VkImageAspectFlags`).
+ */
+@ExperimentalSkikoApi
+class VulkanTextureInfo(
```
igordmn, line 124: Let's sort the args the same way as in [the Skia constructor](https://github.com/JetBrains/skiko/blob/621bf661b095fe2dd276c1d7a3acfa03b849293c/skiko/skiko-graphite/src/jvmMain/cpp/common/VulkanUtils.hh#L40) (and store them in the array in the same order)

skiko/skiko-graphite/src/jvmMain/cpp/common/VulkanUtils.hh
```diff
@@ -0,0 +1,52 @@
...
+
+inline skgpu::graphite::VulkanTextureInfo skikoVulkanTextureInfoFromIntArray(
+        JNIEnv* env,
+        jintArray textureInfoValues) {
+    jint values[8] = {0};
+    if (textureInfoValues) {
+        env->GetIntArrayRegion(textureInfoValues, 0, 8, values);
+    }
+
+    return skgpu::graphite::VulkanTextureInfo(
```
igordmn, line 42: It is better to add a link to the `packToIntArray` call, so it is clear where the indices come from.

skiko/skiko-graphite/build.gradle.kts
```diff
@@ -44,10 +37,13 @@ val graphiteDependencies: SkikoDependencyScope.() -> Unit = {
...
             linux {
                 staticSkiaLibs("skia_graphite_ext")
+                dynamicSystemLibs("vulkan")
+                compilerFlags("-DSK_VULKAN", "-DSK_USE_INTERNAL_VULKAN_HEADERS")
             }
             windows {
-                staticSkiaLibs("skia_graphite_dawn_ext", "dawn_combined")
-                dynamicSystemLibs("onecore_apiset", "dxguid")
+                staticSkiaLibs("skia_graphite_ext")
+                dynamicSystemLibs("vulkan-1", "onecore_apiset", "dxguid")
```
igordmn: (potential blocker) What if it is not available? Will Skiko be able to work?

We had [a situation](https://github.com/JetBrains/skiko/pull/811) where OpenGL.dll isn't available on a CI Windows machine that led to an error on `LibraryLoad`. The same can happen on Linux.

Can be tested by manually removing the Vulcan dll from the system.

If it crashes, dynamic loading of dll/so should help (it makes the code ugly though)

skiko/skiko-graphite/src/nativeJsMain/cpp/GraphiteContext.cc
```diff
@@ -36,10 +49,46 @@ SKIKO_EXPORT KNativePointer org_jetbrains_skia_gpu_graphite_GraphiteContext__1nM
...
     skgpu::graphite::InsertRecordingInfo info{};
     info.fRecording = reinterpret_cast<skgpu::graphite::Recording*>(recordingPtr);
+
+    std::vector<skgpu::graphite::BackendSemaphore> waitSemaphores;
+    if (waitSemaphoresPtrs && waitSemaphoresCount > 0) {
+        waitSemaphores.reserve(waitSemaphoresCount);
+        auto ptrs = reinterpret_cast<KNativePointer*>(waitSemaphoresPtrs);
+        for (KInt i = 0; i < waitSemaphoresCount; ++i) {
+            auto sem = reinterpret_cast<skgpu::graphite::BackendSemaphore*>(ptrs[i]);
+            if (sem) {
```
igordmn: Does it happen when semaphore is closed? It looks better to fail than skip in this case (adding `require` in Kotlin).

skiko/skiko-graphite/src/jvmTest/kotlin/org/jetbrains/skia/gpu/graphite/GraphiteTestHelpers.kt
```diff
@@ -2,30 +2,70 @@ package org.jetbrains.skia.gpu.graphite
...
         _nReleaseMetalObjects(metalObjects[0], metalObjects[1])
     }
+    return context
+}
+
+@OptIn(ExperimentalSkikoApi::class)
+private fun withTestVulkanContext(block: (GraphiteContext) -> Unit) {
+    GraphiteLibrary.load()
+    val vulkanObjects = _nCreateVulkanObjects()
+    if (vulkanObjects.isEmpty()) return
```
igordmn: In case our CI agents have Vulkan, it is better to fail instead of skipping. Otherwise we can miss regressions.

skiko/skiko-graphite/src/commonMain/kotlin/org/jetbrains/skia/gpu/graphite/BackendSemaphore.kt
```diff
@@ -0,0 +1,45 @@
...
+ */
+@ExperimentalSkikoApi
+class BackendSemaphore internal constructor(ptr: NativePointer) : Managed(ptr, _FinalizerHolder.PTR) {
+    companion object {
+        init {
+            GraphiteLibrary.load()
+        }
+
+        /**
+         * Creates a Graphite backend semaphore wrapping a Vulkan `VkSemaphore`.
```
igordmn, line 21: Worth to add that `BackendSemaphore` doesn't take ownership and delete `semaphorePtr` on `close`

igordmn (changes requested):

hub-bla:
Hi @ohlunasa!

Skiko’s main goal is to power Compose Multiplatform. We’re trying to enable the GPU backend that is the default choice for each particular OS first. For macOS, that’s Metal. 

That being said, we definitely want to have a Metal backend in Compose. Vulkan on macOS is not a priority and we need to investigate first whether it’s actually worth enabling.

skiko/skiko-graphite/src/jvmTest/kotlin/org/jetbrains/skia/gpu/graphite/GraphiteTestHelpers.kt
hub-bla: GitHub Actions does not have Vulkan installed so it would fail here. 

I see these options:
- add an appropriate warning
- add a software implementation of Vulkan to the GitHub CI
- add both

wdyt?

skiko/skiko-graphite/build.gradle.kts
hub-bla: >What if it is not available?
It would crash telling that there's vulkan missing. 

I'll switch to load it like the OpenGL. In this case vulkan SDK won't be needed anymore.

skiko/skiko-graphite/src/jvmTest/kotlin/org/jetbrains/skia/gpu/graphite/GraphiteTestHelpers.kt
igordmn: > software implementation

We will be checking not a real thing, and this is probably not a few lines of code.

Warning is okay.

igordmn: Actually, is there a way to install it on Windows/Linux? Won't docker help here?

hub-bla: I think docker won't help here. It looks like these machines do not have drivers for vulkan. That's why as one of the options I suggested trying to install something like https://github.com/google/swiftshader for windows

igordmn: A software implementation of the Vulcan API not developed not by us, and that keeps its contract, may be useful for checking our code. But I would spend only little time trying to setup it, and stop if it requires a lot of effort.

hub-bla: After some investigation:
- Windows has no Vulkan installed on github CI at all, so the vulkan sdk has to stay there
- Swiftshader doesn't provide binaries itself. I needs to be either build from the source or we would need to provide it from community source like: https://github.com/jakoch/install-vulkan-sdk-action
- For Linux there is a different implementation called `lavapipe`, which comes from `mesa` that we already have installed. Unfortunately the `linux-compat` is too old to have it (`lavapipe` is from `mesa` 20.0.3+ while we have a 18.x version there)

So at the current state, it's not straightforward to setup it on the CI. Maybe once we bump the `linux-compat` then we could add the lavapipe there. I tested it on our `linux-amd64` docker and I can confirm that it works.

skiko/skiko-graphite/build.gradle.kts
hub-bla: Done

Vulkan SDK is still needed on CI though because it does not have vulkan there at all

skiko/skiko-graphite/src/jvmMain/cpp/common/VulkanUtils.hh
hub-bla: Done

skiko/skiko-graphite/src/commonMain/kotlin/org/jetbrains/skia/gpu/graphite/BackendTexture.kt
hub-bla: Done

.github/actions/setup-prerequisites/action.yml
hub-bla: Done

skiko/skiko-graphite/src/nativeJsMain/cpp/GraphiteContext.cc
hub-bla: Done

skiko/skiko-graphite/src/commonMain/kotlin/org/jetbrains/skia/gpu/graphite/BackendSemaphore.kt
hub-bla: Done

skiko/skiko-graphite/src/commonMain/kotlin/org/jetbrains/skia/gpu/graphite/VulkanTypes.kt
hub-bla: Done

igordmn:
LGTM. Only the link is needed to be fixed

hub-bla:
@igordmn 
It’s updated now with the fresh Skia release from our fork. I tested it on TeamCity as well. It turns out that the debug build needs some additional libraries from Skia which I added as well. Everything is ready now :)

igordmn (approved):

.github/actions/setup-prerequisites/action.yml
```diff
@@ -19,6 +19,32 @@ runs:
       shell: bash
       run: brew install llvm
 
+    - name: Cache Vulkan SDK
+      if: runner.os == 'Windows'
+      id: vulkan-sdk-cache
+      uses: actions/cache@v4
+      with:
+        path: C:\VulkanSDK\1.4.357.0
+        key: vulkan-sdk-${{ runner.os }}-${{ runner.arch }}-1.4.357.0
```
MatkovIvan: Repeating `1.4.357.0` version (4 times in this file) probably worth extracting into a parameter

skiko/skiko-graphite/src/jvmMain/cpp/common/VulkanLibrary.cc
```diff
@@ -0,0 +1,65 @@
...
+
+namespace {
+
+void* vulkanLibrary = nullptr;
+
+bool loadVulkanLibrary() {
+#if defined(_WIN32)
+    vulkanLibrary = LoadLibrary("vulkan-1.dll");
+#else
+    vulkanLibrary = dlopen("libvulkan.so.1", RTLD_LAZY | RTLD_LOCAL);
```
MatkovIvan: `.1` is a bit suspicious here, do we really need such a suffix here?

MatkovIvan (approved):

.github/actions/setup-prerequisites/action.yml
hub-bla: Done!

PR #1262, opened 2026-08-11
Merge the hierarchies of `Redrawer` and `ContextHandler`

m-sasha:
Revert https://github.com/JetBrains/skiko/pull/1234 and do a straightforward merge of the hierarchies instead.

I tried to split the refactoring into small(ish) steps, and I recommend reviewing commit by commit.

Some minor issues/tasks related to this refactoring (or discovered while doing it) that I want to handle in a follow-up PR:
1. Move `AwtRenderer.isDisposed` and `checkDisposed` up into `Renderer`.
2. Fix `MacOsMetalRenderer.dispose()` and `MacOsOpenGLRenderer.dispose()` (these are in `macosMain`, not `awtMain`).
3. Get rid of `LayerDrawScope.Renderer.draw()`
4. Remove `ContextBasedRenderer.name`.


igordmn:
> Revert https://github.com/JetBrains/skiko/pull/1234 and do a straightforward merge of the hierarchies instead.

What does this PR besides reverting [frame dispatching](https://github.com/JetBrains/skiko/pull/1234/changes/654f268373f68a7a32159aefcdbf59d96b159ab6) extraction? I.e. why revert it fully, not only the second commit with the next small fixes?

Generally, I see the extraction of frame dispatching (including live resize logic, it also drives the frames) as a good thing, it was unfortunate that it was done only partially. I see 3 ways:
1. Revert it and never extract again
2. Revert it and extract properly
3. Don't revert it, finish the extraction.

I don't like p.1, because the redrawers (aka renderers) still contain 2 completely different responsibilities with a lot of code. And this was the goal of the second commit.

m-sasha:
> What does this PR besides reverting [frame dispatching](https://github.com/JetBrains/skiko/pull/1234/changes/654f268373f68a7a32159aefcdbf59d96b159ab6) extraction? I.e. why revert it fully, not only the second commit with the next small fixes?

I didn't feel confident that I understand the entirety of that PR to make changes to it.
Also see my comments on that PR; they don't apply to this one obviously.

> Generally, I see the extraction of frame dispatching (including live resize logic, it also drives the frames) as a good thing, it was unfortunate that it was done only partially.

Then it should be done completely, in a separate PR. Why are we putting a half-done change in an unrelated, otherwise straightforward PR?

> I see 3 ways:
> 
> 1. Revert it and never extract again
> 2. Revert it and extract properly
> 3. Don't revert it, finish the extraction.
> 
> I don't like p.1, because the redrawers (aka renderers) still contain 2 completely different responsibilities with a lot of code. And this was the goal of the second commit.

I don't feel confident doing 3, so I suggest 2.

igordmn:
> I don't feel confident doing 3, so I suggest 2.

Let's discuss with Ivan. If we can't do 3 soon, I am also for 2, as keeping the code in a coupled state leads to long-term pain (it is what we had with `ContextHandler`'s)

igordmn:
> I didn't feel confident that I understand the entirety of that PR to make changes to it.

The first commit is straightforward, and I have verified it myself and with AI, but it will be time-consuming to do it again. If we decide to revert, better to revert only the second commit 

> Also see my comments on that PR; they don't apply to this one obviously.

These are about the second commit

igordmn:
> If we can't do 3 soon

I have a theory that it can be achieved by removing `FrameHost` and inverting the calls `Renderer -> FrameHost` to `OnScreenRedrawer -> Renderer`

m-sasha:
> The first commit is straightforward, and I have verified it myself and with AI, but it will be time-consuming to do it again. If we decide to revert, better to revert only the second commit

I verified mine with AI too.

To add, what this PR does that #1234 doesn't:
1. Fold the native Renderer/ContextHandler counterparts too.
2. Rename `Redrawer` to `Renderer`.
3. Remove `ContextHandler` completely.

igordmn:
> I verified mine with AI too.

It is still better to not do it twice, it requires review and adds hidden mistakes.

> Fold the native Renderer/ContextHandler counterparts too.

> Remove ContextHandler completely.

We can do this additionally

> Rename Redrawer to Renderer.

It is better to do separately (but not so much better, it still can be done with additional small changes PR)

MatkovIvan:
> To add, what this PR does that #1234 doesn't

That PR had reduced scope intentionally - it was agreed with @igordmn that the entire refactor is too large for a single PR. All mentioned changes (in a bit different form) are in my refactor branch already, but not yet landed

> Don't revert it, finish the extraction

It's the way to go. #1273 is the next part

m-sasha:
> That PR had reduced scope intentionally

I think the things I mentioned should have been in the scope, while other things in that PR should not have been.

> 1. Fold the native Renderer/ContextHandler counterparts too.

If you're folding the Kotlin classes, not folding the corresponding natives unfolded leaves the code in a strange state.

> 2. Rename Redrawer to Renderer.

Can be delayed to a later PR, but it's just a renaming.

> 3. Remove ContextHandler completely.

If you're merging the hierarchies, why leave a loose end in `macosMain`?

MatkovIvan:
> why leave a loose end in macosMain?

It's not missed, it's already done and waits in the queue. It's just not a part of the PR to reduce the diff and limit the scope to AWT only. It can be easily cherry-picked into #1273 if you want

PR #1264, opened 2026-08-12
Fix `equals`, `hashCode` and `copy` of `SkiaLayerProperties`

m-sasha:
Fix `equals`, `hashCode` and `copy` of `SkiaLayerProperties` to include new `frameBuffering` and `gpuResourceCacheLimit` properties.


skiko/src/jvmMain/kotlin/org/jetbrains/skiko/SkiaLayerProperties.kt
igordmn, line 41: It is better to add args into `copy` too

igordmn (approved):

skiko/src/jvmMain/kotlin/org/jetbrains/skiko/SkiaLayerProperties.kt
m-sasha: of course

igordmn (approved):

PR #1266, opened 2026-08-13
Decouple AWT redrawers from Ganesh object creation

hub-bla:
AWT native platform code currently creates Ganesh `GrDirectContext` and `GrBackendRenderTarget` objects and passes opaque pointers to Kotlin. 

This PR moves that responsibility to the Kotlin bindings

Fixes - [SKIKO-1168](https://youtrack.jetbrains.com/issue/SKIKO-1168)


m-sasha:
I don't have the expertise to review this, unfortunately.

I can only express my worry that we're now piling changes on top of #1234 which I hope to revisit with Ivan and Igor soon. But maybe it's not a blocker; up to @igordmn

skiko/src/awtMain/cpp/windows/AngleRedrawer.cc
```diff
@@ -207,30 +194,28 @@ extern "C"
...
-            angleDevice->backendContext->fFunctions.fGetIntegerv(GR_GL_FRAMEBUFFER_BINDING, &buffer);
-
-            GrGLFramebufferInfo glInfo = { static_cast<unsigned int>(buffer), GR_GL_RGBA8 };
-            GrBackendRenderTarget renderTarget = GrBackendRenderTargets::MakeGL(width,
-                                                                                height,
-                                                                                0,
-                                                                                8,
-                                                                                glInfo);
-            GrBackendRenderTarget *pRenderTarget = new GrBackendRenderTarget(renderTarget);
+            auto glViewport = reinterpret_cast<void (APIENTRY *)(GLint, GLint, GLsizei, GLsizei)>(eglGetProcAddress("glViewport"));
```
igordmn, line 200:
```suggestion
            static auto glViewport = reinterpret_cast<void (APIENTRY *)(GLint, GLint, GLsizei, GLsizei)>(eglGetProcAddress("glViewport"));
```

skiko/src/awtMain/cpp/windows/AngleRedrawer.cc
```diff
@@ -207,30 +194,28 @@ extern "C"
...
+            auto glGetIntegerv = reinterpret_cast<void (APIENTRY *)(GLenum, GLint *)>(eglGetProcAddress("glGetIntegerv"));
```
igordmn, line 201:
```suggestion
            static auto glGetIntegerv = reinterpret_cast<void (APIENTRY *)(GLenum, GLint *)>(eglGetProcAddress("glGetIntegerv"));
```

igordmn:
> I can only express my worry that we're now piling changes on top of https://github.com/JetBrains/skiko/pull/1234 which I hope to revisit with Ivan and Igor soon.

This looks unrelated to the Redrawer extraction changes, but because there are a lot of them nearby, it still might conflict with them. We can wait for the Redrawer decision that we plan in the next week (should we revert them or quickly fix).

If this PR isn't urgent, we can wait for that. Otherwise we can merge, and then fix conflicts in the Redrawer revert if we decide to revert.

skiko/src/awtMain/cpp/windows/AngleRedrawer.cc
```diff
@@ -4,15 +4,8 @@
 #include <jawt_md.h>
```
igordmn, line 4: AI found that we have a stale `stubs.cc`, which technically should be changed in this PR. But because it is already stale, let's move this to an YouTrack issue and add a comment in `stubs.cc`:
```
// TODO probably this file is out of date, see https://youtrack.jetbrains.com/issue/SKIKO-1172/stubs.cc-is-out-of-date
```

igordmn (commented):
LGTM. Don't approve yet, until we decide to merge it before the Redrawer refactor decision

hub-bla:
> If this PR isn't urgent, we can wait for that. Otherwise we can merge, and then fix conflicts in the Redrawer revert if we decide to revert.

It's not that urgent. It's a prepartion required for extracting the skiko "rendering-glue" into a separate module.

skiko/src/awtMain/cpp/windows/AngleRedrawer.cc
hub-bla: AFAIR, this file was related to the Graal experiments. Its purpose was probably to handle cpp bindings that are conditionally compiled for specific platforms using preprocessor macros.

Like here for example: https://github.com/JetBrains/skiko/blob/master/skiko/src/jvmMain/cpp/common/DirectContext.cc#L35-L51

So when creating a binary lets's say on macOS, the build/linking of Graal would fail because the above function exists on the Kotlin side but it will be missing the C++ side without the stub. With regular JNI we can kind of get away with this. The missing function only becomes a problem at runtime if Kotlin actually tries to call it

We need to decide whether we want to sync the stubs file with the current state of the repo or remove it and refactor some of the bindings to the structure that was done in `skiko-graphite`
For example:
https://github.com/JetBrains/skiko/blob/master/skiko/skiko-graphite/src/jvmMain/cpp/common/GraphiteContext.cc#L21-L38

igordmn: > We need to decide whether we want to sync the stubs file

Now there are broken, so decision at the moment and in this PR - don't sync them.

As for building with Graal - if it just stubs that fail it, I would restore them and add a GitHub check. Otherwise, if it is difficult to support, or it is not required now - remove them.

igordmn (approved):

skiko/src/awtMain/kotlin/org/jetbrains/skiko/graphicapi/InternalDirectXApi.kt
```diff
@@ -7,6 +7,14 @@ import org.jetbrains.skiko.Library
 import org.jetbrains.skiko.hostOs
 import org.jetbrains.skiko.isVideoCardSupported
 
+@JvmInline
+internal value class DxgiFormat(val value: Int) {
+    companion object {
+        val R8G8B8A8_UNORM = DxgiFormat(28)
+        val B8G8R8A8_UNORM = DxgiFormat(87)
```
MatkovIvan, line 15: Please add a comment with link to source of these magic numbers

MatkovIvan (approved):

PR #1267, opened 2026-08-13
Pace SkiaSwingLayer repaints with the JBR FramePacing service

rock3r:
A continuously invalidating scene rendered through `SkiaSwingLayer` (e.g. a Compose draw-phase animation in SwingGraphics mode) calls `repaint()` as fast as the EDT can complete paint cycles, rendering far above the display refresh rate and burning GPU on frames that are never shown. This adds an opt-in frame pacer that gates invalidation-driven repaints on display refresh ticks from the JetBrains Runtime `FramePacing` service.

> [!NOTE]
> Draft until [JetBrainsRuntimeApi#31](https://github.com/JetBrains/JetBrainsRuntimeApi/pull/31) lands and jbr-api 1.11 is published on Maven Central. The reflective `FramePacingService`/`JbrFramePacingApi` seam exists only to avoid a compile-time dependency on the unreleased API; once 1.11 is available it will be replaced with a direct `JBR.getFramePacing()` dependency and removed from this PR.

## Changes

* New `SkiaSwingLayer.needRender()`: requests a repaint like `repaint()`, but paced to the display refresh when pacing is available. Plain `repaint()` is untouched — it is thread-safe, region-carrying, and called by Swing internals, and the pacer needs it as the unpaced primitive.
* New internal `SwingRepaintPacer`: a `FrameDispatcher` frame loop that repaints a requested frame immediately, then suspends until the next FramePacing tick — a continuous stream of requests coalesces to one repaint per tick, while a request arriving when the loop is idle paints with no added latency. The tick wait is bounded by `withTimeout` (~3 refresh periods) so a dead pacing clock cannot stall the loop.
* The display id is re-resolved from the layer's `GraphicsConfiguration` once per frame and the tick subscription re-created when it changes, per the jbr-api javadoc guidance; a display that refuses a subscription is cached and not retried until the layer moves to another display.
* Gated behind `skiko.swing.frame.pacing` (default **false**). When the property is off, the JBR API or service is absent, or the display cannot be paced, behavior is unchanged: requests degrade to plain `repaint()`.
* 10 unit tests drive the pacer through the `FramePacingService` seam with a fake tick source.

## Measured results

Plain-Swing `SkiaSwingLayer` spinner scene, invalidation-driven via `needRender()`, on JBR builds providing the FramePacing service:

| | renders/sec, unpaced → paced | power |
|---|---|---|
| macOS, 120 Hz ProMotion | ~500–570 → **120–121, locked** | scene too cheap to move the GPU off min clock |
| Linux, 1920x1080@120 (UHD 630) | 325.3 → **120.0, locked** | package **11.22 → 3.61 W (−68%)** |

Below the refresh rate pacing is a no-op by design: a render-bound scene (71.5 renders/sec unpaced) is not degraded (72.3 paced).

### DIFF
--- a/skiko/src/awtMain/kotlin/org/jetbrains/skiko/swing/SkiaSwingLayer.kt
+++ b/skiko/src/awtMain/kotlin/org/jetbrains/skiko/swing/SkiaSwingLayer.kt
@@ -74,6 +74,8 @@ open class SkiaSwingLayer(
     private val redrawer: SwingRedrawer?
         get() = redrawerManager.redrawer
 
+    private var repaintPacer: SwingRepaintPacer? = null
+
     val renderApi: GraphicsApi
         get() = redrawerManager.renderApi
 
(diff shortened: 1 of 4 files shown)


rock3r:
Superseded by #1282.

This PR paced repaints using the JBR `FramePacing` service, which made it depend on a runtime API that is not shipping. #1282 does the same work with clocks implemented inside Skiko, so pacing works on any JVM and there is no runtime dependency at all.

The pacer design here — `needRender()`, paint-first/wait-after, the bounded tick wait — carried over unchanged; only the tick source differs.

PR #1272, opened 2026-08-18
Return null from `ColorFilter.makeBlend` for no-op blends

MatkovIvan:
`SkColorFilters::Blend` returns nullptr whenever the requested blend would leave the filtered color untouched, and collapses such no-ops away rather than allocating a filter: `DST` with any color, a fully transparent color with `SRC_OVER`/`DST_OVER`/`DST_OUT`/`SRC_ATOP`/`XOR`/`DARKEN`, and an opaque color with `DST_IN`. `makeBlend` wrapped that pointer unconditionally, so `RefCnt`'s constructor rejected it with "Can't wrap nullptr" and a legal request became a crash.

Return null instead, matching Skia's own signature and the convention already used by the other bindings over `nullptr`-returning factories (`Path.makeCombining`, `FontMgr.matchFamilyStyle`, `Surface.makeImageSnapshot`). Callers read `null` as "apply no filter", which is what Skia means by it.

The return type changes from `ColorFilter` to `ColorFilter?`, which is source- incompatible for callers that use the result without a null check.

[CMP-9226](https://youtrack.jetbrains.com/issue/CMP-9226) Crash when using a transparent tint with SrcAtop (Can't wrap nullptr)
[CMP-10677](https://youtrack.jetbrains.com/issue/CMP-10677) iOS: process aborts from `MetalRedrawer.draw` when a `Canvas` draws a gradient `Brush` under a per-draw `BlendModeColorFilter`

Requires Compose-side adoption after the release

### DIFF
--- a/skiko/src/commonMain/kotlin/org/jetbrains/skia/ColorFilter.kt
+++ b/skiko/src/commonMain/kotlin/org/jetbrains/skia/ColorFilter.kt
@@ -28,9 +28,18 @@ class ColorFilter : RefCnt {
             }
         }
 
-        fun makeBlend(color: Int, mode: BlendMode): ColorFilter {
+        /**
+         * Returns a filter that blends [color] into the filtered color with [mode], or null when
+         * that combination leaves the color untouched. Skia collapses such no-ops away, so a null
+         * result means "apply no filter": [BlendMode.DST] with any color, a fully transparent
+         * [color] with [BlendMode.SRC_OVER], [BlendMode.DST_OVER], [BlendMode.DST_OUT],
+         * [BlendMode.SRC_ATOP], [BlendMode.XOR] or [BlendMode.DARKEN], and an opaque [color] with
+         * [BlendMode.DST_IN].
+         */
+        fun makeBlend(color: Int, mode: BlendMode): ColorFilter? {
             Stats.onNativeCall()
-            return ColorFilter(_nMakeBlend(color, mode.ordinal))
+            val ptr = _nMakeBlend(color, mode.ordinal)
+            return if (ptr == NullPointer) null else ColorFilter(ptr)
         }
 
         fun makeMatrix(matrix: ColorMatrix): ColorFilter {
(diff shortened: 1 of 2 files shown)


hub-bla (approved):

skiko/src/commonMain/kotlin/org/jetbrains/skia/ColorFilter.kt
```diff
@@ -28,9 +28,18 @@ class ColorFilter : RefCnt {
...
-        fun makeBlend(color: Int, mode: BlendMode): ColorFilter {
+        /**
+         * Returns a filter that blends [color] into the filtered color with [mode], or null when
+         * that combination leaves the color untouched. Skia collapses such no-ops away, so a null
+         * result means "apply no filter": [BlendMode.DST] with any color, a fully transparent
+         * [color] with [BlendMode.SRC_OVER], [BlendMode.DST_OVER], [BlendMode.DST_OUT],
+         * [BlendMode.SRC_ATOP], [BlendMode.XOR] or [BlendMode.DARKEN], and an opaque [color] with
+         * [BlendMode.DST_IN].
+         */
+        fun makeBlend(color: Int, mode: BlendMode): ColorFilter? {
```
igordmn, line 39: > Requires Compose-side adoption after the release

Please prepare the changes before merging. There are different ways of doing this - throw, ignore, change the type of the using functions (`ColorFilter.asSkiaColorFilter`), etc

MatkovIvan: > Please prepare the changes before merging

It was done even before opening this PR

> There are different ways of doing this - throw, ignore, change the type of the using functions (ColorFilter.asSkiaColorFilter), etc

"etc" is my variant here 😅

igordmn: > It was done even before opening this PR

Can you make a PR or show a branch?

MatkovIvan: https://github.com/JetBrains/compose-multiplatform-core/pull/3319

igordmn (approved):

PR #1273, opened 2026-08-19
AWT frame-driving extraction

MatkovIvan:
Finishes the render-stack extraction started in #1234. Four commits, meant to be read one by one:

1. **Post-merge review fixes.** The non-controversial fixes from the #1234 threads. No structural change.
2. **Structure + rename.** The shared `Redrawer` interface moved to `macosMain`, the only source set where two implementations actually share one type. On AWT the frame loop becomes a plain class held directly by `SkiaLayer`, and the family gets the names proposed in review: per-API backends `MetalRenderer` / `Direct3DRenderer` / `AngleRenderer` / ... with base `AwtRenderer`, and the loop is `FrameDriver` - it decides *when* frames happen, the renderer decides *how*. 
3. **The OpenGL cross-window batches move out of the backends** into the driver layer, as two objects behind one `GLFrameBatch` base.
4. **`FrameHost` is deleted.** Live-resize driving and Metal's early-record dispatcher move into `FrameDriver`. Backends forward exactly three native live-resize events through `LiveResizeListener`; the driver decides what each event means and calls back down through `renderPlatformDrivenFrame` / `requestPlatformDrivenFrame`. The before-shown frame gets one decision point (`BeforeShownFrame`: `NONE` / `RENDER` / `BACKEND`).


skiko/src/commonMain/kotlin/org/jetbrains/skiko/TransparentBackground.kt
```diff
@@ -0,0 +1,11 @@
+package org.jetbrains.skiko
+
+internal fun defaultIsTransparentBackgroundSupported(layer: SkiaLayer): Boolean {
```
igordmn: It is better to move to awtMain.

I would also, just inline it into AWTRenderer

skiko/src/awtMain/kotlin/org/jetbrains/skiko/renderer/AbstractOpenGLRenderer.kt
```diff
@@ -38,6 +38,10 @@ internal abstract class AbstractOpenGLRedrawer(
         glContext?.flush()
     }
 
+    internal fun drawBatchFrame(scope: LayerDrawScope) {
```
igordmn: We can make `drawFrame` public instead.

Because it looks the same and doesn't have batch-specific implementation ("batching" is the property of the caller, not of renderer)

skiko/src/awtMain/kotlin/org/jetbrains/skiko/swing/SkiaSwingLayer.kt
```diff
@@ -63,19 +63,19 @@ open class SkiaSwingLayer(
...
     }
 
-    private val redrawerManager = RedrawerManager<SwingRedrawer>(
+    private val rendererManager = RenderApiFallbackManager<SwingRenderer>(
         properties.renderApi,
-        redrawerFactory = { renderApi, oldRedrawer ->
-            oldRedrawer?.dispose()
-            createSwingRedrawer(swingLayerProperties, renderDelegateWithClipping, renderApi, analytics)
+        factory = { renderApi, oldContext ->
+            oldContext?.dispose()
```
igordmn:
```suggestion
        factory = { renderApi, oldRenderer ->
            oldRenderer?.dispose()
```

skiko/src/awtMain/kotlin/org/jetbrains/skiko/SkiaLayer.awt.kt
```diff
@@ -353,11 +352,11 @@ actual open class SkiaLayer internal constructor(
...
     private var isDisposed = false
 
-    private val redrawerManager = RedrawerManager<Redrawer>(
+    private val frameDriverManager = RenderApiFallbackManager<FrameDriver>(
         defaultRenderApi = properties.renderApi,
-        redrawerFactory = { renderApi, oldRedrawer ->
-            oldRedrawer?.dispose()
-            renderFactory.createRedrawer(this, renderApi, analytics, properties).also {
+        factory = { renderApi, oldContext ->
+            oldContext?.dispose()
```
igordmn:
```suggestion
        factory = { renderApi, oldFrameDriver ->
            oldFrameDriver?.dispose()
```

skiko/src/awtMain/kotlin/org/jetbrains/skiko/renderer/AwtRenderer.kt
```diff
@@ -0,0 +1,162 @@
...
+     * each one.
+     * Written on the EDT, read from platform threads.
+     */
+    @Volatile
+    protected var frameEvents: LiveResizeListener? = null
+        private set
+
+    internal fun attachFrameEvents(events: LiveResizeListener) {
+        frameEvents = events
+    }
```
igordmn:
```suggestion
    protected var liveResizeListener: LiveResizeListener? = null
        private set

    internal fun setLiveResizeListener(events: LiveResizeListener) {
        this.liveResizeListener = liveResizeListener
    }
```

igordmn: A context, just for understanding decisions. Previously we discussed with @m-sasha that:
- renderers/redrawers might be not a good place for live-resizing driving (but we didn't came up with a better solution)
- we hid the logic under Redrawer, not exposing it (but if we extract frame-driving, we have to).

In the current code, I would also want to have live-resizing logic out of renderers, but it is currently simple to have it here, because all the native interaction with native window is here.

igordmn: @m-sasha, in https://github.com/JetBrains/skiko/pull/1234#discussion_r3747894229, you mentioned that live-resize is implementation detail. Is this similar to what I said, or should something be done differently?

As an option we can omit "live" from the name

skiko/src/awtMain/kotlin/org/jetbrains/skiko/renderer/AwtRenderer.kt
```diff
@@ -0,0 +1,162 @@
...
+
+    /**
+     * The driver's listener for platform live-resize events; `null` until [attachFrameEvents] runs.
+     * Native resize callbacks are JNI-bound to this class, so the backend receives them and forwards
+     * each one.
+     * Written on the EDT, read from platform threads.
+     */
+    @Volatile
+    protected var frameEvents: LiveResizeListener? = null
```
m-sasha: The "live-resize" functionality is an implementation detail of how the Metal and Direct3D renderers make sure the window resizes smoothly. It doesn't need to be exposed to outside the renderer.

In fact, I want to try and implement smooth resizing when resizing programmatically too, and if that succeeds, there would perhaps be no need for custom behavior during "live resize".

skiko/src/awtMain/kotlin/org/jetbrains/skiko/renderer/AwtRenderer.kt
```diff
@@ -0,0 +1,162 @@
...
+
+    /**
+     * How this backend produces the frame that must be ready before the window is first shown, so
+     * the window background does not flash. The driver asks before recording anything: [NONE] skips
+     * that frame and its analytics entirely, [RENDER] presents it through [renderFrame], [BACKEND]
+     * calls [renderBeforeShown] so the backend presents it its own way.
+     */
+    enum class BeforeShownFrame { NONE, RENDER, BACKEND }
+
+    open val beforeShownFrame: BeforeShownFrame get() = BeforeShownFrame.RENDER
```
m-sasha: It would be more flexible to let the renderer decide the behavior rather than hardcoding the set of possible behaviors.

skiko/src/awtMain/kotlin/org/jetbrains/skiko/renderer/AwtRenderer.kt
MatkovIvan: The resize mechanics are an implementation detail, and they stay private to the renderers, it is only the 3-event listener here.

The reason for such a shape is that "when frames happen" should become a component that can be replaced without touching the backends. My follow-up work exposes the way for callers to drive frames, and a renderer can be used off-screen, where no live resize exists at all. This works only if the renderer holds no scheduling state or policy of its own.
If the state machine moved back into the renderer, the renderer would instead call up into `FrameDriver` for recording, for pausing the loop, and for the end-of-resize policy - so every backend would still depend on exactly the class the follow-ups replace, and would need changes again.

igordmn: > The "live-resize" functionality is an implementation detail of how the Metal and Direct3D renderers

I see that there can be benefit of keeping it as implementation detail because it hides a complicated logic, but I also see major issues.

We discussed this during its introduction, then we came up to conclusion that it just an easiest way to keep it in renderers, because currently only they have simple access to the native window.

Issues:
1. One more way to drive frames, but from the inside the renderer, not from outside of it:
- (outside) SkiaLayer.resize
- (outside) SkiaLayer.redrawImmediately
- (outside) SkiaLayer.needRedraw - async request
- (was inside, but now outside) FrameDispatcher - needed only to handle the outside async request and throttle to vsync
- (inside) onResize

This PR moves the resize driving outside renderers. It is done by just notifying onResize, not driving the rendering on it.

2. Resizing is semantically bound to an OS window, not to a graphic API.

3. Multiple checks isInLiveResize and races with rendering - we can't move it fully outside for now, because it has native counterpart, but moving only Kotlin code in one place that is separated from rendering simplifies logic in my opinion.

> I want to try and implement smooth resizing when resizing programmatically too,

Do we need to have live resize as implementation details for this?

m-sasha:
> > I want to try and implement smooth resizing when resizing programmatically too,
>
> Do we need to have live resize as implementation details for this?

The dependence on knowing we're "in live-resize" is a crutch. We use it to temporarily disable AWT behaviors and modify our own behaviors.

If I can make programmatic resize smooth it will (probably) mean we won't need this crutch because resizing programmatically is a "one-shot" event. If a one-shot event works then a series of one-shot events will (again, probably) also work.

skiko/src/awtMain/kotlin/org/jetbrains/skiko/renderer/AwtRenderer.kt
igordmn: >  (probably) mean we won't need this crutch

It will be good.

> resizing programmatically is a "one-shot" event. 

Can you describe the approach more?

Currently I see that moving live-resizing logic to the Driver brings a lot of benefits, and we still can remove this after your improvements.

It still can be controlled on the renderer level - just don't call the callbacks, but it might interfere with your changes.

skiko/src/awtMain/kotlin/org/jetbrains/skiko/renderer/MetalRenderer.kt
```diff
@@ -318,9 +300,9 @@ internal class MetalRedrawer(
         }
     }
 
-    override fun syncBounds() = synchronized(drawLock) {
+    override fun syncBoundsFromPlatformComponent() = synchronized(drawLock) {
         check(isEventDispatchThread()) { "Method should be called from AWT event dispatch thread" }
-        if (isHandlingLiveResizeNow) return
+        if (isLiveResizing) return
```
igordmn: Maybe we can remove `isLiveResizing` from here, and move the check to `FrameDriver`

skiko/src/awtMain/kotlin/org/jetbrains/skiko/renderer/GLFrameBatch.kt
```diff
@@ -0,0 +1,138 @@
...
+        // Without clearing we will have a memory leak
+        toRedrawCopy.clear()
+    }
+}
+
+/**
+ * Draws every window under its drawing-surface lock, then swaps the fastest vsync-enabled window
+ * with swap interval 1 and the rest unblocked.
+ */
+internal object LinuxGLFrameBatch : GLFrameBatch<LinuxOpenGLRenderer>() {
```
igordmn: These are tightly coupled with renderers (call "almost private" functions), better to move them in these files.

skiko/src/awtMain/kotlin/org/jetbrains/skiko/renderer/GLFrameBatch.kt
```diff
@@ -0,0 +1,138 @@
...
+    }
+
+    override suspend fun drawAndPresent() {
+        val drawingSurfaces = toRedrawVisible.associateWith { (_, renderer) ->
+            lockLinuxDrawingSurface(renderer.layer.backedLayer)
+        }
+        try {
+            for (window in toRedrawVisible) {
+                val (driver, renderer) = window
+                drawingSurfaces[window]!!.makeCurrent(renderer.context)
```
igordmn: `makeCurrent` should be inside `driver.inFrame`.

I would also move `makeCurrent` into `drawFrame`, then we can move this cycle into the base class

skiko/src/awtMain/kotlin/org/jetbrains/skiko/renderer/GLFrameBatch.kt
```diff
@@ -0,0 +1,138 @@
...
+import org.jetbrains.skiko.unlockLinuxDrawingSurface
+
+/**
+ * The OpenGL backends draw every on-screen window from one cross-window batch per family, waiting
+ * for vsync once per tick instead of once per window: per-window vsync waits would serialize, so
+ * N windows would run at 1/N of the refresh rate. A window enters the batch each time its
+ * [FrameDriver] schedules a frame and leaves it when the tick completes; disposed and hidden
+ * windows are skipped at draw time.
+ */
+internal abstract class GLFrameBatch<R : AbstractOpenGLRenderer> {
```
igordmn, line 14: The connection is complicated:

- `FrameDriver` uses `GLFrameBatch as FrameSchedule`
- `FrameDriver` uses `AWTRenderer`
- `GLFrameBatch ` uses `FrameDriver`
- `GLFrameBatch ` uses `AWTRenderer`

can AI improve that? Personally, I don't see easy ways at the moment.

skiko/src/awtMain/kotlin/org/jetbrains/skiko/renderer/MetalRenderer.kt
MatkovIvan: Moved, it seems it's actually easier

skiko/src/awtMain/kotlin/org/jetbrains/skiko/renderer/GLFrameBatch.kt
MatkovIvan: Moved

skiko/src/awtMain/kotlin/org/jetbrains/skiko/renderer/GLFrameBatch.kt
MatkovIvan: Done

skiko/src/awtMain/kotlin/org/jetbrains/skiko/renderer/GLFrameBatch.kt
MatkovIvan: It doesn't seem that we should touch this *now* (yet/at the moment).

🤖 (Fable 5):
> The cycle is the scheduler-callback shape: the driver hands "when frames happen" to the shared batch, and the batch calls back per window because recording and frame analytics are per-window state that lives in the driver. It is the same contract as `FrameDispatcher` (schedule out, tick back), so I don't think the cycle itself can be removed while cross-window batching is a separate component; putting an interface between the two would rename the edges without removing any.
What this round does reduce: the batch now talks to the renderer through four calls (`drawFrame`, `makeCurrent`, `swapBuffers`, lock/unlock), the draw loop moved into the `GLFrameBatch` base, and each platform object sits in its renderer's file next to the functions it calls.
How I expect it to actually go away: the batch has two halves - scheduling (one vsync wait per tick, then fan out over the scheduled windows) and present policy (swap order under the X11 surface locks, the single `dwmFlush`). The scheduling half is not GL-specific; it is the shape of any shared frame source, so the plan is to merge it into the frame-driving side when frame driving gets a proper display-level tick source. At that point the driver-batch edge disappears because both are the same component, and the renderers keep only their present policy. Moving the batch to the driving side of the seam in this PR is what makes that merge possible without touching the backends again.

MatkovIvan: 2bbc8ca4a97945ab2014803e89ed532a265e1ae5 is a try to extract this in exchange of introducing a new entity. PTAL

### Before
```mermaid
flowchart TD
    SkiaLayer -->|requests frames| FrameDriver
    FrameDriver -->|schedules| GLFrameBatch
    GLFrameBatch -.->|records via driver| FrameDriver
    FrameDriver --> AwtRenderer
    GLFrameBatch --> AwtRenderer
    linkStyle 2 stroke:#E24B4A,color:#E24B4A
```

### After
```mermaid
flowchart TD
    SkiaLayer -->|requests frames| FrameDriver
    FrameDriver -->|scheduleFrame, pause| Scheduler["FrameScheduler"]
    Scheduler -->|runs frames through| Producer
    FrameDriver --> Producer[FrameProducer]
    Scheduler --> AwtRenderer
    Producer --> AwtRenderer
```

skiko/src/awtMain/kotlin/org/jetbrains/skiko/renderer/AwtRenderer.kt
```diff
@@ -0,0 +1,162 @@
...
+    }
+
+    protected fun checkDisposed() {
+        check(!isDisposed) { "${this.javaClass.simpleName} is disposed" }
+    }
+
+    abstract val renderInfo: String
+
+    /**
+     * Renders and presents exactly one frame at [scope]'s size. [immediate] selects the synchronous-redraw
```
m-sasha: > -redraw

I'd drop it, as it's not just about *re*draws.

skiko/src/awtMain/kotlin/org/jetbrains/skiko/renderer/AwtRenderer.kt
```diff
@@ -0,0 +1,162 @@
...
+     * variant.
+     *
+     * Throwing [org.jetbrains.skiko.RenderException] means the frame failed, and makes
+     * [SkiaLayer.inDrawScope] fall back to the next render API.
+     */
+    abstract suspend fun renderFrame(scope: LayerDrawScope, immediate: Boolean)
```
m-sasha: We can make `LayerDrawScope` a receiver everywhere. It's more idiomatic and would save lots of `with(scope)` lines.

skiko/src/awtMain/kotlin/org/jetbrains/skiko/renderer/AwtRenderer.kt
```diff
@@ -0,0 +1,162 @@
...
+    }
+
+    /**
+     * The driver's listener for platform live-resize events; `null` until the [FrameDriver]
+     * installs itself. Native resize callbacks are JNI-bound to this class, so the backend
+     * receives them and forwards each one.
+     * Written on the EDT, read from platform threads.
+     */
+    @Volatile
+    internal var liveResizeListener: LiveResizeListener? = null
```
m-sasha, line 87: Again, I think the generic handling of "live-resize" is not a good idea. It:
1. Adds an API which on the face of it renderers need to call. How does a developer know that `WindowsOpenGLRenderer` does not need to call this listener?
2. Adds unnecessary indirection and coupling.
3. Adds unneeded genericity and makes renderer-specific changes difficult.

The "live-resize" thing is not a feature. It's a hack, and we're building a whole infrastructure around it.

skiko/src/awtMain/kotlin/org/jetbrains/skiko/renderer/MetalRenderer.kt
```diff
@@ -227,13 +195,7 @@ internal class MetalRedrawer(
...
             invokeOnEventThreadAndWait {
                 if (isDisposed) return@invokeOnEventThreadAndWait
-                frameHost?.inForcedSizeFrame(Dimension(width, height)) { scope ->
-                    if (!isDisposed) {  // may be disposed in user code, during `update`
-                        // The present must run on the AppKit main thread to join the resize transaction, so
-                        // only record here; `finishFrameSync` presents below on the AppKit main thread
-                        performFrame(scope, finishFrame = false)
-                    }
-                }
+                liveResizeListener?.onLiveResizeFrame(width, height, isResizeFrame = true)
```
m-sasha: Why is `isResizeFrame` always `true`? We need to pass the value from native code.

m-sasha: This is a good example where the generic nature of our handling of live-resize makes us do unnecessary things.

Metal doesn't care whether it's a resize frame, but the API forces us to either lie (like in the current code) or write unnecessary code which could then confuse the reader (they would ask - why do we need to pass `isResizeFrame` to only ignore it).

m-sasha (changes requested):
We are again mixing simple, uncontroversial, changes with complex changes that need discussion.

I would prefer if we first had a PR that completes the merging of `ContextHandler`s and `Redrawer`s:
1. Merge the native counterparts (this PR doesn't seem to do that).
2. Rename "Redrawer" to "Renderer".
3. Keep a simple hierarchy of Renderers that do all the work as before, without separation into "FrameDriver", "FrameProducer", "FrameScheduler" etc.)

After that it would be easier to discuss and *understand* further changes.

skiko/src/awtMain/kotlin/org/jetbrains/skiko/renderer/AwtRenderer.kt
igordmn: > The "live-resize" thing is not a feature. It's a hack, and we're building a whole infrastructure around it.

I agree it is a hack, but I don't agree that we are building infrastructure around it. We are building infrastructure about frame driving, and live-resizing just comes with it, not moved individually.

How would you move frame driving, keeping live resizing private? Live resizing by itself is frame driving.

On the other hand, I don't see an issue in exposing some things to a correct client like `FrameDriver`, the issue would be exposing them to a wrong client, like `SkiaLayer` or `ComposePanel`

m-sasha: > I don't agree that we are building infrastructure around it. We are building infrastructure about frame driving, and live-resizing just comes with it, not moved individually.

Yes, the model/architecture/separation in this PR forces us to do this. Maybe it's a hint that the model is too generic/over-engineered.

igordmn: Or it is a hint that it is not where this hack should be in the first place, this was my argument in the PR introducing it, I didn't want to be it in the redrawers, and tried to refactor (that refactor was similar to this). 

I consider it is in a wrong place because it:
- drives the frames, but it shouldn't, it should only react on driving. You can argue that existing `FrameDispatcher`  already did that, but I argue that it drives only in response of external `needRedraw` and that it was also not a good place to have it in Redrawer. Further adding similar logic to `Redrawer` makes individual redrawers hard to understand. I lost track of what happens in Direct3D/Metal redrawers because of live-resizing, but this PR makes things more clear - what piece should do what.
- has nothing to do with rendering, it is more coupled with a window itself

igordmn: > Direct3D/Metal redrawers because of live-resizing, but this PR makes things more clear - what piece should do what.

Refactoring within Direct3D/Metal redrawers and splitting each into 2 different pieces (one of them is driver, another is renderer) would also make things simpler for me - it would have 2 isolated features separately, not inter-coupled across 2 files (Kotlin + native) over 1000 lines that hard to wrap mind around at once.

So, this would be another approach to refactor, it is less intrusive as well (but eventually it could still lead to this PR state)

igordmn: Considering all this, live-resizing itself doesn't look to me as a thing that should be taken in favor or against this PR. I suggest to just compare - should we keep FrameDispatcher's within Redrawer's or not. If we should - live resizing just stays where it is now. If not - than live-resizing should move with it.

To make things simple we can decide on a state that was before live-resizing.

skiko/src/awtMain/kotlin/org/jetbrains/skiko/renderer/MetalRenderer.kt
MatkovIvan: Moved to Direct3DRenderer's field

m-sasha: It's now better from the point of view of the API, but worse from an implementation point of view: holding something in a property that is not really a property of the the redrawer but a call parameter.

If we're going to keep the live-resize machinery I'd prefer to pass it (with the correct value) after all, with a comment that it's currently unused in `MetalRenderer`.

skiko/src/awtMain/kotlin/org/jetbrains/skiko/renderer/AwtRenderer.kt
m-sasha: > Or it is a hint that it is not where this hack should be in the first place, this was my argument in the PR introducing it, I didn't want to be it in the redrawers

You are trying to achieve a separation of concerns that is good in theory, but doesn't work well in practice. The graphics API implementations differ in too many aspects, so when you try to extract these aspects you either end up having 2 or more objects to support the same graphics API or you end up with "generic" objects that need to support all possible behaviors, making them more complicated than necessary.

> I lost track of what happens in Direct3D/Metal redrawers because of live-resizing, but this PR makes things more clear - what piece should do what.

I have the opposite experience. For me the live-resizing-in-each-redrawer was clear, and the way things are in this PR is confusing.

m-sasha: Going forward, my concern is that when I will need to add a new hack to an implementation of a graphics API that involves functionality that is (in this PR) under the responsibility of the generic classes, instead of a simple local hack, I would have to change the API and behavior of these objects. This would be more difficult, and if not done very carefully could affect other graphics API implementations.

igordmn: > You are trying to achieve a separation of concerns that is good in theory, but doesn't work well in practice.

If we look at this PR in practice, I see that it helps in understandability - I can wrap my mind around it, because there is no one big coupled thing.

Multiple abstractions (like "FrameDriver", "FrameProducer", "FrameScheduler") I don't see as minuses if they have robust independent semantics - it help reducing coupling and wrapping a mind around them. Without them, they still exist in your mind for understanding what is going on, they just don't have a name and intercoupled with each other.

I see your point that you think we need different driving or live resizing for different renderers, but I think it might was a mistake, and the current state of reality is just the result of this mistake. I can't figure out any reason why would I write it render-dependent if I would write it from scratch.

About generalization - we have to draw a "general line" somewhere - we wouldn't write `SkiaLayer` for each renderer, for example.

igordmn: > For me the live-resizing-in-each-redrawer was clear

It is better to think how clear it is for other developers. I understand that this is a hack, but we should still write a clear code for it. To know if it is clear we can find objective properties.

For me, in live-resizing feature, it is hard to understand and wrap my mind around:
- the data flow (It is a separate flow with reverse direction)
- the all connections with rendering and `FrameDispatcher` logic (they became connected with live-resizing too)
- concurrent logic

To connect all the dots, I have to read all the classes with their details. And I think it will be the same for others.

As a separate individual issue, it can be solved on individual Redrawer level - by separating each Direct3DRedrawer/MetalRedrawer into privates `Direct3DDriver`, `Direct3DRenderer`.

If you strongly disagree with this PR, we can go in another direction with smaller steps - extract driving from Redrawer's into separate classes where it is needed, to simplify redrawers.

@MatkovIvan, @m-sasha, does it work for you?

As I understand this PR paved the road to 2 features Ivan works on, if I am not mistaken:
- provides offscreen driving, but reusing the same rendering code
- provides public API for the rendering code/context for external applications, so they can add custom 3D rendering

m-sasha: As I mentioned, I see the merging of renderers and context handlers as a separate, non-controversial refactoring, and I would like to complete it first. Then it would be easier to understand additional changes.

igordmn: > As I mentioned, I see the merging of renderers and context handlers as a separate, non-controversial refactoring,

But this is either finished, or not finished in non-related parts like native.

If we talk about the name separately in the context of ContectHandler/Redrawer - I would keep just Redrawer, not Renderer (we did that too early)

igordmn: > But this is either finished, or not finished in non-related parts like native.

So we still need to discuss things we do in this PR - they are separate and we can do them in parallel

m-sasha: It would be easier to look at a smaller set of changes that have a specific purpose.

> But this is either finished, or not finished in non-related parts like native.

So not finished. The original PR did part of that refactoring plus unrelated changes. This PR goes another step forward but still doesn't finish it. But adds more unrelated changes.

> If we talk about the name separately in the context of ContectHandler/Redrawer - I would keep just Redrawer, not Renderer (we did that too early)

But it's not a *re*drawer. It's not responsible only for *re*drawing, right?

igordmn: > It would be easier to look at a smaller set of changes that have a specific purpose.

Agree

> So not finished. 

Agree, we still need to finish this part.

> But it's not a redrawer. It's not responsible only for redrawing, right?

If we talk how to name the current thing in `master`, then by default I would just keep what was there for the implementation and the base classes without thinking what is a better name.

If we want to think how to rename it, we can see what will be left after refactorings.

The initial name `Redrawer` was chosen as `Thing that should redraw SkiaLayer for external requests`:
- the `update` is just a thing needed for `draw`
- a different-paced update/draw calls in MetalRedrawer - is an optimization
- live resizing - added later and I argued that it doesn't belong there, but we kept it for simplicity
- extra Metal layer creation/resizing - is a helper to fake SkiaLayer redrawing

I am not against other names like `FrameDriver` or `Rerenderer` if we keep the same functionality.

I am for `Renderer` if we extract the frame driving functionality.

m-sasha: > The initial name Redrawer was chosen as Thing that should redraw SkiaLayer for external requests

Why does a "thing that **_re_**draws SkiaLayer for external requests" need a per-graphics-API (Metal, D3D etc.) implementation? My argument is that it's not a REdrawer. It's just a "drawer", "painter", "renderer". 

Or even better - explicitly tie the name to its purpose, which is to implement a common API for the functionality we need from the native graphics API. Something like `PlatformGraphics`, `GraphicsImpl`...

igordmn: You are right. `re` just tells more explicitly that we draw on something that was already drawn. But it is not much needed to tell that.

`PlatformGraphics` good to me too. But subjectively it sounds like a stateless thing - if we rename to it, it is better to move the state into suspend functions or generic objects.

skiko/src/awtMain/kotlin/org/jetbrains/skiko/renderer/MetalRenderer.kt
m-sasha: Let's pass the flag from native and if there's nothing else from Igor, merge it.

skiko/src/awtMain/kotlin/org/jetbrains/skiko/renderer/FrameDriver.kt
```diff
@@ -0,0 +1,121 @@
...
+ * The frame policy of one [SkiaLayer]: routes frame requests to the [scheduler], or to the
+ * platform's own frame loop during a live resize. It produces the immediate and before-shown
+ * frames itself; disposing it disposes the whole render stack.
+ */
+@OptIn(ExperimentalSkikoApi::class)
+internal class FrameDriver(
+    private val layer: SkiaLayer,
+    private val producer: FrameProducer,
+    private val scheduler: FrameScheduler,
+) : LiveResizeListener {
```
igordmn:
```suggestion
) {
```

It is better to create a field, to not expose unneeded API to `FrameDriver` callers

skiko/src/awtMain/kotlin/org/jetbrains/skiko/renderer/FrameProducer.kt
```diff
@@ -0,0 +1,70 @@
...
+import java.awt.Dimension
+import java.util.concurrent.atomic.AtomicBoolean
+
+/**
+ * Produces the frames of one [SkiaLayer]: records the layer's content, renders it through the
+ * [renderer], and carries the per-frame analytics. [FrameDriver] and the [FrameScheduler] decide
+ * when each step runs.
+ */
+@OptIn(ExperimentalSkikoApi::class)
+internal class FrameProducer(
```
igordmn, line 15: In the future we can move `draw`, `update`, from `SkiaLayer` here (with `renderDelegate`, `picture` and `fpsCounter`). This functionality looked out of place in `SkiaLayer` and redrawers had to reach out its own owner.

skiko/src/awtMain/kotlin/org/jetbrains/skiko/renderer/FrameDriver.kt
```diff
@@ -0,0 +1,121 @@
...
+
+    val presentsOnLayout: Boolean get() = renderer.presentsOnResize && !isPlatformDrivingFrames
+
+    fun needRender(throttledToVsync: Boolean) {
+        check(!isDisposed) { "FrameDriver is disposed" }
+
+        if (isPlatformDrivingFrames) {
+            renderer.requestPlatformDrivenFrame()
+        } else {
+            producer.requestUpdate()
```
igordmn: `requestUpdate`/`updateIfRequested` were added with `earlyRecordDispatcher`, and it is better to keep requesting logic there in one place:
- move `requestUpdate`/`updateIfRequested` as private functions of `SingleFrameScheduler`
- call `requestUpdate()` in the beginning of `SingleFrameScheduler.scheduleFrame`
- expose just `FrameProducer.updateIfRequested(time) = update(time)`

skiko/src/awtMain/kotlin/org/jetbrains/skiko/renderer/FrameDriver.kt
```diff
@@ -0,0 +1,121 @@
...
+            scheduler.scheduleFrame(throttledToVsync)
+        }
+    }
+
+    fun renderImmediately() {
+        check(!isDisposed) { "FrameDriver is disposed" }
+        layer.update(renderTime())
```
igordmn: We can replace it by `producer.update` after replacing `updatIfRequested`. It looks unnecessary, but it looks like a responsibility of the producer to `update`, and it makes clear that they are 2 steps of one functionality:
```
producer.update(renderTime())
something {
    producer.inFrame or drawFrame
}
```

igordmn (approved):

m-sasha (approved):

PR #1275, opened 2026-08-20
[benchmarks] Add Skiko benchmarks module

nopjar:
Introduced a new module for benchmarking Skiko across platforms. This includes configurable benchmark modes, example cases, result comparison scripts, and detailed documentation for usage.

Ticket can be found: [here](https://youtrack.jetbrains.com/issue/SKIKO-1175)

### DIFF
--- a/benchmarks/SkikoBenchmarks/README.md
+++ b/benchmarks/SkikoBenchmarks/README.md
@@ -40,13 +40,13 @@ You can also run the Gradle convenience task:
 ./gradlew -p benchmarks/SkikoBenchmarks runBrowserAndSaveStats -Pskiko.version=<skiko-version>
 ```
 
-JVM/AWT can be run with:
+JVM can be run with:
 
 ```bash
-./benchmarks/SkikoBenchmarks/run_benchmarks.main.kts awt version=<skiko-version>
+./benchmarks/SkikoBenchmarks/run_benchmarks.main.kts jvm version=<skiko-version>
 ```
 
-The JVM/AWT task prints the same JSON report directly to the terminal.
+The JVM task prints the same JSON report directly to the terminal.
 
 Additional arguments:
 
@@ -138,10 +138,10 @@ The included-build path follows the web samples: `org.jetbrains.skiko:skiko` is 
 with the local `skiko` project, and the wasm runtime is copied from the included build's
 `skikoWasmJar` output.
 
-AWT can also run against the current checkout:
+JVM can also run against the current checkout:
 
 ```bash
-./benchmarks/SkikoBenchmarks/run_benchmarks.main.kts awt \
+./benchmarks/SkikoBenchmarks/run_benchmarks.main.kts jvm \
     version=<version-label> \
     composite=true
 ```
--- a/benchmarks/SkikoBenchmarks/README.md
+++ b/benchmarks/SkikoBenchmarks/README.md
@@ -1,0 +1,147 @@
+# Skiko Benchmarks
+
+This project contains lightweight multiplatform benchmarks for comparing Skiko artifacts.
+The same benchmark definitions are used by JVM, JS, and wasmJs entry points; wasmJs is the
+primary target.
+
+Current benchmark coverage:
+
+```text
+surface_allocation      raster surface creation and teardown
+rect_grid_draw          many small filled rect draws
+path_parse_and_draw     SVG path parsing and stroked path rendering
+image_snapshot_encode   image snapshot and PNG encoding
+image_scale_draw        scaled image draws with sampling
+clip_transform_draw     save/restore, clip, transform, and rounded-rect drawing
+path_boolean_ops        path union/intersection operations
+surface_read_pixels     raster readback into Bitmap
+text_blob_draw          positioned glyph blob drawing
+```
+
+## Run Against a Maven Artifact
+
+```bash
+./benchmarks/SkikoBenchmarks/run_benchmarks.main.kts web version=<skiko-version>
+```
+
+This follows the Compose benchmark flow: Gradle starts a local result server, starts the wasm
+browser benchmark app, waits for the browser to POST the completed report, and archives the result to:
+
+```text
+benchmarks/SkikoBenchmarks/build/benchmarks/archive/web/<skiko-version>_run1/skiko-wasm.json
+```
+
+Use any resolvable Skiko Maven version. The version is required; the benchmark
+project intentionally has no default Skiko artifact version.
+
+You can also run the Gradle convenience task:
+
+```bash
+./gradlew -p benchmarks/SkikoBenchmarks runBrowserAndSaveStats -Pskiko.version=<skiko-version>
+```
+
+JVM/AWT can be run with:
+
+```bash
+./benchmarks/SkikoBenchmarks/run_benchmarks.main.kts awt version=<skiko-version>
+```
+
+The JVM/AWT task prints the same JSON report directly to the terminal.
+
+Additional arguments:
+
+```bash
+./benchmarks/SkikoBenchmarks/run_benchmarks.main.kts web version=<skiko-version> runs=3 benchmarks=rect_grid_draw,path_parse_and_draw
+```
+
+Each run prints a compact table to stdout and archives one JSON file per benchmark.
+
+## Benchmark Modes
+
+Skiko benchmarks support the same `modes=` argument shape as Compose, with modes
+that make sense for synchronous Skia operations:
+
+```bash
+./benchmarks/SkikoBenchmarks/run_benchmarks.main.kts web \
+    version=<skiko-version> \
+    modes=SIMPLE,VSYNC_EMULATION,STARTUP
+```
+
+Supported modes:
+
+```text
+SIMPLE            warmed repeated operation timing; default mode
+VSYNC_EMULATION   repeated timing plus 60 Hz budget-oriented percentile data
+STARTUP           first operation timing with no warmup
+```
+
+`REAL` mode is not supported. Compose uses it to measure real display/vsync
+behavior, but these Skiko benchmarks run offscreen synchronous Skia operations.
+
+## Compare Artifacts
+
+```bash
+./benchmarks/SkikoBenchmarks/compare_benchmarks.main.kts \
+    v1=<old-skiko-version> \
+    v2=<new-skiko-version> \
+    platform=web \
+    runs=3 \
+    modes=SIMPLE \
+    metric=average
+```
+
+If `v2` is omitted, the comparison target is the current checkout using the
+composite build path:
+
+```bash
+./benchmarks/SkikoBenchmarks/compare_benchmarks.main.kts \
+    v1=<published-skiko-version> \
+    platform=web \
+    runs=3 \
+    modes=SIMPLE,STARTUP
+```
+
+The comparison script runs missing benchmark data, reads archived JSON reports, and prints a
+stdout table with the selected timing metric, percentage difference, and status.
+`metric=average` is the default. Use `metric=median` to compare by `medianMillis`.
+
+To reuse existing archives:
+
+```bash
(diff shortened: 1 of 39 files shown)


benchmarks/SkikoBenchmarks/src/commonMain/kotlin/org/jetbrains/skiko/benchmarks/cases/image/ImageScaleDraw.kt
```diff
@@ -0,0 +1,55 @@
...
+import org.jetbrains.skia.Rect
+import org.jetbrains.skia.SamplingMode
+import org.jetbrains.skia.Surface
+import org.jetbrains.skia.impl.use
+import org.jetbrains.skiko.benchmarks.BenchmarkCase
+import kotlin.math.roundToLong
+
+val imageScaleDrawBenchmark = BenchmarkCase("image_scale_draw") {
+    makePatternImage(128, 128).use { image ->
+        Surface.makeRasterN32Premul(512, 512).use { surface ->
```
eymar: As far as I remember, `Surface.makeRasterN32Premul` creates a CPU-backed surface. Such benchmarks won't involve GPU.

 Ideally, the benchmarks would be as close to real apps as possible. While CPU-only rendering is useful too, benchmarking Skiko operations on GPU-backed Surface would help us observe more representative numbers and catch real regressions.

To involve GPU, I believe this solution requires having additional expect/actual so platforms can provide proper a GPU-backed Surface.

WDYT?

nopjar: Good point, thank you!
I'd still keep the CPU backed benchmark just in case, but having a GPU based benchmark is definitely something needed.

I'll add them asap :)

benchmarks/SkikoBenchmarks/src/commonMain/kotlin/org/jetbrains/skiko/benchmarks/BenchmarkConfig.kt
```diff
@@ -0,0 +1,53 @@
+package org.jetbrains.skiko.benchmarks
+
+enum class BenchmarkMode(val argumentName: String, val resultSuffix: String?) {
+    SIMPLE("SIMPLE", null),
+    VSYNC_EMULATION("VSYNC_EMULATION", "vsync_emulation"),
```
pjBooms: In CMP, vsync_emulation mode really emulates vsync doing nothing between frames to let gc do its job between. It was important for K/N team. Here, I see we just gather additional stat for the mode.

benchmarks/SkikoBenchmarks/README.md
```diff
@@ -0,0 +1,147 @@
...
+
+Supported modes:
+
+```text
+SIMPLE            warmed repeated operation timing; default mode
+VSYNC_EMULATION   repeated timing plus 60 Hz budget-oriented percentile data
+STARTUP           first operation timing with no warmup
+```
+
+`REAL` mode is not supported. Compose uses it to measure real display/vsync
```
pjBooms: I don't think it is worth to mention about 'real' mode here.

benchmarks/SkikoBenchmarks/build.gradle.kts
```diff
@@ -0,0 +1,331 @@
...
+        }
+        binaries.executable()
+    }
+
+    sourceSets {
+        commonMain.dependencies {
+            implementation(libs.skiko)
+        }
+
+        val awtMain by getting {
```
pjBooms: "awt" is a not standard name for jvm target. If it  is made for purpose, I don't mind

benchmarks/SkikoBenchmarks/run_benchmarks.main.kts
```diff
@@ -0,0 +1,413 @@
...
+        } else {
+            runArgs.add(key)
+        }
+    }
+
+    when (platform) {
+        Platform.WEB -> {
+            println("Starting benchmark server in background...")
+            val benchmarkServer = LocalBenchmarkServer(
+                saveStatsToJSON = true,
```
pjBooms: It seems that user arg saveStatsToJSON  is overriden here

benchmarks/SkikoBenchmarks/run_benchmarks.main.kts
```diff
@@ -0,0 +1,413 @@
...
+    val resultsDir = projectDir.resolve("build/benchmarks/json-reports")
+    if (resultsDir.exists()) {
+        resultsDir.listFiles { file -> file.extension == "json" }?.forEach { it.delete() }
+    }
+}
+
+fun List<String>.valueFor(key: String): String? =
+    firstOrNull { it.startsWith("$key=") }?.substringAfter("=")
+
+fun findBenchmarkProjectDir(): File {
```
pjBooms: There are three such functions declared in 3 scripts. Can they be deduplicated?

benchmarks/SkikoBenchmarks/run_benchmarks.main.kts
```diff
@@ -0,0 +1,413 @@
...
+    return false
+}
+
+class LocalBenchmarkServer(
+    private val saveStatsToJSON: Boolean,
+    private val serverToken: String,
+    private val extraGradleArgs: List<String>,
+    private val port: Int = BENCHMARK_SERVER_PORT
+) {
+    val stopped = AtomicBoolean(false)
```
pjBooms: I am not sure about two independent AtomicBooleans here. Maybe introduce state here?

pjBooms (commented):
Looks good in general. I left a few comments, all are not critical.

benchmarks/SkikoBenchmarks/src/commonMain/kotlin/org/jetbrains/skiko/benchmarks/BenchmarkConfig.kt
nopjar: I'll just remove it then

benchmarks/SkikoBenchmarks/build.gradle.kts
nopjar: It is not done by purpose perse. Is "jvm" or "desktop" a better name?

pjBooms: They are more "standard"

benchmarks/SkikoBenchmarks/run_benchmarks.main.kts
nopjar: Yes, good point. I'll add a common script which shares the logic.

benchmarks/SkikoBenchmarks/run_benchmarks.main.kts
nopjar: It was actually on purpose, but I guess nothing speaks against it to also allow to disable it on the web.

benchmarks/SkikoBenchmarks/run_benchmarks.main.kts
nopjar: Good point, yes.

benchmarks/SkikoBenchmarks/build.gradle.kts
nopjar: Alright, I'll rename it to jvm then (considering that we may want to have different desktop targets besides the jvm)

benchmarks/SkikoBenchmarks/src/commonMain/kotlin/org/jetbrains/skiko/benchmarks/cases/image/ImageScaleDraw.kt
```diff
@@ -0,0 +1,87 @@
...
+
+private fun closeGpuSurfaceProvider() {
+    gpuSurfaceProvider?.close()
+    gpuSurfaceProvider = null
+    gpuSurfaceProviderInitialized = false
+}
+
+private fun runImageScaleDraw(surfaceProvider: BenchmarkSurfaceProvider): Long {
+    makePatternImage(128, 128).use { image ->
+        return surfaceProvider.withSurface(512, 512) { surface ->
```
pjBooms, line 29: I guess we can use now gpuSurface for other benchmarks as well, cannot we?

pjBooms (approved):

benchmarks/SkikoBenchmarks/src/commonMain/kotlin/org/jetbrains/skiko/benchmarks/cases/image/ImageScaleDraw.kt
nopjar: Ah, yes the current implementation would introduce a lot of duplicated code. I've refactored it so other benchmarks can also use it.

benchmarks/SkikoBenchmarks/src/commonMain/kotlin/org/jetbrains/skiko/benchmarks/BenchmarkConfig.kt
```diff
@@ -0,0 +1,52 @@
+package org.jetbrains.skiko.benchmarks
+
+enum class BenchmarkMode(val argumentName: String, val resultSuffix: String?) {
+    SIMPLE("SIMPLE", null),
```
eymar, line 13: Could you please add a KDOc here describing the modes?

benchmarks/SkikoBenchmarks/src/commonMain/kotlin/org/jetbrains/skiko/benchmarks/cases/drawing/ClipTransformDraw.kt
```diff
@@ -0,0 +1,47 @@
...
+
+val clipTransformDrawBenchmark = BenchmarkCase("clip_transform_draw") {
+    runClipTransformDraw(RasterBenchmarkSurfaceProvider)
+}
+
+val clipTransformDrawGpuBenchmark = BenchmarkCase("clip_transform_draw_gpu",
+    isSupported = { GpuBenchmarkSurfaceProvider.isSupported() },
+    tearDown = { GpuBenchmarkSurfaceProvider.close() },
+) {
+    runClipTransformDraw(GpuBenchmarkSurfaceProvider.get()!!)
```
eymar, line 20: as i understand the benchmark runner invokes this lambda repeatedly?

Is my understanding corret:
- on every iteration, this creates a new surface? (heavy operation)

So it would measure not just drawing operations, but a mix of init+draw. 

Compose benchmars initialize the scene only once - https://github.com/JetBrains/compose-multiplatform/blob/master/benchmarks/multiplatform/benchmarks/src/webMain/kotlin/Main.web.kt#L30 and do not include the init in the measurements. (cc: @pjBooms could you please confirm or correct me?)

Measuring the init is valuable too, but separately from drawing. 

I'd suggest to refactor the implementation so that the repetetive benchmarks runs re-use the surface.

WDYT?

nopjar: Your concern is absolutely valid, but `#get()` will cache the surface until `#close()` is called. So only the first time a new surface will be created (heavy operation) while with the other runs, it will simply reuse it.

The creation itself will in SIMPLE mode not be part of the timing as all of the benchmarks have a certain amount of warmups, which will create the surface.

In STARTUP there are no warmups so we will also see the impact of the surface creation.

In my personal opinion, this should be enough of a separation, but let me know what you think! :)

eymar: I see. Thanks :) 

I didn't notice that.

eymar (approved):

PR #1276, opened 2026-08-21
Fix Linux system theme detection returning unknown

hub-bla:
`libdbus-1.so` is a symlink to `libdbus-1.so.3` and is only available when the `libdbus-1-dev` package is installed. This PR updates the load logic to search for `libdbus-1.so.3` first with `libdbus-1.so` as a fallback.

Fixes - [SKIKO-1177](https://youtrack.jetbrains.com/issue/SKIKO-1177)

### DIFF
--- a/skiko/src/awtMain/cpp/linux/theme.cc
+++ b/skiko/src/awtMain/cpp/linux/theme.cc
@@ -12,7 +12,10 @@ static dbus_bool_t initDbusThreads(void* lib) {
 }
 
 static void* loadLibDbus() {
-    static void* dlOpenResult = dlopen("libdbus-1.so", RTLD_LAZY | RTLD_LOCAL);
+    static void* dlOpenResult = dlopen("libdbus-1.so.3", RTLD_LAZY | RTLD_LOCAL);
+    if (!dlOpenResult) {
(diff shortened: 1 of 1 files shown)


MatkovIvan (approved):

skiko/src/awtMain/cpp/linux/theme.cc
```diff
@@ -12,7 +12,10 @@ static dbus_bool_t initDbusThreads(void* lib) {
 }
 
 static void* loadLibDbus() {
-    static void* dlOpenResult = dlopen("libdbus-1.so", RTLD_LAZY | RTLD_LOCAL);
+    static void* dlOpenResult = dlopen("libdbus-1.so.3", RTLD_LAZY | RTLD_LOCAL);
+    if (!dlOpenResult) {
+        dlOpenResult = dlopen("libdbus-1.so", RTLD_LAZY | RTLD_LOCAL);
```
MatkovIvan: Wait. It will be executed on each call if not successful, it was not like this before

hub-bla: Right, fixed

PR #1281, opened 2026-08-27
Make `Version` resolve in IDEA

m-sasha:


hub-bla (approved):

m-sasha:
@hub-bla I have only a superficial understanding of Gradle phases and built-in/common tasks, so the changes here need a good look to see if I didn't break anything.

hub-bla:
@m-sasha I did a second look at it. I've tested it manually, with llm and checked if it has an effect on ci. From my perspective, everything looks fine ;)

m-sasha:
Awesome, thanks!

PR #1282, opened 2026-08-28
Pace SkiaSwingLayer repaints to the display refresh

rock3r:
A scene that invalidates continuously through `SkiaSwingLayer` calls `repaint()` as fast as the EDT can finish paint cycles. A Compose draw-phase animation in SwingGraphics mode is one example. It renders far above the display refresh rate, and the GPU draws frames that are never shown.

This PR adds opt-in frame pacing that holds those repaints to the display's refresh rate. All of it is inside Skiko, so it works on any JVM.

Fixes CMP-8124

> [!NOTE]
> Measurements and reasoning: https://specs.sebastiano.dev/frame-pacing/
> Review guide for this code: https://specs.sebastiano.dev/frame-pacing/skiko-pacer-review

## Effect

Measured on Linux with a Radeon 780M, a 180 Hz display and a 500×320 window:

| | unpaced | paced |
|---|---|---|
| renders/sec | 2422.6 | **180.0** |
| GPU power | 26.17 W | **8.78 W** (−66%) |

This is close to the best case. The saving follows the ratio of unpaced render rate to refresh rate. A renderer that cannot reach the refresh rate is not slowed further.

Pacing removes frames that were never shown. It does not make a shown frame cheaper: `SkiaSwingLayer` still reads each frame back from the GPU and blits it through Swing. That is separate work.

## Changes

- **`SkiaSwingLayer.needRender()`**, new public API. It requests a repaint like `repaint()`, but paced when pacing is available. `repaint()` is unchanged: Swing calls it internally, and the pacer uses it as the unpaced primitive.
- **`SwingRepaintPacer`**, internal. A `FrameDispatcher` loop on the EDT. It repaints a requested frame at once, then suspends until the next display tick. A continuous stream of requests becomes one repaint per tick. A request that arrives while the loop is idle paints with no added latency.
- **`SkikoFramePacing`**, internal. One clock per display. A Kotlin daemon thread waits on the display's vblank. The native code only opens, waits on and closes the source, and has no threads of its own.
- Pacing is off by default. The system property `skiko.swing.frame.pacing` turns it on.

| Platform | Clock |
|---|---|
| macOS 14+ | `CADisplayLink` of the display's `NSScreen` |
| Windows | `IDXGIOutput::WaitForVBlank` on the display's output, with no fallback |
| Linux | Kernel DRM vblank of the CRTC whose mode matches the display's refresh rate |
| macOS 13 and older, Linux without DRM access | Timer at the nominal refresh rate |

Display ids come from public AWT API only (`GraphicsDevice.getIDstring()`). There is no `sun.awt` access.

Windows needs no fallback in any session we could produce. A Remote Desktop session and a Hyper-V VM without a GPU both still expose a DXGI output with a blocking vblank (measured: 32 Hz and 64 Hz).

## When a display is not paced

Requests pass through as plain `repaint()` calls when:

- the property is off, or no clock is available;
- the toolkit reports no refresh rate, on Linux or with the timer. Both need the rate, and neither guesses one;
- a clock accepts a subscription and then never ticks. The pacer does not retry that display until the component moves.

Every wait for a tick has a timeout, so a clock that stops cannot stall the frame loop. The pacer resolves the display again on every frame, so a window dragged to a display with another rate follows it. It also checks the display's refresh rate about once a second and subscribes again when the mode changes. A clock whose thread has ended, or that was made for another rate, is never shared with a new subscriber. It releases the subscription after 1 s without requests, so an idle component keeps no clock thread alive.

## Known limit on Linux

The toolkit gives Skiko no connector name, so the CRTC is matched by refresh rate. Two displays at the same rate share the first CRTC that matches. The rate is correct for both, but the phase is wrong for one of them and drifts: we measured one full cycle in about 8 minutes. During part of each cycle that window drops and doubles frames. Binding by connector name would fix this, but it needs toolkit internals.

## Tests

40 tests in `awtTest`. `SwingRepaintPacerTest` (15) drives the frame loop with a fake clock. `SkikoFramePacingServiceTest` (16) covers clock selection, clock sharing and subscription lifecycle. `NativeVBlankClockTest` (7) runs the clock loop against fake sources that spin, fail or cannot be opened. `SkiaSwingLayerNeedRenderTest` (2) covers `needRender()`.


igordmn:
@m-sasha, could you also look at it?

igordmn:
@MatkovIvan, this might interfere with your refactoring

skiko/src/awtMain/kotlin/org/jetbrains/skiko/swing/SkiaSwingLayer.kt
```diff
@@ -97,19 +99,46 @@ open class SkiaSwingLayer(
...
+     * Must be called on the AWT event dispatch thread.
+     */
+    fun needRender() {
+        check(isEventDispatchThread()) { "Method should be called from AWT event dispatch thread" }
+        val repaintPacer = repaintPacer
+        if (repaintPacer != null && !isDisposed) {
+            repaintPacer.requestRepaint()
+        } else {
+            repaint()
+        }
```
m-sasha, line 140: Is there a reason we only check `isDisposed` in the `repacer != null` branch?

skiko/src/awtMain/kotlin/org/jetbrains/skiko/swing/SwingRepaintPacer.kt
```diff
@@ -0,0 +1,298 @@
...
+        ensureSubscription(service)
+        if (subscription == null) return
+
+        try {
+            withTimeout(tickTimeoutMillis) {
+                suspendCancellableCoroutine { continuation ->
+                    tickContinuation.set(continuation)
+                    continuation.invokeOnCancellation {
+                        tickContinuation.compareAndSet(continuation, null)
+                    }
```
m-sasha: Can we use `Channel<Unit>(CONFLATED)` here instead of manually managing the continuation?

skiko/src/awtMain/kotlin/org/jetbrains/skiko/swing/SwingRepaintPacer.kt
```diff
@@ -0,0 +1,298 @@
...
+     */
+    private suspend fun awaitTickIfPaced() {
+        val service = service ?: return
+
+        // The subscription must stay alive for as long as this wait: the idle timer would otherwise close it and
+        // leave the loop suspended until its own timeout.
+        idleTimer.stop()
+
+        ensureSubscription(service)
+        if (subscription == null) return
```
m-sasha: Maybe have `ensureSubscription` return a success/failure flag.

m-sasha (approved):
The Kotlin side looks ok, left a few comments. Can't comment on the native side.

Can this cause a desync between a regular Swing component and Compose? For example, have Swing and Compose draw a colored square. Then do
```
swingBox.setColor(BLUE)  // Calls `repaint`
composeBox.setColor(BLUE)  // Calls `needRender`
delay(8.milliseconds)
swingBox.setColor(RED)
composeBox.setColor(RED)
```
Would we end up with red in Swing and blue in Compose for a frame?

rock3r:
> Can this cause a desync between a regular Swing component and Compose? [...] Would we end up with red in Swing and blue in Compose for a frame?

Yes, it's possible, but not really something we can do anything about unless pacing is done in JBR and shared between Swing and Skiko. It's at most a one-frame disagreement, and that is exactly the same thing that happens when using `SkiaLayer`, so I assumed it was fine here too. This is merely aligning the Swing compositing path in line with the AWT path.

skiko/src/awtMain/kotlin/org/jetbrains/skiko/swing/SkiaSwingLayer.kt
rock3r: We check `isDisposed` out of an abundance of caution — we could depend on just the null check: the pacer is nulled before the flag is set to false, and it is a synchronous code block that does both, so reeeeally defensive coding more than anything else. It's here because if for whatever reason the behaviour changes and the lockstep execution of the nulling out and flag setting get separated, this checks both. As of today: can be removed, it's future robustness against an impl detail (field setting call ordering).

Assuming this is what you were asking, since checking `isDisposed` explicitly in the else branch seems redundant to me so I am guessing that isn't what you meant

skiko/src/awtMain/kotlin/org/jetbrains/skiko/swing/SwingRepaintPacer.kt
rock3r: We can switch if you prefer, but there is a slight difference in semantics: a conflated channel buffers a tick that arrives when there is nobody listening, whereas the current behaviour drops it. There is a tiny chance that a tick arriving while the loop is idle but the sub is still active may cause one extra unneeded redraw if we used conflated; rendezvous sounds closer to the current behaviour. I asked GLM-5.3 to fact check me on this since I was having existential doubts, and it confirms my choice:

> One real semantic difference: the manual version drops a tick that arrives with nobody waiting (`getAndSet` returns `null`). `CONFLATED` buffers one. Consequences:
> * Tick during the tiny repaint window of the loop → next `receive()` completes instantly → skips one pacing period (rare; window is sub-ms vs a ~16.7 ms period).
> * Tick while the loop is idle but the subscription is still alive (up to the 1 s `IDLE_RELEASE_DELAY_MILLIS`) → the first wait after idle consumes the stale tick immediately → one extra unpaced repaint burst before re-locking to the cadence. Bounded, minor, but it does weaken the "at most one repaint per tick" contract.
> * `Channel<Unit>()` with `RENDEZVOUS` (the default capacity) is the exact semantic match — `trySend` only succeeds when a receiver is suspended, i.e. drop-if-no-waiter, same as today.

skiko/src/awtMain/kotlin/org/jetbrains/skiko/swing/SwingRepaintPacer.kt
rock3r: Makes sense, coming soon

skiko/src/awtMain/kotlin/org/jetbrains/skiko/swing/SkiaSwingLayer.kt
m-sasha: I meant why not just check it at the top?
```
if (isDisposed) return
```

skiko/src/awtMain/kotlin/org/jetbrains/skiko/swing/SwingRepaintPacer.kt
m-sasha: `RENDEZVOUS` is fine with me too; I just don't like the low-level continuation juggling.

You could also do CONFLATED and move the `component.repaint()` call to after `awaitTickIfPaced()` for the same effect. But then it would probably need a different name.

m-sasha:
> Yes, it's possible, but not really something we can do anything about unless pacing is done in JBR and shared between Swing and Skiko. It's at most a one-frame disagreement, and that is exactly the same thing that happens when using `SkiaLayer`, so I assumed it was fine here too. This is merely aligning the Swing compositing path in line with the AWT path.

SkiaLayer doesn't guarantee sync with Swing; that's what `SkiaSwingLayer` tries to do.

But anyway, just wanted to highlight this; not a blocker.

rock3r:
Yep I understand that, but today it is "in sync" only if it's in the hot path spewing frames as fast as possible. Wouldn't call it a fully mechanical syncing, it's more a spray-and-pray sync :D

skiko/src/awtMain/kotlin/org/jetbrains/skiko/swing/SkiaSwingLayer.kt
rock3r: Ah, yeah sure. Same behaviour as today, might as well.

m-sasha:
> Yep I understand that, but today it is "in sync" only if it's in the hot path spewing frames as fast as possible. Wouldn't call it a fully mechanical syncing, it's more a spray-and-pray sync :D

But that's how Swing works.

In `SkiaLayer` I added a flag to `needRender`: `throttledToVsync` with the idea to set it to true only for animations. I forget why we didn't end up using it from Compose.

skiko/src/awtMain/kotlin/org/jetbrains/skiko/swing/SwingRepaintPacer.kt
rock3r: I think rendezvous here is the right call — I'll switch it over. I got into the rabbit hole with the manual continuations when rendezvous was there all along :D I prefer the easier solution.

PS: `CONFLATED` and move `component.repaint()` after `awaitTickIfPaced()` is **not** the same thing as the current behaviour, or `RENDEZVOUS`. That becomes wait-then-paint, and that breaks the paint-first contract ("a frame that is already requested should never wait for vsync"). We'd add up-to-one-refresh-period latency to the first frame after idle. The pacer's whole idea is repaint-immediately-then-suspend, so I'll do rendezvous as that keeps the intended behaviour.

rock3r:
I do not really know, I can only make a guess. I found the unmerged branch `m-sasha/skiko-needRedraw` (last change in Sep '25, on skiko 0.9.25), it plumbed `onNextFrame` through `invalidate`, and the on-screen path was wired, but the Swing path was left with a literal `// TODO: if onNextFrame is true, delay until next frame` — which couldn't be done back then; there was nothing for `SkiaSwingLayer` to delay on. If I understand it correctly, that TODO is basically what this PR implements :)

It also changed the `invalidate` signature across the scene factories and then predates the `needRedraw` → `needRender` rename, so when Compose bumped skiko the no-arg call just won by default. I reckon that to merge that flag you'd need to have this change in anyway.

I'd say this supersedes the Swing-path half of that branch — the delay until next frame TODO is exactly what `needRender()` here does, and once it lands the Compose side for Swing interop is a one-liner, no `onNextFrame` plumbing needed. The on-screen `throttledToVsync` stuff in the same branch is still a separate thing that may be worth picking back up but is outside the scope of this change.

rock3r:
PS: could you re-trigger the CI now that I addressed all comments and fixed the tests?

igordmn:
Still reviewing as well...

skiko/src/awtMain/cpp/linux/FramePacing.cc
```diff
@@ -0,0 +1,405 @@
...
+// For pthread_timedjoin_np; must precede every glibc header.
+#ifndef _GNU_SOURCE
+#define _GNU_SOURCE
+#endif
+#endif
+
+#include <jni.h>
+
+/*
+ * Skiko-owned per-CRTC vblank pacing clock for
```
igordmn: Can you reduce all the comments in the whole PR and check their validity? These are typical AI comments I see in a lot of PRs, and they need some work before they become helpful.

Some of the code in the PR don't need explanation, or it is difficult to read, or it is misleading, or it describes "what you didn't do", or it describes unrelated implementation details, or describes another unrelated part of code.

I would ask AI to do reduce it 😅. It writes okay comments if you point out the issues.

Examples from this comment:

### Self-explanatory
- "Skiko-owned" - we have only Skiko
- "org.jetbrains.skiko.swing.LinuxDrmVBlankClock"  - this is visible from the method names
- "no libdrm dependency" - "with libdrm dependency" sounds no different.

### Describing implementation details of the code in the middle
When it is written here, there is no context and it is not understandable, when you move it to the place where it is used, it might become self-explanatory.
- "the clock thread blocks in DRM_IOCTL_WAIT_VBLANK"
- "The CRTC is bound by matching each active CRTC's mode period against"

### Useful comments
- "Two displays running at the same refresh rate share the first CRTC that" and its experiment results - this is a major flaw that needed to be known.

skiko/src/awtMain/cpp/linux/FramePacing.cc
```diff
@@ -0,0 +1,405 @@
...
+ * display's advertised refresh period, across all DRM card nodes. This is
+ * approximate in two known ways, both of which only cost accuracy, never
+ * correctness.
+ *
+ * Two displays running at the same refresh rate share the first CRTC that
+ * matches, so one of them is paced by the other's scanout. Measured on a
+ * 2560x1440 and a 1920x1080 panel both set to 120 Hz: the rate stays right to
+ * 0.00205 Hz (0.0017%) and only the phase is wrong, sliding through a full
+ * frame every 58529 frames, about 8 minutes. Frames are neither dropped nor
+ * doubled; what is lost is phase alignment to the intended display.
```
igordmn: With a wrong phase, the second display experiences frequent frame drops (we start rendering too late), more frequent when we are closer to the end of the current vsync period.

skiko/src/awtMain/cpp/linux/FramePacing.cc
```diff
@@ -0,0 +1,405 @@
...
+ * anything on screen is changing, which is what a subscription clock needs.
+ *
+ * The CRTC is bound by matching each active CRTC's mode period against the
+ * display's advertised refresh period, across all DRM card nodes. This is
+ * approximate in two known ways, both of which only cost accuracy, never
+ * correctness.
+ *
+ * Two displays running at the same refresh rate share the first CRTC that
+ * matches, so one of them is paced by the other's scanout. Measured on a
+ * 2560x1440 and a 1920x1080 panel both set to 120 Hz: the rate stays right to
```
igordmn: If the second display has 165Hz, does it use 120Hz rate of the first display?

Is it difficult to be bound to the exact vsync of the current display?

skiko/src/jvmMain/kotlin/org/jetbrains/skiko/SkikoProperties.kt
```diff
@@ -97,6 +97,24 @@ object SkikoProperties {
...
+     */
+    val swingFramePacingEnabled: Boolean
+        by lazyBooleanProperty("skiko.swing.frame.pacing", defaultValue = false)
+
+    /**
+     * Forces Skiko's own pacing clocks to the phase-aligned timer even where a native display
+     * clock (CADisplayLink, DXGI vblank, DRM vblank) is available. Debug and measurement escape
+     * hatch.
+     */
+    val swingFramePacingForceTimer: Boolean
```
igordmn:
```suggestion
    val swingFramePacingFrameLimiting: Boolean
```

This is called frame limiting, TODO

skiko/src/awtMain/kotlin/org/jetbrains/skiko/swing/SwingRepaintPacer.kt
```diff
@@ -0,0 +1,296 @@
...
+
+/**
+ * Paces invalidation-driven repaints of a Swing component to the display refresh, using Skiko's own display clocks
+ * ([SkikoFramePacingService]), so it works on any JVM.
+ *
+ * Without pacing, a continuously invalidating scene (e.g., a draw-phase animation) calls [Component.repaint] as fast as
+ * the EDT can complete paint cycles, rendering far above the display refresh rate and burning GPU on frames that are
+ * never shown.
+ *
+ * The pacer runs a [FrameDispatcher] frame loop on the EDT, in the same shape as the redrawers' frame schedulers: a
```
igordmn: (related to the comment's issue)

`FrameDispatcher` - is implementation details of this class

`the same shape as the redrawers' frame schedulers` - the implementation details of redrawers

(The rest of the paragraph describes a public behavior, and is very useful)

skiko/src/awtTest/kotlin/org/jetbrains/skiko/swing/SwingRepaintPacerTest.kt
```diff
@@ -0,0 +1,472 @@
...
+        waitUntil { onEdtGet { panel.repaintCount.get() } == baseline + 2 }
+        onEdt { pacer?.requestRepaint() }
+        waitUntil { onEdtGet { panel.repaintCount.get() } == baseline + 3 }
+        settle()
+        assertEquals(1, service.subscribeAttempts)
+    }
+
+    @Test
+    fun aThrowingBackendDegradesToUnpacedInsteadOfKillingTheFrameLoop() {
+        // A missing native entry point raises UnsatisfiedLinkError from subscribe. If that escaped
```
igordmn: If `UnsatisfiedLinkError` is thrown it is a critical bug needed to be fixed. I would not `catch` it or test it.

Are there another types of exceptions?

skiko/src/awtTest/kotlin/org/jetbrains/skiko/swing/SkikoProbeMain.kt
```diff
@@ -0,0 +1,12 @@
+package org.jetbrains.skiko.swing
+
+/**
+ * Standalone entry point for [SkikoFramePacingJitterProbe], to run the probe outside the
+ * Gradle test worker (whose environment measurably degrades wait precision on Windows).
+ */
+object SkikoProbeMain {
```
igordmn: How do you run it?

In IDEA/Windows it is:

Image

And:
```
No matching tests found in any candidate test task.
    Requested tests:
        Test pattern org.jetbrains.skiko.swing.SkikoProbeMain in task :awtTest
```
If I run via context menu - Run

skiko/src/awtTest/kotlin/org/jetbrains/skiko/swing/SkikoFramePacingJitterProbe.kt
```diff
@@ -0,0 +1,91 @@
...
+ * platform-native clock via [SkikoFramePacingService] and, for comparison, a raw [TimerClock] at
+ * the same period — and writes them to `build/reports/pacing-probe.txt`.
+ *
+ * Opt-in via the `SKIKO_PACING_PROBE=true` environment variable so normal test runs skip it.
+ */
+class SkikoFramePacingJitterProbe {
+
+    @Test
+    fun `measure tick interval statistics`() {
+        assumeTrue(System.getenv("SKIKO_PACING_PROBE") == "true")
```
igordmn: It is better to use a Gradle property

skiko/src/awtMain/kotlin/org/jetbrains/skiko/swing/SwingRepaintPacer.kt
```diff
@@ -0,0 +1,296 @@
...
+        tickChannel.trySend(Unit)
+    }
+
+    private fun resolveDisplayId(service: FramePacingService): Long {
+        val graphicsConfiguration = component.graphicsConfiguration ?: return UNKNOWN_DISPLAY_ID
+        return service.displayId(graphicsConfiguration)
+    }
+
+    private fun closeSubscription() {
+        subscription?.close()
```
igordmn: Does it block until the current `onTick` is called? If not, we can have races

skiko/src/awtMain/kotlin/org/jetbrains/skiko/swing/SwingRepaintPacer.kt
```diff
@@ -0,0 +1,296 @@
...
+    fun subscribe(
+        displayId: Long,
+        onTick: (displayId: Long, timeNanos: Long) -> Unit
+    ): AutoCloseable?
+}
+
+/**
+ * Resolves the [FramePacingService] the pacer uses: Skiko's own display clocks, so pacing works on any runtime.
+ */
+internal object FramePacingServices {
```
igordmn: Looks unnecessary, we can use `SkikoFramePacingService.instance` directly

skiko/src/awtMain/kotlin/org/jetbrains/skiko/swing/SkikoFramePacing.kt
```diff
@@ -0,0 +1,512 @@
...
+        return GraphicsEnvironment.getLocalGraphicsEnvironment().screenDevices
+            .firstOrNull { it.type == GraphicsDevice.TYPE_RASTER_SCREEN && deviceDisplayId(it) == displayId }
+    }
+
+    companion object {
+        private const val FALLBACK_PERIOD_NANOS = 1_000_000_000L / 60
+
+        /** The band of refresh rates [refreshPeriodNanos] is willing to believe. */
+        private const val MIN_PLAUSIBLE_REFRESH_HZ = 20
+        private const val MAX_PLAUSIBLE_REFRESH_HZ = 1_000
```
igordmn:
```suggestion
        private const val MAX_PLAUSIBLE_REFRESH_HZ = 2_000
```
https://www.lg.com/global/newsroom/news/media-entertainment-solution/lg-electronics-introduces-worlds-first-native-1000hz-full-hd-gaming-monitor/

2000 sounds enough 😂

skiko/src/awtMain/kotlin/org/jetbrains/skiko/swing/SkikoFramePacing.kt
```diff
@@ -0,0 +1,512 @@
...
+        ): DisplayClock = when {
+            forceTimer -> TimerClock(displayId, periodNanos)
+
+            hostOs == OS.MacOS && MacDisplayLinkClock.available(displayId) ->
+                MacDisplayLinkClock(displayId, periodNanos)
+
+            hostOs == OS.Windows && WinNativeClock.vblankAvailable() ->
+                WinNativeClock(displayId, periodNanos, useVBlank = true)
+
+            // Even without DWM the native clock paces on Windows' high-resolution waitable timer,
```
igordmn: In which case there can be no DWM?

If this case is unrealistic, I would prefer removing this mode, as it requires a lot of code, and using high-resolution timer harms power consumption:
https://learn.microsoft.com/en-us/windows/win32/sync/waitable-timer-objects

skiko/src/awtMain/cpp/windows/FramePacing.cc
```diff
@@ -0,0 +1,505 @@
...
+ * indices follow the same enumeration order, so a screen index is enough to
+ * find the matching output.
+ */
+
+typedef HRESULT (WINAPI *DwmGetCompositionTimingInfoType)(HWND, DWM_TIMING_INFO *);
+typedef HRESULT (WINAPI *CreateDXGIFactory1Type)(REFIID, void **);
+
+static DwmGetCompositionTimingInfoType pDwmGetCompositionTimingInfo = NULL;
+static CreateDXGIFactory1Type pCreateDXGIFactory1 = NULL;
+static JavaVM *framePacingJvm = NULL;
```
igordmn: You can use `extern "C" JavaVM *jvm;` instead of loading it again.

skiko/src/awtMain/cpp/windows/FramePacing.cc
```diff
@@ -0,0 +1,505 @@
...
+
+static DWORD WINAPI vblankThreadProc(LPVOID param)
+{
+    FramePacingClock *clock = (FramePacingClock *)param;
+    JNIEnv *env;
+    if (framePacingJvm->AttachCurrentThreadAsDaemon((void **)&env, NULL) != JNI_OK) {
+        return 0;
+    }
+
+    HANDLE timer = createHighResolutionTimer();
```
igordmn: Can we avoid using a timer for vsync?

If this is because of:
> A display in power save stays attached to the desktop but stops scanning
>     * out, and WaitForVBlank then returns immediately and successfully

it looks suspicious, is it really "successfully" or does it fail here?
```
FAILED(output->WaitForVBlank())
```

I see a lot of code related to non-vsync case and it will be difficult to support it. On the other hand it is fine for Skiko to be stopped if there is no display attached, or tick with less frequency (see, for example [occluded window 300ms-interval](https://github.com/JetBrains/skiko/blob/47dd4713a62713b50ba7d92ce4ab70a434777e92/skiko/src/awtMain/kotlin/org/jetbrains/skiko/renderer/MetalRenderer.kt#L136) update on macOs)

Can we remove all the timer-related code in all Os'es, and if there is a real need, reuse the Kotlin `TimerClock`?

skiko/src/awtMain/cpp/windows/FramePacing.cc
```diff
@@ -0,0 +1,505 @@
...
+    }
+
+    return (jlong)(intptr_t)clock;
+}
+
+JNIEXPORT void JNICALL
+Java_org_jetbrains_skiko_swing_WinNativeClock_nativeStart(JNIEnv *env, jclass cls, jlong ptr)
+{
+    FramePacingClock *clock = (FramePacingClock *)(intptr_t)ptr;
+    clock->thread = CreateThread(NULL, 0,
```
igordmn: Is it possible to rewrite the code to avoid threading on the native side? I.e. move it to Kotlin (including the code for macOs/Linux)

More native code - more difficult it to maintain. And if it also involves threading, it makes it even more difficult.

skiko/src/awtMain/cpp/windows/FramePacing.cc
```diff
@@ -0,0 +1,505 @@
...
+
+    HANDLE handles[2] = { clock->stopEvent, timer };
+    return WaitForMultipleObjects(2, handles, FALSE, INFINITE) == WAIT_OBJECT_0 + 1;
+}
+
+static DWORD WINAPI vblankThreadProc(LPVOID param)
+{
+    FramePacingClock *clock = (FramePacingClock *)param;
+    JNIEnv *env;
+    if (framePacingJvm->AttachCurrentThreadAsDaemon((void **)&env, NULL) != JNI_OK) {
```
igordmn: Here is a possible memory leak, we need `DetachCurrentThread` (or [move](https://github.com/JetBrains/skiko/pull/1282/changes#r3959556879) threading to Kotlin)

skiko/src/awtMain/cpp/windows/FramePacing.cc
```diff
@@ -0,0 +1,505 @@
...
+        free(clock);
+        return NULL;
+    }
+    return clock;
+}
+
+extern "C" {
+
+JNIEXPORT jboolean JNICALL
+Java_org_jetbrains_skiko_swing_WinNativeClock_nativeProbeVBlank(JNIEnv *env, jclass cls)
```
igordmn: It duplicates `findOutput`, can it be extracted to one function?

And we possibly have an overhead, can we pass the found adapter further and reuse it instead of calling `findOutput`?

igordmn:
> Can this cause a desync between a regular Swing component and Compose? [...] Would we end up with red in Swing and blue in Compose for a frame?

From my understanding it is not the job of the vsync-pacer. It only initiates the animation frames, aligning them with vsync. But it doesn't mean that a frame can't be initiated another way, for example, on resizing.

skiko/src/awtMain/kotlin/org/jetbrains/skiko/swing/SkikoFramePacing.kt
```diff
@@ -0,0 +1,512 @@
...
+    protected open fun onRelease() = Unit
+
+    /**
+     * Delivers one tick to all listeners; exceptions are isolated per listener. Tick sources call
+     * this once per display refresh; delivery is skipped after the clock stopped.
+     */
+    protected fun deliver(timeNanos: Long) {
+        if (stopped) return
+
+        for (listener in listeners) {
```
ApoloApps, line 197: This allocates iterator but since CopyOnWriteArrayList is RandomAccess, can't it be iterating using indices instead to avoid allocating iterator?? I don't know if the iterator does some kind of synchronization, rendering this suggestion useless...

skiko/src/awtMain/kotlin/org/jetbrains/skiko/swing/SwingRepaintPacer.kt
```diff
@@ -0,0 +1,296 @@
...
+        subscribedDisplayId = displayId
+        failedDisplayId = UNKNOWN_DISPLAY_ID
+        return true
+    }
+
+    /**
+     * Called by the FramePacing service on a non-EDT thread; must return immediately.
+     */
+    private fun onTick() {
+        receivedTick = true
```
igordmn: Maybe we can move it after `tickChannel.receive()`? Otherwise it can race with `receivedTick = false` inside `ensureSubscription`

skiko/src/awtMain/kotlin/org/jetbrains/skiko/swing/SkikoFramePacing.kt
rock3r: Yeah, fair point. Not a huge cost but one we can spare anyway.

skiko/src/awtMain/cpp/linux/FramePacing.cc
rock3r: Fair enough, I mostly reviewed the code (and particularly the Kotlin side, since I am as you know, not super familiar with C++). I only skimmed the comments and didn't clean them up as I should have. Will do a pass.

skiko/src/awtMain/cpp/linux/FramePacing.cc
rock3r: Yeah this phrasing is incorrect, agree. Will tweak — phase misalignment can (but doesn't necessarily) cause frame drop/sync issues.

skiko/src/awtMain/cpp/linux/FramePacing.cc
rock3r: >If the second display has 165Hz, does it use 120Hz rate of the first display?

I verified when running the tests that the issue only happens if you have two screens attached running at the same refresh rate, since the CRTCs become undistinguishable. It's a relatively minor issue since that mostly affects phase, as in that case the refresh rate is indeed the same for both displays. But I could not find a way to further discriminate the screens in that case without JBR-side changes...

skiko/src/awtTest/kotlin/org/jetbrains/skiko/swing/SwingRepaintPacerTest.kt
rock3r: I only saw ULE

skiko/src/awtMain/kotlin/org/jetbrains/skiko/swing/SkikoFramePacing.kt
rock3r: Hm but I want to check whether this is indeed the case before I do change

skiko/src/awtTest/kotlin/org/jetbrains/skiko/swing/SkikoProbeMain.kt
rock3r: Yeah would need a proper Gradle task to run, I'll probably just drop it at this point

skiko/src/awtMain/kotlin/org/jetbrains/skiko/swing/SkikoFramePacing.kt
rock3r: Dangit those crazy new screens :D

skiko/src/awtMain/kotlin/org/jetbrains/skiko/swing/SkikoFramePacing.kt
rock3r: Hadn't considered power consumption on laptops due to the timers, I'll just drop it since I can't tell for sure when DWM would be entirely missing, and in that case we can just treat the pacing as unavailable

skiko/src/awtMain/cpp/windows/FramePacing.cc
rock3r: My lack of native code skills is in full display here :D

skiko/src/awtMain/cpp/windows/FramePacing.cc
rock3r: I think we can indeed move it all on the Kotlin side. I had assumed doing it on the native code side would be faster, but it is indeed expensive to maintain. I'll see what can be done.

skiko/src/awtMain/cpp/windows/FramePacing.cc
rock3r: See above, yes, will try to move as much as possible to Kotlin

skiko/src/awtMain/cpp/windows/FramePacing.cc
rock3r: Will move it all to Kotlin, should remove a bunch of these issues

skiko/src/awtMain/kotlin/org/jetbrains/skiko/swing/SwingRepaintPacer.kt
rock3r: Ah, yeah, it could race. Will do

skiko/src/awtMain/kotlin/org/jetbrains/skiko/swing/SkikoFramePacing.kt
rock3r: Checked against the JDK source before changing it, and I'm going to keep the iterator.

`CopyOnWriteArrayList.iterator()` returns a `COWIterator` over a snapshot of the
backing array (`new COWIterator<E>(getArray(), 0)`), so the whole delivery loop sees one consistent array. If I switched to `get(i)` and `size()` I'd lose that: each call re-reads the *live* array (`elementAt(getArray(), index)`/`getArray().length`).

So an index loop takes its bound from one array snapshot, and its elements from whatever array exists at that moment. If a listener unsubscribes mid-delivery — which can happen on display change or idle release — the loop can skip a listener or throw `IndexOutOfBoundsException`. Getting the snapshot behaviour back without the iterator would mean re-implementing it, so it's not worth it.

The cost is one small object allocation per tick, at a few hundred ticks per second at most.

skiko/src/awtMain/kotlin/org/jetbrains/skiko/swing/SwingRepaintPacer.kt
```diff
@@ -0,0 +1,296 @@
...
+        tickTimeoutMillis = tickTimeoutMillisFor(service.refreshPeriodNanos(displayId))
+
+        // Set before subscribing: the backend may tick as soon as the subscription exists.
+        receivedTick = false
+        // A backend that throws instead of returning null — a missing native entry point raises UnsatisfiedLinkError
+        // here — has to degrade to unpaced repaints like any other refusal. An exception escaping the frame loop would
+        // cancel it, and the layer would then stop repainting altogether.
+        val newSubscription = try {
+            service.subscribe(displayId) { _, _ -> onTick() }
+        } catch (_: Throwable) {
```
igordmn: Do we need it? In which cases `subscribe` can throw?

skiko/src/awtTest/kotlin/org/jetbrains/skiko/swing/SwingRepaintPacerTest.kt
igordmn: > I only saw ULE

But this shouldn't happen in the apps, let's not test/handle it

rock3r: Yep I have already removed it, doing a few more tests as I caught a possible null pointer issue while checking the mostly-kotlin impl on Windows :)

skiko/src/awtMain/kotlin/org/jetbrains/skiko/swing/SwingRepaintPacer.kt
rock3r: It's for the tick not the subscribe

rock3r: I could arguably move it inside the lambda tho :)

m-sasha:
This "Fixes" https://youtrack.jetbrains.com/issue/CMP-8124/ComposePanel.-Support-vsync-with-compose.swing.render.on.graphicstrue, yes?

Let's add it as a "Fixes" line in the PR summary?

rock3r:
Added

skiko/src/awtMain/kotlin/org/jetbrains/skiko/swing/SkikoFramePacing.kt
```diff
@@ -0,0 +1,434 @@
...
+    protected abstract fun onStop()
+
+    /** Delivers one tick to every listener. A throwing listener does not stop the others. */
+    protected fun deliver(timeNanos: Long) {
+        for (listener in listeners) {
+            // Re-checked per listener: the list is a snapshot, and the clock can stop mid-delivery.
+            if (stopped) return
+            try {
+                listener(displayId, timeNanos)
+            } catch (_: Throwable) {
```
igordmn:
```suggestion
            } catch (e: Throwable) {
                e.printStackTrace()
```

Otherwise we hide an exception appeared in the listener

skiko/src/awtMain/kotlin/org/jetbrains/skiko/swing/SwingRepaintPacer.kt
```diff
@@ -0,0 +1,224 @@
...
+        val displayId = resolveDisplayId(service)
+        if (displayId == UNKNOWN_DISPLAY_ID || displayId == failedDisplayId) {
+            closeSubscription()
+            return false
+        }
+
+        if (subscription != null && displayId == subscribedDisplayId) return true
+
+        closeSubscription()
+        tickTimeoutMillis = tickTimeoutMillisFor(service.refreshPeriodNanos(displayId))
```
igordmn: What if we change the display refresh rate in OS settings? Maybe should we check `refreshPeriodNanos` every time too?

igordmn (commented):
Only minor issues remain, the rest is good.

Also we need:
- fix failing tests
- test it again on all platforms, when the code is ready

skiko/src/awtMain/kotlin/org/jetbrains/skiko/swing/SwingRepaintPacer.kt
igordmn: I see the last change:
```
if (subscription != null && displayId == subscribedDisplayId && !refreshRateChanged(service)) return true
```

You asked AI to review, as I did, right? 🤣

skiko/src/awtMain/kotlin/org/jetbrains/skiko/swing/SkikoFramePacing.kt
igordmn: Also, we might remove this broken listener, to not spam errors

rock3r:
Failing tests were already fixed when I rebased :) I'll do a run of testing tomorrow on all my fleet.

skiko/src/awtMain/kotlin/org/jetbrains/skiko/swing/SwingRepaintPacer.kt
rock3r: Yep I did tell you I ran Astra as I was rebasing

skiko/src/awtMain/kotlin/org/jetbrains/skiko/swing/SkikoFramePacing.kt
rock3r: This was explicitly designed this way when it was in the JBR since the JBR should not care about a listener that's faulty, but yeah I'll change to log-and-remove

rock3r:
@igordmn re-run checks on all OSes after the small change you asked me for, all fine. There are failing checks on main which are not caused by this PR.

image

I should finish my own re-check soon btw

rock3r:
@igordmn done, nothing substantive found, just a test naming nit that is now addressed (and a good excuse to rerun the CI since the docs job had timed out last run). Good for me.

PR #1284, opened 2026-08-31
Introduce SkiaLayer.makeWithContextAttributes to pass arbitrary EmscriptenWebGLContextAttributes

eymar:
With a new method, it will be easier to experiment with different attributes values in Compose.

For, example: 
- we should consider `antialias: true` to enabled GPU based strategy when Skia decides how to render a Path. Currently skia choose a CPU based approach.


Schahen (approved):
Discussed couple of moments in Slack,  has nothing to reject to

PR #1285, opened 2026-08-31
Remove stubs.cc

hub-bla:
This PR removes the JVM `stubs.cc` completely. The stubs are only needed to satisfy native symbols at link time when using static linking, which we don't do with Skiko because we build dynamically linked libraries that are loaded at runtime.

If someone calls a function that is implemented only on a specific OS, it will crash at runtime with an `UnsatisfiedLinkError`

Fixes [SKIKO-1172](https://youtrack.jetbrains.com/issue/SKIKO-1172)


igordmn:
There are 2 stubs.cc, what about the second one?

igordmn (approved):

hub-bla:
`linux/stubs.cc` seemed unrelated to the one in this PR. It looks like some kind of compiler compatibility. I ran skiko tests with `linux/stubs.cc` removed as well. All tests passed but I think it still needs a separate investigation

PR #1287, opened 2026-08-31
Fold the macOS context handlers into their redrawers

MatkovIvan:
Another #1234's follow-up

Each native-macOS redrawer owned a ContextHandler whose only job was to hold the Skia context and surface for that same redrawer, so every frame crossed an extra object boundary and the GPU state of one backend lived in two places. Fold `MacOsMetalContextHandler` and `MacOSOpenGLContextHandler` into `MacOsMetalRedrawer` and `MacOsOpenGLRedrawer`, and delete `ContextHandler`: one object per graphics API now owns its GPU state end to end, still driven through the Redrawer interface.

Three Metal behavior fixes come with the fold. A drawable that is acquired but never presented is now released on every exception path and on dispose: the drawable pool holds about three, so a leaked one stalls the next `nextDrawable()` call for roughly a second. `nextDrawable()` returning `null` now raises `RenderException` rather than surfacing as a `NullPointerException` from a `!!` assertion. And dispose now detaches the layer from its `NSView` - clearing `nsView.layer` and `wantsLayer` - when the renderer still owns it, so a disposed renderer leaves no dangling `CALayer` on the view.

The OpenGL path is brought in line with Metal on two points: it leaked the previous `BackendRenderTarget` and `Surface` on every resize, and it asked Skia for a 0x0 render target whenever the layer size reached zero, where Metal already skipped the frame.

One more latent bug is fixed alongside: the macOS-native `contentScale` getter, plus the `syncContentScale()` of both redrawers, dereferenced the `NSWindow` of a layer that has not been attached yet, throwing an NPE on
`backingScaleFactor`.


igordmn:
Please insert the commit message to the PR description - it says that the behavior is changed

igordmn (approved):

m-sasha (approved):

PR #1289, opened 2026-09-04
[web] Introduce verifyWasmProposalsManifest and dumpWasmProposalsManifest for verification of skiko.wasm artifact

Schahen:
The goal of this PR is to be more safe and detect earlier incidents related to browser compatibility. 
Recently we've encountered a situation when the artifact built was built with newer proposals - and while each of them was beneficial and made sense both in terms of performance and memory consumption - some of them were not supported by browsers that we have an agreement to support. 

When this PR will be merged, the `verifyWasmProposalsManifest` will build the the list of proposals used in particular skiko.wasm artifact and will compare it to a reference file (which, in turn, can be update via `dumpWasmProposalsManifest` task.

The report also contains information on minimal browser version that supports particular proposal

```
proposal,chrome_min,firefox_min,safari_min
mutable-globals,74,61,13.1
sign-extension-ops,74,62,14.1
threads,74,79,14.1
nontrapping-float-to-int,75,64,15
bulk-memory-operations,75,79,15
multi-value,85,78,13.1
simd,91,89,16.4
legacy-exception-handling,95,100,15.2
reference-types,96,79,15
tail-call,112,121,18.2
typed-function-references,119,120,18
gc,119,120,18.2
exception-handling,137,131,18.4

```

### DIFF
--- a/skiko/buildSrc/src/main/kotlin/WasmProposalsManifestTask.kt
+++ b/skiko/buildSrc/src/main/kotlin/WasmProposalsManifestTask.kt
@@ -1,0 +1,577 @@
+import org.gradle.api.DefaultTask
+import org.gradle.api.file.RegularFileProperty
+import org.gradle.api.tasks.InputFile
+import org.gradle.api.tasks.OutputFile
+import org.gradle.api.tasks.PathSensitive
+import org.gradle.api.tasks.PathSensitivity
+import org.gradle.api.tasks.TaskAction
+
+abstract class WasmProposalsManifestTask : DefaultTask() {
+
+    @get:InputFile
+    @get:PathSensitive(PathSensitivity.NONE)
+    abstract val wasmFile: RegularFileProperty
+
+    @get:OutputFile
+    abstract val manifestFile: RegularFileProperty
+
+    @TaskAction
+    fun generate() {
+        val wasm = wasmFile.get().asFile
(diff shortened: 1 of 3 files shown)


skiko/buildSrc/src/main/kotlin/WasmProposalsManifestTask.kt
```diff
@@ -0,0 +1,577 @@
...
+            (importSection != null && detectMutableGlobalImport(importSection)) ||
+                (exportSection != null && globalSection != null && detectMutableGlobalExport(bytes))
+        )
+
+        // Sign extension ops: 0xC0-0xC4
+        if (codeSection != null) {
+            val signExtOpcodes = listOf(0xC0, 0xC1, 0xC2, 0xC3, 0xC4).map { it.toByte() }
+            entries += proposalEntry(
+                "sign-extension-ops",
+                signExtOpcodes.any { it in codeSection }
```
eymar: as I understand this will be `true` if the specific byte is present anywhere in the `codeSection`.

`0xC0-0xC4` - can't these values be found among the operands of the instrctions?  If they can then in a big-enough binary all proposals will be flagged as present.  (I mean it will find those bytes anywhere: in instructions, in operands, etc)

Can we use `wasm-opt skiko.wasm --print-features` rather than implementing the decoder ourselves?

PR #1292, opened 2026-09-07
Extract skiko rendering code into separate module

hub-bla:
This branch introduces a new `skiko-rendering` module that separates platform rendering integration from the core skiko library.

The new module owns the higher-level rendering infrastructure, including SkiaLayer, frame dispatching, render delegates, AWT/Swing integration, platform renderers, ANGLE support.

Fixes [SKIKO-1189](https://youtrack.jetbrains.com/issue/SKIKO-1189)


skiko/settings.gradle.kts
```diff
@@ -22,3 +22,4 @@ include("import-generator")
 include("test-utils")
 include("skiko-skottie")
 include("skiko-graphite")
```
igordmn, line 24: Please update `skiko/README.md` on how to include the new module

skiko/skiko-rendering/src/awtMain/kotlin/org/jetbrains/skiko/renderer/MetalRenderer.kt
```diff
@@ -39,7 +39,7 @@ internal class MetalRenderer(
 
     companion object {
         init {
-            Library.load()
+            RenderingLibrary.load()
```
igordmn, line 42: We need to also call it in standalone functions like:
```
initializeCAccessible
JumpList/JumpListBuilder
```
Previously it was also needed, but the Skiko library is loading more often then the rendering library.

skiko/settings.gradle.kts
```diff
@@ -22,3 +22,4 @@ include("import-generator")
...
+include("skiko-rendering")
```
igordmn, line 25:
```suggestion
include("skiko-platform")
```

It is more about connecting the Skiko API with graphic API, windows and other things like accessibility, jump list, etc. Most are even not related to Skia, but for now we can treat it as a general purpose "platform" module.

@hub-bla, @MatkovIvan, what do you think?

skiko/skiko-rendering/src/awtMain/kotlin/org/jetbrains/skiko/renderer/MetalRenderer.kt
hub-bla: Done!

skiko/settings.gradle.kts
hub-bla: I agree

Done!

skiko/skiko-rendering/src/awtMain/kotlin/org/jetbrains/skiko/renderer/MetalRenderer.kt
hub-bla: Done!

igordmn (approved):
Only skiko/README.md needs to be updated

igordmn (approved):

PR #1295, opened 2026-09-07
Fix composite build for skiko samples.

terrakok:


Schahen (approved):
Everything works like charm, thank you so much for taking care of this!

PR #1301, opened 2026-09-16
Extract ganesh into separate module

hub-bla:
This PR extracts Ganesh GPU backend support from the core skiko module into a new `skiko-ganesh` module. 

Consumers using Ganesh APIs will need to add the `skiko-ganesh` dependency and update imports for APIs moved to `org.jetbrains.skia.gpu.ganesh`

Fixes [SKIKO-1190](https://youtrack.jetbrains.com/issue/SKIKO-1190)


skiko/src/commonMain/kotlin/org/jetbrains/skia/Bitmap.kt
```diff
@@ -15,15 +15,6 @@ class Bitmap internal constructor(ptr: NativePointer) : Managed(ptr, _FinalizerH
             }
         }
 
-        fun makeFromImage(image: Image, context: DirectContext): Bitmap {
```
igordmn: Is this related to Ganesh?

igordmn:
Please, test in Compose first, before merging

hub-bla:
I already did it locally but I'll definitely create a PR in Compose with all the necessary changes before merging this one

skiko/build.gradle.kts
```diff
@@ -42,7 +42,6 @@ val coreDependencies: SkikoDependencyScope.() -> Unit = {
         all {
```
igordmn, line 42: Please, register the new libs in https://github.com/JetBrains/skiko/blob/f6156fa3ccf64bc8073c71b525736ef0c53ee2ca/skiko/ci/build.gradle.kts#L19-L84

igordmn, line 42: We need to also add them to Dokka https://github.com/JetBrains/skiko/blob/f6156fa3ccf64bc8073c71b525736ef0c53ee2ca/skiko/docs/build.gradle.kts#L17-L24

skiko/skiko-rendering/build.gradle.kts
```diff
@@ -221,12 +230,14 @@ if (supportWeb) {
             - implementation("org.jetbrains.skiko:skiko-rendering-x")
          */
         compileOnly(project(":"))
+        compileOnly(project(":skiko-ganesh"))
```
igordmn, line 233:
```suggestion
        implementation(project(":skiko-ganesh"))
```

Looks like the current renderers works only with Ganesh? Maybe it is good idea to change it only when we support Graphite.

skiko/skiko-ganesh/src/awtMain/cpp/windows/InternalDirectXApi.cc
igordmn, line 1: Looks like this file and OpenGLLibrary belong to `skiko-rendering`?

hub-bla: `OpenGLLibrary` is used for the OpenGL and WGL symbols required by Skia’s ganesh GL backend when linking `skiko-ganesh` as a dynamic library on Windows:

https://github.com/JetBrains/skiko/blob/c4b1940eaa73abfcbaa3f3f5d5972ab0ac7b9810/skiko/src/awtMain/cpp/windows/OpenGLLibrary.cc#L9-L70

Since I believe we intentionally avoid linking `OpenGL.lib` these wrapped symbols satisfy Skia ganesh's linktime symbol requirements and thats why `OpenGLLibrary` is in `skiko-ganesh`

skiko/build.gradle.kts
hub-bla: Thanks for pointing this out! Totally forgot about it.

Done!

hub-bla: Done!

skiko/skiko-rendering/build.gradle.kts
hub-bla: The problem is that `skiko-ganesh` has multiple publications which makes it ambiguous for Gradle. Because of that, switching it directly to implementation will fail. The solution I initially found was to keep using `compileOnly,` but I've revisited the problem and found another approach using Gradle dependency substitution which resolves the ambiuity. I updated it to the substitution solution now

hub-bla: Looks like the new approach is failing. 

Also, after thinking about it some more, even with `implementation(project(":skiko-ganesh"))`, users would still need to add `skiko-ganesh-awt-runtime-x` separately while `skiko-ganesh` would be pulled in transitively.

Keeping it as `compileOnly` is a bit more verbose, since users would need to explicitly add both `skiko-ganesh` and `skiko-ganesh-awt-runtime-x` but it might actually be cleaner that way. wdyt?

skiko/skiko-ganesh/src/awtMain/cpp/windows/InternalDirectXApi.cc
igordmn: Make sense for `OpenGLLibrary`. But I would still move `InternalDirectXApi` - it actually not the API, just a bunch of Utils for renderers.

skiko/skiko-rendering/build.gradle.kts
igordmn: > I updated it to the substitution solution now

It would probably worked only for this project, but we need to fix it for the users.

> need to explicitly add both skiko-ganesh and skiko-ganesh-awt-runtime-x but it might actually be cleaner that way

Agree, if we still require defining `skiko-ganesh-awt-runtime` explicitly, we have to require defining `skiko-ganesh` too. 

But I would think if we can't not require defining `skiko-ganesh-awt-runtime` explicitly, because `Ganesh` is still implementation details for `skiko-rendering` users.

Maybe we can make transitive dependencies `skiko-rendering-awt-runtime-x -> skiko-ganesh-awt-runtime-x`.

Also, worth to think how we want it to be used with Graphite when it is implemented for `skiko-rendering`, maybe `skiko-rendering-awt-runtime-x -> skiko-ganesh-awt-runtime-x` won't be an option then.

skiko/skiko-platform/src/awtMain/cpp/windows/InternalDirectXApi.cc
hub-bla: Similar to `OpenGLLibrary`, `InternalDirectXApi` also had symbols required to build `skiko-ganesh`. Left them in [‎skiko/skiko-ganesh/src/awtMain/cpp/windows/Direct3DLibrary.cc](https://github.com/JetBrains/skiko/pull/1301/changes/97b29886a813d54de31f2e3aa0fd6ac1941bfe56#diff-432a606b034c1b34788c089e37427caeb999821ea31e31fba9d75fd9fc2130ab). The rest was moved to `skiko-platform`

